PR #2, opened 2020-10-15
Define Array.size

kaushikcfd:
- Defined 'size = prod(self.shape)`
- Had to rename some slice attributes in the process


inducer:
Thx!

PR #5, opened 2020-11-01
Define pytato.reshape

kaushikcfd:
Current limitation: Can only handle integral oldshape, newshape

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1481,11 +1481,13 @@ def reshape(array: Array, newshape: Sequence[int], order: str = "C") -> Array:
     if newshape.count(-1) > 1:
         raise ValueError("can only specify one unknown dimension")
 
-    if not isinstance(array.size, int):
+    if not all(isinstance(axis_len, int) for axis_len in array.shape):
         raise ValueError("reshape of arrays with symbolic lengths not allowed")
 
     if order != "C":
         raise ValueError("Reshapes to a 'F'-ordered arrays not implemented")
+
+    newshape_explicit = []
 
     for new_axislen in newshape:
         if not isinstance(new_axislen, int):
@@ -1494,6 +1496,8 @@ def reshape(array: Array, newshape: Sequence[int], order: str = "C") -> Array:
         if not(new_axislen > 0 or new_axislen == -1):
             raise ValueError("newshape should be either sequence of positive ints or"
                     " -1")
+
+        # {{{ infer the axis length corresponding to axis marked "-1"
 
         if new_axislen == -1:
             size_of_rest_of_newaxes = -1 * product(newshape)
@@ -1504,13 +1508,15 @@ def reshape(array: Array, newshape: Sequence[int], order: str = "C") -> Array:
 
             new_axislen = array.size // size_of_rest_of_newaxes
 
-        newshape_sans_minus_1.append(new_axislen)
+        # }}}
 
-    if product(newshape_sans_minus_1) != array.size:
+        newshape_explicit.append(new_axislen)
+
+    if product(newshape_explicit) != array.size:
         raise ValueError(f"cannot reshape array of size {array.size}"
                 f" into {newshape}")
 
-    return Reshape(array, tuple(newshape_sans_minus_1), order)
+    return Reshape(array, tuple(newshape_explicit), order)
 
 
 def make_dict_of_named_arrays(data: Dict[str, Array]) -> DictOfNamedArrays:
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1077,7 +1077,31 @@ class AxisPermutation(IndexRemappingBase):
 
 class Reshape(IndexRemappingBase):
     """
+    Reshape an array.
+
+    .. attribute:: array
+
+        The array to be reshaped
+
+    .. attribute:: newshape
+
+        The output shape
     """
+
+    _fields = Array._fields + ("array", "newshape")
+    _mapper_method = "map_reshape"
+
+    def __init__(self,
+            array: Array,
+            newshape: Tuple[int, ...],
+            tags: Optional[TagsType] = None):
+        # TODO: should also take in 'order'
+        super().__init__(array, tags)
+        self.newshape = newshape
+
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        return self.newshape
 
 # }}}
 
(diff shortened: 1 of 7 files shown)


pytato/array.py
```diff
@@ -1076,8 +1078,32 @@ def shape(self) -> ShapeType:
...
     """
+    Reshape an array.
+
+    .. attribute:: array
+
+        The array to be reshaped
+
+    .. attribute:: newshape
+
+        The output shape
```
inducer, line 1089: Absent an explicitly passed `order`, should document which is used.

pytato/array.py
```diff
@@ -1431,6 +1457,51 @@ def _make_slice(array: Array, starts: Sequence[int], stops: Sequence[int]) -> Ar
...
+
+        reshapes of arrays with symbolic shapes not yet implemented.
+    """
+    from pytools import product
+
+    if newshape.count(-1) > 1:
+        raise ValueError("can only specify one unknown dimension")
+
+    if not isinstance(array.size, int):
+        raise ValueError("reshape of arrays with symbolic lengths not allowed")
```
inducer, line 1475: I'm not sold that this is sufficient. Do an `all(...)` check instead?

pytato/array.py
```diff
@@ -1431,6 +1457,51 @@ def _make_slice(array: Array, starts: Sequence[int], stops: Sequence[int]) -> Ar
...
+        if new_axislen == -1:
+            size_of_rest_of_newaxes = -1 * product(newshape)
+
+            if array.size % size_of_rest_of_newaxes != 0:
+                raise ValueError(f"cannot reshape array of size {array.size}"
+                        f" into ({size_of_rest_of_newaxes})")
+
+            new_axislen = array.size // size_of_rest_of_newaxes
+
+        newshape_sans_minus_1.append(new_axislen)
```
inducer, line 1496: Block-comment that the role of this is to figure out the concrete value for `-1`.

pytato/array.py
```diff
@@ -1431,6 +1457,51 @@ def _make_slice(array: Array, starts: Sequence[int], stops: Sequence[int]) -> Ar
...
+    """
+    from pytools import product
+
+    if newshape.count(-1) > 1:
+        raise ValueError("can only specify one unknown dimension")
+
+    if not isinstance(array.size, int):
+        raise ValueError("reshape of arrays with symbolic lengths not allowed")
+
+    newshape_sans_minus_1 = []
```
inducer, line 1477:
```suggestion
    newshape_explicit = []
```
(Or another name? `sans_minus_1` made me think that the -1 would simply be removed.)

inducer (commented):
This looks good to me except for a few minor style issues. Thanks for working on this!

pytato/array.py
```diff
@@ -1076,8 +1078,32 @@ def shape(self) -> ShapeType:
 
 class Reshape(IndexRemappingBase):
```
inducer, line 1079: Needs to be supported by the `DependencyMapper`.

pytato/array.py
kaushikcfd: 78cb131c5e3fa7e6316a61a5141841604fa07b4d

pytato/array.py
kaushikcfd: d62b4d0c39b5fafaa1a24c95496c04ef8ad56817

pytato/array.py
kaushikcfd: d62b4d0c39b5fafaa1a24c95496c04ef8ad56817

pytato/array.py
kaushikcfd: e860821819dc5224f80b3da9c039bbe7f85204d9

inducer:
Ready?

kaushikcfd:
Yep. (Oops, forgot to hit request for review)

pytato/array.py
kaushikcfd: d62b4d0

pytato/array.py
```diff
@@ -247,11 +247,12 @@ def ref(self, name: str) -> Array:
 
 # {{{ shape
 
-ShapeType = Tuple[ScalarExpression, ...]
+AxisLenType = Union[int, prim.Expression]
```
inducer, line 250: Introduce an `IntegralScalarExpression` instead, in `scalar_expr`, and link to the mypy bug there.

pytato/array.py
```diff
@@ -1076,8 +1078,40 @@ def shape(self) -> ShapeType:
...
+    _fields = Array._fields + ("array", "newshape", "order")
+    _mapper_method = "map_reshape"
+
+    def __init__(self,
+            array: Array,
+            newshape: Tuple[int, ...],
+            order: str,
+            tags: Optional[TagsType] = None):
+        # FIXME: Get rid of this restriction
+        assert order == "C"
```
inducer, line 1105: Assert is not acceptable for input checking. Use `if`/`raise`.

pytato/array.py
```diff
@@ -1431,6 +1465,60 @@ def _make_slice(array: Array, starts: Sequence[int], stops: Sequence[int]) -> Ar
...
+    from pytools import product
+
+    if newshape.count(-1) > 1:
+        raise ValueError("can only specify one unknown dimension")
+
+    if not all(isinstance(axis_len, int) for axis_len in array.shape):
+        raise ValueError("reshape of arrays with symbolic lengths not allowed")
+
+    if order != "C":
+        raise ValueError("Reshapes to a 'F'-ordered arrays not implemented")
```
inducer, line 1488:
```suggestion
        raise NotImplementedError("Reshapes to a 'F'-ordered arrays")
```

test/test_codegen.py
```diff
@@ -405,6 +405,36 @@ def test_stack(ctx_factory, input_dims):
...
+    from numpy.random import default_rng
+    rng = default_rng()
+    x_in = rng.random(size=oldshape)
+
+    namespace = pt.Namespace()
+    x = pt.make_data_wrapper(namespace, x_in)
+    expected_out = np.reshape(x_in, newshape=newshape)
+
+    prog = pt.generate_loopy(
+            pt.reshape(x, newshape=newshape),
```
inducer, line 430: There should only be one call to `mod.reshape`, with `mod` varying over `np` and `pt`.

kaushikcfd: Being fixed at https://github.com/inducer/pytato/pull/3#discussion_r523705649

pytato/array.py
kaushikcfd: `Reshape` is not user-facing but `reshape` is, that's where we `raise`.

inducer: OIC

pytato/array.py
kaushikcfd: ce52200

I think the link to the mypy bug is correctly placed, as the problem is: mypy cannot infer that `int` is a subclass of `Number`, which is where we add `type:ignore` and the bug link to go with it.

The reason we had to switch from ScalarExpression to IntegralScalarExpression in this PR was because of another similar mypy bug where it couldn't figure out that result of binary arith. operations between `Number` and `int` is a `Number`. But, having ScalarExpression as the type for the length of an axis was imprecise to begin with.

kaushikcfd:
Thanks for the comments. Fixed some, raised arguments for some. Let me know what do you think.

inducer:
Thanks!

PR #6, opened 2020-11-04
Improves codegen for dict of named array

kaushikcfd:


pytato/transform.py
```diff
@@ -186,6 +234,15 @@ def copy_dict_of_named_arrays(source_dict: DictOfNamedArrays,
         data[name] = copy_mapper(val)
     return DictOfNamedArrays(data)
 
+
+def get_dependencies(expr: DictOfNamedArrays) -> Dict[str, FrozenSet[Array]]:
+    dep_mapper = DependencyMapper()
+
+    if isinstance(expr, Array):
+        return dep_mapper(expr)
```
inducer, line 242: This seems to contradict the type signature. Should the type signature be a union type?

pytato/transform.py
```diff
@@ -147,6 +148,53 @@ def map_data_wrapper(self, expr: DataWrapper) -> Array:
     def map_size_param(self, expr: SizeParam) -> Array:
         return SizeParam(self.namespace, name=expr.name, tags=expr.tags)
 
+
+class DependencyMapper(Mapper):
```
inducer, line 152: Add to docs?

inducer (commented):
Thanks for working on this! This looks good, however it has a CI failure. I also noticed a few minor things.

doc/conf.py
```diff
@@ -31,6 +31,7 @@
     'sphinx.ext.autodoc',
     'sphinx.ext.intersphinx',
     'sphinx.ext.mathjax',
+    'sphinx_autodoc_typehints',
```
kaushikcfd, line 34: Not sure about this. Was getting doc building failure like [these](https://github.com/inducer/pytato/runs/1388809911#step:4:602). Adding this extension solves it.

pytato/transform.py
kaushikcfd: 7ce2067

pytato/transform.py
kaushikcfd: 604a13b

kaushikcfd:
In order to pass the type-checking, needs some type annotation on the pytools end: https://github.com/inducer/pytools/pull/50.

.test-conda-env-py3.yml
```diff
@@ -10,3 +10,4 @@ dependencies:
 - mako
 - pyopencl
 - islpy
+- sphinx-autodoc-typehints
```
inducer, line 13: I think this is redundant (to install the package).

doc/conf.py
inducer: I tried this locally, it made no difference (Py3.8, sphinx 3.3.0 and 3.3.1)

I also don't think it deserves to make a difference, as my understanding is that all that got merged into sphinx proper.

inducer (commented):
Thanks! I really like the cod parts of this, but I'm not sure what to make of the doc tool part. Could you take a look?

doc/conf.py
kaushikcfd: Very strange. On my local machine (identical configuration), without the extension I get `docstring of pytato.codegen.PreprocessResult:: WARNING: py:class reference target not found: Tuple[str, ...]`, which is what the CI reported too.(https://github.com/inducer/pytato/runs/1388809911#step:4:602)

pytato/codegen.py
```diff
@@ -837,8 +864,11 @@ def generate_loopy(result: Union[Array, DictOfNamedArrays],
         _ = cg_mapper(val, state)
 
     # Generate code for outputs.
-    for name, expr in outputs.items():
-        add_store(name, expr, cg_mapper(expr, state), state)
+    for name in compute_order:
+        expr = outputs[name]
+        insn_id = add_store(name, expr, cg_mapper(expr, state), state)
+        # replace "expr" with the created stored variable
+        state.results[expr] = StoredResult(name, expr.ndim, frozenset([insn_id]))
```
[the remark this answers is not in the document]
kaushikcfd: Yep, agreed.

.test-conda-env-py3.yml
kaushikcfd: I don't think so. If you see ba31c29 fails but 304b98f passes.

inducer: Oh nvm. I didn't spot that the above was for Gitlab. D'oh. All good.

doc/conf.py
inducer: Yeah, I had similarly weird results. My sense is that sphinx + type annotations isn't totally mature tech. I'm OK with merging this. We can always dredge this up if it causes problems later. Thanks for taking a look.

PR #11, opened 2020-12-06
Generate reproducible loopy kernels

kaushikcfd:
Traversal order of `DataWrapper`'s determines order of `LoopKernel.args", having them reproducible is necessary for on-disk cache hits.


inducer:
Thx!

PR #12, opened 2020-12-10
Allow dict as an input to pt.generate_loopy

kaushikcfd:
- Would be convenient to have this, especially in cases when one doesn't intend to use the components of `DictOfNamedArrays` in a lazy manner.


inducer:
Not opposed. LGTM. Thanks!

PR #13, opened 2021-01-05
Adds only dependencies of the array expression to the generated kernel

kaushikcfd:
Closes https://gitlab.tiker.net/inducer/pytato/-/issues/13

### DIFF
--- a/pytato/transform.py
+++ b/pytato/transform.py
@@ -212,19 +212,19 @@ class DependencyMapper(Mapper):
 # {{{ mapper frontends
 
 def copy_namespace(source_namespace: Namespace,
-        copy_mapper: CopyMapper, onlynames: Optional[FrozenSet[str]]) -> Namespace:
+        copy_mapper: CopyMapper, only_names: Optional[FrozenSet[str]]) -> Namespace:
     """Copy the elements of *namespace* into a new namespace.
 
     :param source_namespace: The namespace to copy
     :param copy_mapper: A mapper that performs copies into a new namespace
-    :param onlynames: Elements of *namespace* that are to be copied into the
-        new namespace. All elements are copied if *onlynames=None*,
+    :param only_names: Elements of *namespace* that are to be copied into the
+        new namespace. All elements are copied if *only_names=None*,
     :returns: A new namespace containing copies of the items in *source_namespace*
     """
-    if onlynames is None:
-        onlynames = frozenset(source_namespace.keys())
+    if only_names is None:
+        only_names = frozenset(source_namespace.keys())
 
-    for name in onlynames:
+    for name in only_names:
         mapped_val = copy_mapper(source_namespace[name])
         if name not in copy_mapper.namespace:
             copy_mapper.namespace.assign(name, mapped_val)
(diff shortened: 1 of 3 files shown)


pytato/transform.py
```diff
@@ -212,15 +212,20 @@ def map_concatenate(self, expr: Concatenate) -> FrozenSet[Array]:
 # {{{ mapper frontends
 
 def copy_namespace(source_namespace: Namespace,
-        copy_mapper: CopyMapper) -> Namespace:
+        copy_mapper: CopyMapper, onlynames: Optional[FrozenSet[str]]) -> Namespace:
```
inducer, line 215:
```suggestion
        copy_mapper: CopyMapper, only_names: Optional[FrozenSet[str]]) -> Namespace:
```

test/test_codegen.py
```diff
@@ -65,19 +65,6 @@ def test_scalar_placeholder(ctx_factory):
...
-    ctx = ctx_factory()
-    queue = cl.CommandQueue(ctx)
-
-    namespace = pt.Namespace()
-    n = pt.make_size_param(namespace, name="n")
-    pt.make_placeholder(namespace, name="x", shape="(n,)", dtype=np.int)
-    prog = pt.generate_loopy(n, target=pt.PyOpenCLTarget(queue))
-    x_in = np.array([1, 2, 3, 4, 5])
-    _, (n_out,) = prog(x=x_in)
-    assert n_out == 5
```
inducer, line 78: How come this is being removed?

inducer (commented):
Thanks! This looks good to me. Just two minor concerns.

test/test_codegen.py
kaushikcfd: According to this test "n" depends on "x", whereas "x" depends on "n". So, it was incorrect to being with. Because all the arrays were fed as arguments to the kernel this test sneakily made it through the CI.

pytato/transform.py
kaushikcfd: Thanks! See 1ecb3a0

test/test_codegen.py
inducer: I kind of disagree here, I think the code in this test should be allowed. Technically, the shape of `x` depends on `n`, however `n` can be inferred from `x` if given.

kaushikcfd: Agreed upon solution: This is a good convenience feature to have and should stay. Pytato should support it by having it's own shape inference solver, without having to piggyback on loopy.

For now, this can be `xfail`-ed.

kaushikcfd: 8b8c7da

inducer: #15

test/test_codegen.py
```diff
@@ -65,6 +65,7 @@ def test_scalar_placeholder(ctx_factory):
     assert np.array_equal(x_out, x_in)
 
 
+@pytest.mark.xfail  # shape inference solver: not yet implemented
```
inducer, line 68:
```suggestion
# https://github.com/inducer/pytato/issues/15
@pytest.mark.xfail  # shape inference solver: not yet implemented
```

inducer:
Thanks!

PR #18, opened 2021-01-10
Prefer IndexLambda for elementwise expressions

inducer:
@kaushikcfd How do you feel about this?

x-ref: #17

### DIFF
--- a/doc/design.rst
+++ b/doc/design.rst
@@ -63,7 +63,10 @@ Computation and Results
 
     Operations that introduce nontrivial mappings on indices (e.g. reshape,
     strided slice, roll) are identified as potential candidates for being captured
-    in their own high-level node vs. as an :class:`pytato.array.IndexLambda`.
+    in their own high-level node vs. as an :class:`~pytato.array.IndexLambda`.
+    
+    Operations that *can* be expressed as :class:`~pytato.array.IndexLambda`
+    without loss of information, *should* be expressed that way.
 
 Naming
 ------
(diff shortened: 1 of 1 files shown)


doc/design.rst
```diff
@@ -63,7 +63,10 @@ Computation and Results
 
     Operations that introduce nontrivial mappings on indices (e.g. reshape,
     strided slice, roll) are identified as potential candidates for being captured
-    in their own high-level node vs. as an :class:`pytato.array.IndexLambda`.
+    in their own high-level node vs. as an :class:`~pytato.array.IndexLambda`.
+    
+    Operations that *can* be expressed as :class:`~pytato.array.IndexLambda`
+    without loss of information, *should* be expressed that way.
```
kaushikcfd, line 69: I don't think we've defined what is considered as "loss of information"?

Are math function to index lambdas not loss of information?

kaushikcfd: One reason for math functions have their own nodes is possibly one could perform some DAG traversals to perform algebraic simplifications.  (such as applying some trigonometric identities)

inducer: That's equally possible if they're in `IndexLambda`s.

inducer: > I don't think we've defined what is considered as "loss of information"?

Gave this a shot in c57ee7d.

kaushikcfd: Thank you, clear now.

kaushikcfd:
LGTM from my end.

inducer:
Thanks for taking a look!

PR #16, opened 2021-01-10
Support loopy callables branch

kaushikcfd:

PR #20, opened 2021-01-18
Implements pytato.loopy.call_loopy

kaushikcfd:
- Branched from #16 
- Adds support for calling loopy kernels

TODO:
- undo change to https://github.com/inducer/pytato/pull/20/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,7 @@ jobs:
                 python-version: '3.x'
         -   name: "Main Script"
             run: |
+                sed -i "s/loopy.git/loopy.git@kernel_callables_v3-edit2/g" requirements.txt
                 curl -L -O -k https://tiker.net/ci-support-v0
                 . ci-support-v0
                 build_py_project_in_conda_env
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,5 @@ jobs:
                 curl -L -O -k https://tiker.net/ci-support-v0
                 . ci-support-v0
                 build_py_project_in_conda_env
-                # Circular references in type hints. Sphinx currently doesn't deal well with those.
-                # e.g. https://gitlab.tiker.net/inducer/pytato/-/jobs/202899
-                build_docs --no-check
+                build_docs
 
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+SPHINXOPTS    ?= -n
 SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -170,7 +170,7 @@ from functools import partialmethod
 import operator
 from typing import (
         Optional, Callable, ClassVar, Dict, Any, Mapping, Iterator, Tuple, Union,
-        Protocol, Sequence, cast, TYPE_CHECKING, FrozenSet)
+        Protocol, Sequence, cast, TYPE_CHECKING)
 
 import numpy as np
 import pymbolic.primitives as prim
@@ -627,25 +627,25 @@ class NamedArray(Array):
     def __init__(self,
             dict_of_named_arrays: DictOfNamedArrays,
             name: str,
-            tags: TagsType = frozenset()):
+            tags: TagsType = frozenset()) -> None:
         super().__init__(tags=tags)
         self.dict_of_named_arrays = dict_of_named_arrays
         self.name = name
 
     @property
-    def expr(self):
+    def expr(self) -> Array:
         return self.dict_of_named_arrays._data[self.name]
 
     @property
-    def shape(self):
+    def shape(self) -> ShapeType:
         return self.expr.shape
 
     @property
-    def dtype(self):
+    def dtype(self) -> np.dtype:
         return self.expr.dtype
 
 
-class DictOfNamedArrays(Mapping[str, Array]):
+class DictOfNamedArrays(Mapping[str, NamedArray]):
     """A container that maps valid Python identifiers
     to instances of :class:`Array`. May occur as a result
     type of array computations.
@@ -669,9 +669,9 @@ class DictOfNamedArrays(Mapping[str, Array]):
         self._data = data
 
     def __contains__(self, name: object) -> bool:
-        return name in self._namespace
+        return name in self.namespace
 
-    def __getitem__(self, name: str) -> Array:
+    def __getitem__(self, name: str) -> NamedArray:
         return self._named_arrays[name]
 
     def __iter__(self) -> Iterator[str]:
@@ -679,10 +679,6 @@ class DictOfNamedArrays(Mapping[str, Array]):
 
     def __len__(self) -> int:
         return len(self._named_arrays)
-
-    @property
-    def exprs(self) -> FrozenSet[Array]:
-        return frozenset(self._data.values())
 
     @memoize_method
     def __hash__(self) -> int:
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -118,7 +118,6 @@ Built-in Expression Nodes
 .. autoclass:: MatrixProduct
 .. autoclass:: Stack
 .. autoclass:: Concatenate
-.. autoclass:: AttributeLookup
 
 Index Remapping
 ^^^^^^^^^^^^^^^
@@ -889,20 +888,6 @@ class Concatenate(Array):
         return (self.arrays[0].shape[:self.axis]
                 + (common_axis_len,)
                 + self.arrays[0].shape[self.axis+1:])
-
-# }}}
-
-
-# {{{ attribute lookup
-
-class AttributeLookup(Array):
-    """An expression node to extract an array from a :class:`DictOfNamedArrays`.
-
-    .. warning::
-
-        Not yet implemented.
-    """
-    pass
 
 # }}}
 
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -116,7 +116,6 @@ Built-in Expression Nodes
 .. autoclass:: IndexLambda
 .. autoclass:: Einsum
 .. autoclass:: MatrixProduct
-.. autoclass:: LoopyCall
 .. autoclass:: Stack
 .. autoclass:: Concatenate
 .. autoclass:: AttributeLookup
@@ -284,7 +283,7 @@ def _truediv_result_type(arg1: DtypeOrScalar, arg2: DtypeOrScalar) -> np.dtype[A
 
 
 class Array(Taggable):
-    """
+    r"""
     A base class (abstract interface + supplemental functionality) for lazily
     evaluating array expressions. The interface seeks to maximize :mod:`numpy`
     compatibility, though not at all costs.
@@ -301,9 +300,9 @@ class Array(Taggable):
 
     .. attribute:: shape
 
-        A tuple of integers or scalar-shaped :class:`~pytato.array.Array`s.
+        A tuple of integers or scalar-shaped :class:`~pytato.array.Array`\ s.
         Array-valued shape components may be (at most affinely) symbolic in terms of
-        :class:`~pytato.array.SizeParam`s.
+        :class:`~pytato.array.SizeParam`\ s.
 
         .. note::
 
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -170,11 +170,12 @@ canonicalize type references. Once Sphinx 4.0 is released, we should use the
 
 # }}}
 
+from abc import ABC, abstractmethod
 from functools import partialmethod
 import operator
 from typing import (
-        Optional, Callable, ClassVar, Dict, Any, Mapping, Iterator, Tuple, Union,
-        Protocol, Sequence, cast, TYPE_CHECKING, List)
+        Optional, Callable, ClassVar, Dict, Any, Mapping, Tuple, Union,
+        Protocol, Sequence, cast, TYPE_CHECKING, List, Iterator)
 
 import numpy as np
 import pymbolic.primitives as prim
@@ -631,20 +632,23 @@ class NamedArray(Array):
 
     .. automethod:: __init__
     """
-    _fields = Array._fields + ("dict_of_named_arrays", "name")
+    _fields = Array._fields + ("_container", "name")
     _mapper_method = "map_named_array"
 
     def __init__(self,
-            dict_of_named_arrays: DictOfNamedArrays,
+            container: AbstractResultWithNamedArrays,
             name: str,
             tags: TagsType = frozenset()) -> None:
         super().__init__(tags=tags)
-        self.dict_of_named_arrays = dict_of_named_arrays
+        self._container = container
         self.name = name
 
     @property
     def expr(self) -> Array:
-        return self.dict_of_named_arrays._data[self.name]
+        if isinstance(self._container, DictOfNamedArrays):
+            return self._container._data[self.name]
+        else:
+            raise TypeError("only permitted when container is a DictOfNamedArrays")
 
     @property
     def shape(self) -> ShapeType:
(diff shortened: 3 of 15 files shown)


kaushikcfd:
Pylint failures are tricky to handle cleany, as most of the `pytato.call_loopy` depends on interfaces that aren't part of loopy's master branch yet and adding those many "ignore" annotations would be quite noisy. Happy to hear if there's a better option.

test/test_codegen.py
```diff
@@ -497,6 +506,108 @@ def test_math_functions(ctx_factory, dtype, function_name):
...
+    x = pt.make_placeholder(ns, shape="(M, N)", dtype=np.float, name="x")
+    y = 3*x
+
+    knl = lp.make_kernel(
+            "{[i, j]: 0<=i<m and 0<=j<n}",
+            """
+            Z[i] = 7*sum(j, Y[i, j])
+            """, name="callee", lang_version=(2018, 2))
+
+    loopyfunc = call_loopy(ns, knl, bindings={"Y": y, "m": "M", "n": "N"})
```
inducer, line 603: To `**kwargs` or not to `**kwargs`? Maybe have a kwargs'd version?

inducer, line 603: Shape inference? (OK to bump to issue for later)

pytato/array.py
```diff
@@ -755,25 +784,32 @@ class DictOfNamedArrays(Mapping[str, Array]):
...
+    def __getitem__(self, name: str) -> NamedArray:
+        return self._named_arrays[name]
 
     def __iter__(self) -> Iterator[str]:
-        return iter(self._data)
+        return iter(self._named_arrays)
 
     def __len__(self) -> int:
-        return len(self._data)
+        return len(self._named_arrays)
```
inducer, line 808: Why this change of traversal behavior? This broke `pytato.visualization` (in the sense that that now only shows the `NamedArray`s, which isn't useful).

cc @VincentWells

inducer: @VincentWells It seems like this diff gets the visualizer going again:
```diff
diff --git a/pytato/visualization.py b/pytato/visualization.py
index c4f22f1..09aff97 100644
--- a/pytato/visualization.py
+++ b/pytato/visualization.py
@@ -201,7 +201,7 @@ def get_dot_graph(result: Union[Array, DictOfNamedArrays]) -> str:
 
     nodes: Dict[Array, DotNodeInfo] = {}
     mapper = ArrayToDotNodeInfoMapper()
-    for elem in itertools.chain(namespace.values(), outputs.values()):
+    for elem in itertools.chain(namespace.values(), outputs._data.values()):
         mapper(elem, nodes)
 
     input_arrays: List[Array] = []
@@ -239,7 +239,7 @@ def get_dot_graph(result: Union[Array, DictOfNamedArrays]) -> str:
                 emit('%s -> %s [label="%s"]' % (tail, head, dot_escape(label)))
 
         # Emit output/namespace name mappings.
-        _emit_name_cluster(emit, outputs, array_to_id, id_gen, label="Outputs")
+        _emit_name_cluster(emit, outputs._data, array_to_id, id_gen, label="Outputs")
         _emit_name_cluster(emit, namespace, array_to_id, id_gen, label="Namespace")
 
     return emit.get()
```
There are still a big pile of (mostly unused) `DataWrapper`s and their names that clutter up the graph, but maybe you can figure out a way to filter them out. (PR welcome!)

@kaushikcfd Any clue what's up with those unused `DataWrapper`s that hang out in the namespace?

inducer, line 808: @kaushikcfd Any clue what's up with those unused `DataWrapper`s that hang out in the namespace?

kaushikcfd: Are you pointing at the unused DataWrappers in the DG operator DAG?
If so, it maybe that there are many calls of this pattern: https://github.com/inducer/meshmode/blob/9bf8f9f1b77aeb2f7eaff18aaa5f4e66dbaf2986/meshmode/discretization/connection/direct.py#L207.

I.e. DataWrappers get frozen and again thawed.

kaushikcfd: > Why this change of traversal behavior? This broke pytato.visualization (in the sense that that now only shows the NamedArrays, which isn't useful).

We had to separate out results of a DictOfNamedArrays from the expression inside it. Slightly odd but helps in the a more concrete definition of a LoopyFunction (sub-classed from DictOfNamedArrays) and LoopyFunctionResult (sub-classed from NamedArray).

inducer:
This picked up a  few conflicts.

inducer:

matthiasdiener:
I think this is ready for another review @inducer @kaushikcfd .

pytato/array.py
```diff
@@ -626,7 +626,32 @@ class CountNamed(UniqueTag):
 
 # {{{ dict of named arrays
 
-class DictOfNamedArrays(Mapping[str, Array]):
+class NamedArray(Array):
```
inducer, line 629: Add docstring, include in docs?

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+
+.. autoclass:: LoopyFunctionResult
+
+.. autofunction:: call_loopy
+"""
+
+
+class LoopyFunction(DictOfNamedArrays):
+    """
+    Call to a :mod:`loopy` program.
```
inducer, line 52:
```suggestion
    An array expression node representing a call to an entrypoint in a
    :mod:`loopy` translation unit.
```

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+
+.. autoclass:: LoopyFunction
+
+.. autoclass:: LoopyFunctionResult
+
+.. autofunction:: call_loopy
+"""
+
+
+class LoopyFunction(DictOfNamedArrays):
```
inducer, line 50: Add to docs.

pytato/array.py
```diff
@@ -642,21 +667,28 @@ class DictOfNamedArrays(Mapping[str, Array]):
         This container deliberately does not implement
         arithmetic.
     """
```
inducer, line 669: Updated docs a bit for `NamedArray`.

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+        loopy_arg = self.dict_of_named_arrays.entry_kernel.arg_dict[  # type:ignore
+                self.name]
+        dtype = loopy_arg.dtype
+
+        if isinstance(dtype, np.dtype):
+            return dtype
+        elif isinstance(dtype, NumpyType):
+            return np.dtype(dtype.numpy_dtype)
+        else:
+            raise NotImplementedError(f"Unknown dtype type '{dtype}'")
```
inducer, line 130:
```suggestion
        return loopy_arg.dtype.numpy_dtype
```

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+        if ((self.entrypoint == other.entrypoint)
+             and (self.bindings == other.bindings)
+             and (self.program == other.program)):
+            return True
+        return False
+
+
+class LoopyFunctionResult(NamedArray):
+    """
+    Named array for :class:`LoopyFunction`'s result.
```
inducer, line 100:
```suggestion
    Named array for :class:`LoopyFunction`'s result.
    Inherits from :class:`~pytato.array.NamedArray`.
```

pytato/loopy.py
inducer, line 50:
```suggestion
class LoopyCall(DictOfNamedArrays):
```

pytato/array.py
```diff
@@ -642,21 +667,28 @@ class DictOfNamedArrays(Mapping[str, Array]):
...
+    _mapper_method = "map_dict_of_named_arrays"
 
     def __init__(self, data: Dict[str, Array]):
+        self._named_arrays = {name: NamedArray(self, name)
+                              for name in data}
         self._data = data
```
inducer, line 675: `LoopyFunction` (`LoopyCall`) should call this superclass constructor. (Maybe just allow supplying `_named_arrays`?)

inducer, line 675:
```suggestion
        # kept around mainly for efficient __hash__
        self._data = data
```
(if that's true... otherwise `_data` is pretty redundant with `_named_arrays`.)

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+
+        self.program = program
+        self.bindings = bindings
+        self.entrypoint = entrypoint
+
+        entry_kernel = program[entrypoint]
+
+        self._named_arrays = {name: LoopyFunctionResult(self, name)
+                              for name, lp_arg in entry_kernel.arg_dict.items()
+                              if lp_arg.is_output}
```
inducer, line 70: Should set `_named_arrays` via the superclass constructor.

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+
+    @memoize_method
+    def to_pytato(self, expr: ScalarExpression) -> ScalarExpression:
```
inducer, line 73: Underscore? (I'm guessing this is not meant to be user-facing.)

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+        from pymbolic.mapper.substitutor import make_subst_func
+        return SubstitutionMapper(make_subst_func(self.bindings))(expr)
+
+    @property
+    def entry_kernel(self) -> lp.LoopKernel:
```
inducer, line 78: Underscore? (I'm guessing this is not meant to be user-facing.)

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+def call_loopy(program: "lp.Program",
+               bindings: Dict[str, Union[Array, Number]],
+               entrypoint: Optional[str] = None) -> LoopyFunction:
+    """
+    Operates a general :class:`loopy.Program` on the array inputs as specified
+    by *bindings*.
+
+    Restrictions on the structure of ``program[entrypoint]``:
+
+    * array arguments of ``program[entrypoint]`` should either be either
```
inducer, line 142:
```suggestion
    * array arguments of ``program[entrypoint]`` must either be either
```

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+
+        if isinstance(dtype, np.dtype):
+            return dtype
+        elif isinstance(dtype, NumpyType):
+            return np.dtype(dtype.numpy_dtype)
+        else:
+            raise NotImplementedError(f"Unknown dtype type '{dtype}'")
+
+
+def call_loopy(program: "lp.Program",
```
inducer, line 133: Add to docs.

inducer, line 133: Program -> TranslationUnit

pytato/loopy.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class LoopyFunction(DictOfNamedArrays):
+    """
+    Call to a :mod:`loopy` program.
+    """
+    _mapper_method = "map_loopy_function"
+
+    def __init__(self,
+            program: "lp.Program",
```
inducer, line 57: Program -> TranslationUnit (Attribute should also be renamed)

.github/workflows/ci.yml
```diff
@@ -34,6 +34,7 @@ jobs:
                 python-version: '3.x'
         -   name: "Main Script"
             run: |
+                sed -i "s/loopy.git/loopy.git@kernel_callables_v3-edit2/g" requirements.txt
```
inducer, line 37: File an issue so we remember to point this back to main.

inducer (commented):
Thanks! This looks great. Just a few (mostly cosmetic) things.

pytato/array.py
matthiasdiener: Added. Correct like it is @kaushikcfd ?

pytato/array.py
matthiasdiener: Updated.

pytato/loopy.py
matthiasdiener: Renamed.

pytato/loopy.py
matthiasdiener: added underscore

pytato/loopy.py
matthiasdiener: added underscore

pytato/loopy.py
matthiasdiener: changed

matthiasdiener: added

pytato/loopy.py
matthiasdiener: added

pytato/loopy.py
matthiasdiener: Done (I think).

pytato/array.py
matthiasdiener: Done (I think).

matthiasdiener:
> Thanks! This looks great. Just a few (mostly cosmetic) things.

Thanks! I think I addressed your comments @inducer.

inducer:
> Apply suggestions from code review

:( Please use commit messages that'll still make sense from the perspective of the commit history later on. Those commits won't get squashed away.

pytato/loopy.py
```diff
@@ -0,0 +1,217 @@
...
+    _mapper_method = "map_loopy_function"
+
+    def __init__(self,
+            translation_unit: "lp.TranslationUnit",
+            bindings: Dict[str, Union[Array, Number]],
+            entrypoint: str):
+        entry_kernel = translation_unit[entrypoint]
+        super().__init__({name: LoopyCallResult(self, name)
+                              for name, lp_arg in entry_kernel.arg_dict.items()
+                              if lp_arg.is_output})
```
inducer, line 63: I think these will get "double-wrapped" now, once in `LoopyCallResult` and again in `NamedArray`. I think if you want to pass `NamedArray`s to the superclass, you'll need to make a separate argument.

inducer, line 63: Assuming I'm right about the double-wrapping, how come none of the tests noticed this?

inducer (commented):
Thanks! A few more comments below.

pytato/array.py
```diff
@@ -642,21 +672,29 @@ class DictOfNamedArrays(Mapping[str, Array]):
         This container deliberately does not implement
         arithmetic.
     """
+    _mapper_method = "map_dict_of_named_arrays"
 
     def __init__(self, data: Dict[str, Array]):
+        self._named_arrays = {name: NamedArray(self, name)
+                              for name in data}
+        # kept around mainly for efficient __hash__
         self._data = data
```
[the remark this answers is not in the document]
kaushikcfd: "kept around for efficient `__hash__`", I'm guessing that was for "_data" attribute, then not quite. A NamedArray comprises of a DictOfNamedArray and it's name in it i.e. it's more of a reference in the DictOfNamedArray. However, `_data` is the one which actually contains the array expressions. `NamedArray.expr` might explain the need for this.

The reason that a NamedArray doesn't contain the array expression is because NamedArray shouldn't be anything by itself, as it is always comes in a group. Take for example a `LoopyFunctionResult`, we cannot emit expressions that would only compute one of the outputs of a `LoopyFunction` i.e. each of them is just a reference to in a group of results.

Probably I should add docs highlighting this if you agree with the decision taken here of keeping things centralized at DictOfNamedArrays.

pytato/loopy.py
kaushikcfd: Oh, yep these shouldn't be "double-wrapped'. Didn't get caught in the tests as the mappers for the named array just point back all the operations to it's DictOfNamedArrays. Will fix this.

kaushikcfd: Continuing the discussion at https://github.com/inducer/pytato/pull/20/files#r645290705

pytato/loopy.py
```diff
@@ -112,66 +112,66 @@ def expr(self) -> Array:
...
                 self.name]
-        shape: ShapeType = self.dict_of_named_arrays.to_pytato(  # type:ignore
+        shape: ShapeType = self.dict_of_named_arrays._to_pytato(  # type:ignore
                 loopy_arg.shape)
         return shape
 
     @property
-    def dtype(self) -> np.dtype[Any]:
-        loopy_arg = self.dict_of_named_arrays.entry_kernel.arg_dict[  # type:ignore
+    def dtype(self) -> Any:
```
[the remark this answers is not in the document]
kaushikcfd: We always emit `Any` for dtype.

pytato/loopy.py
```diff
@@ -0,0 +1,217 @@
...
+import loopy as lp
+import pymbolic.primitives as prim
+from typing import Dict, Optional, Any, Union
+from numbers import Number
+from pytato.array import DictOfNamedArrays, Array, ShapeType, NamedArray
+from pytato.scalar_expr import SubstitutionMapper, ScalarExpression
+from pytools import memoize_method
+from pytools.tag import TagsType
+
+__doc__ = """
```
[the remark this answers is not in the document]
matthiasdiener: I think it is in the docs (reference.rst)

pytato/loopy.py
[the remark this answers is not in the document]
kaushikcfd: A decision that was chosen quite early on, and I had inferred that's because to accomodate for "whatever 'np.dtype' can take in.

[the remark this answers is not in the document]
matthiasdiener: (side note - mypy complained about np.dtype[Any])

pytato/array.py
kaushikcfd: The docs are accurate.

pytato/loopy.py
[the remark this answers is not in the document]
kaushikcfd: I think we had forgot to wrap it in numpy dtype: c7e5875

pytato/array.py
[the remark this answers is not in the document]
inducer: Thanks for explaining! I was clearly confused here; there's no redundancy at all. I think a little bit of explanation might be useful here, along the lines of "There are two sets of nodes here. _data contains the expressions used to populate the container, whereas _named_arrays contains array nodes that are used by expressions depending on values in the container."

pytato/loopy.py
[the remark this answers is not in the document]
matthiasdiener: Added a nitpick argument to sphinx which will warn about these

doc/Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+SPHINXOPTS    ?= -n
```
inducer, line 6: There shouldn't be a need. `build_docs` adds this by default:

https://github.com/inducer/ci-support/blob/4edcdff1f7126d437630cdd0a6e2551a4ec5fb59/ci-support.sh#L423-L427

matthiasdiener: Ok, but isn't it convenient for local builds to gave this enabled?

inducer: Then
```suggestion
SPHINXOPTS    ?= -W --keep-going -n
```
?

matthiasdiener: Well, it currently does not build with `-W -n`, at least on my system, and independent of this PR.

inducer: 78b9aec

inducer: (Long story short here: I had disabled doc checking because pre-4.0 sphinx couldn't handle the type annotations here. That commit reenables them.)

pytato/array.py
```diff
@@ -626,38 +619,103 @@ class CountNamed(UniqueTag):
 
 # {{{ dict of named arrays
 
-class DictOfNamedArrays(Mapping[str, Array]):
-    """A container that maps valid Python identifiers
-    to instances of :class:`Array`. May occur as a result
-    type of array computations.
+class NamedArray(Array):
+    """An entry in a :class:`DictOfNamedArrays`. Holds a reference back to the
+    containing instance as well as the name by which *self* is known there.
```
inducer, line 624:
```suggestion
    """An entry in a :class:`AbstractResultWithNamedArrays`. Holds a reference
    back to thecontaining instance as well as the name by which *self* is
    known there.
```

pytato/array.py
```diff
@@ -626,38 +619,103 @@ class CountNamed(UniqueTag):
...
+    @property
+    def dtype(self) -> np.dtype[Any]:
+        return self.expr.dtype
+
+
+class AbstractResultWithNamedArrays(Mapping[str, NamedArray], ABC):
+    r"""An abstract array computation that results in multiple :class:`Array`\ s,
+    each named. The way in which the values of these arrays are computed
+    is determined by concrete subclasses of this class, e.g.
+    :class:`pytato.loopy.LoopyCall`.
```
inducer, line 659:
```suggestion
    :class:`pytato.loopy.LoopyCall` or :class:`DictOfNamedArrays`.
```

inducer:
LGTM. Thanks to both of you for working on this!

PR #31, opened 2021-02-02
MLIR backend

kaushikcfd:


pytato/mlir.py
```diff
@@ -0,0 +1,708 @@
+from __future__ import annotations
```
inducer, line 1: Add copyright header

inducer, line 1: `pytato.target.mlir`? (along with `pytato.target.loopy`? If we do that, please submit a separate move PR with no functional changes)

PR #32, opened 2021-02-04
Modularize pytato.(target|codegen)

kaushikcfd:


pytato/target/__init__.py
```diff
@@ -29,61 +29,60 @@
...
-class PyOpenCLTarget(Target):
-    """A :mod:`pyopencl` code generation target.
+@dataclass(init=True, repr=False, eq=False)
+class BoundProgram:
+    """A wrapper around a :mod:`loopy` kernel for execution.
+
+    .. attribute:: program
 
-    .. attribute:: queue
+        The underlying :class:`loopy.LoopKernel`.
```
inducer, line 67: Still loopy-specific.

pytato/codegen/loopy.py
```diff
@@ -23,55 +23,32 @@
 """
 
 import dataclasses
```
inducer, line 25: Why have both `target.loopy` and `codegen.loopy` (rather than `target.loopy.codegen` maybe)? Shouldn't we gather all the target-specific stuff in one spot?

inducer (commented):
Thanks! Could you please confirm that there are no functional changes in here?

kaushikcfd:
Thanks for taking a look! Sorry the diff is not quite helpful to understand functional changes. Changelog:
1. Cut "codegen.py"'s code into 2 parts: general (codegen/__init__.py) and loopy specific (codegen/loopy.py). No new routines/classes introduced.
2. Condensed the interfaces of `pytato.program` & `pytato.target` into one.
  - Then, separated abstract interfaces (like Target, BoundProgram) into target/__init__.py and loopy specific logic (LoopyTarget, PyOpenCLTarget, BoundPyOpenCLProgram) into target/loopy.py.
    - Notice that we introduced a new class `LoopyTarget`.
3. Renaming to accommodate for (1) and (2).

pytato/target/__init__.py
kaushikcfd: 9730a08

pytato/codegen/loopy.py
kaushikcfd: Agreed. 58dac45.
I think after these set of changes we should also introduce `generate_code` which just takes in a DictOfNamedArrays and a target and then builds a bound program for the target.

inducer:
Problems I noticed:

- Why does `BoundProgram` store a queue? Storing queues leads to synchronization bugs.
- Why is the target for Loopy's PyOpenCL target just called `PyOpenCLTarget`, instead of `LoopyPyOpenCLTarget` (or `PyOpenCLLoopyTarget`)?
- Why did `codegen/__init__.py` become `codegen.py` with no need?

inducer:
Picked up conflicts from the merge of #16.

kaushikcfd:
> Why does BoundProgram store a queue? Storing queues leads to synchronization bugs.

Current master does that too. (bumped that to an issue #37, as it's out of scope of this PR)

Agreed, PyOpenCLTarget is now LoopyPyOpenCLTarget

codegen/__init__.py was unnecessary (current master only contains codegen.py). codegen.py now only contains common utils routines that's reusable across all target. Target-specific codegen has moved to `pytato/target/loopy/codegen.py`.

inducer:
Thanks!

PR #38, opened 2021-03-05
Reductions

kaushikcfd:


kaushikcfd:
Hah. Tests are failing because reductions are broken in kernel callables. Turning to draft.

kaushikcfd:
@inducer: These tests pass too and could be reviewed.

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+        if i not in reduction_axes:
+            new_shape.append(axis_len)
+
+    return tuple(new_shape), reduction_axes
+
+
+def _get_reduction_indices_bounds(shape: ShapeType,
+        axes: Tuple[int, ...]) -> Tuple[
+                List[ScalarExpression],
+                Mapping[str, Tuple[ScalarExpression, ScalarExpression]]]:
```
inducer, line 1880:
```suggestion
        axes: Tuple[int, ...]) -> Tuple[
                Sequence[ScalarExpression],
                Mapping[str, Tuple[ScalarExpression, ScalarExpression]]]:
```

pytato/array.py
```diff
@@ -162,7 +163,7 @@
 import operator
 from typing import (
         Optional, Callable, ClassVar, Dict, Any, Mapping, Iterator, Tuple, Union,
-        Protocol, Sequence, cast, TYPE_CHECKING)
+        Protocol, Sequence, cast, TYPE_CHECKING, List)
```
inducer, line 166: Prefer `Sequence` to `List`.

pytato/scalar_expr.py
```diff
@@ -155,4 +121,53 @@ def substitute(expression: Any, variable_assigments: Mapping[str, Any]) -> Any:
...
+            neutral = math.inf
+        elif self.value == "sum":
+            neutral = 0
+        elif self.value == "product":
+            neutral = 1
+        else:
+            raise NotImplementedError(f"Unknown reduction op {self}.")
+
+        # https://github.com/python/mypy/issues/3186
+        return neutral   # type: ignore
```
inducer, line 144: Can you make these classes (instead of enum values)?

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+
+
+def make_index_lambda(namespace: Namespace,
+        expression: Union[str, ScalarExpression],
+        bindings: Dict[str, Array],
+        shape: ShapeType,
+        dtype: Any) -> IndexLambda:
+    if isinstance(expression, str):
+        raise NotImplementedError("Sorry the developers were too lazy to implement"
+                " a parser.")
```
inducer, line 1826:
```suggestion
        raise NotImplementedError
```
(no need to apologize)

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
 
 # }}}
 
+
+# {{{ make_index_lambda
+
+INDEX_RE = re.compile("_r?(0|([1-9][0-9]*))")
```
inducer, line 1816: I feel like this is misleading: These are reduction indices.

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+        shape: ShapeType,
+        dtype: Any) -> IndexLambda:
+    if isinstance(expression, str):
+        raise NotImplementedError("Sorry the developers were too lazy to implement"
+                " a parser.")
+
+    # {{{ sanity checks
+
+    from pytato.scalar_expr import get_dependencies
+    unknown_dep = (get_dependencies(expression)
```
inducer, line 1831: `get_dependencies` should maybe have a mode where it doesn't report the expression dependent on `_r0` for `sum(_r0, f(_r0))`.

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+                   - set(namespace) - set(bindings))
+    for dep in unknown_dep:
+        if not INDEX_RE.fullmatch(dep):
```
inducer, line 1834: This exclusion is hacky IMO, see above.

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+
+# }}}
+
+
+# {{{ reductions
+
+def _preprocess_reduction_axes(
+        shape: ShapeType,
+        reduction_axes: Optional[Union[int, Tuple[int]]]
+        ) -> Tuple[ShapeType, Tuple[int, ...]]:
```
inducer, line 1852: Add a docstring to explain what this is for.

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+                scalar_expr.ReductionOp.SUM,
+                redn_bounds),
+            {"in": a},
+            new_shape,
+            a.dtype)
+
+
+def amax(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+    """
+    Sums array *a*'s elements along the *axis* axes.
```
inducer, line 1923: *cough*

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+                scalar_expr.ReductionOp.MAX,
+                redn_bounds),
+            {"in": a},
+            new_shape,
+            a.dtype)
+
+
+def amin(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+    """
+    Sums array *a*'s elements along the *axis* axes.
```
inducer, line 1944: *cough*

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+
+    :arg axis: The axes along which the elements are to be sum-reduced.
```
inducer, line 1946: > sum-reduced.

*cough*

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+            {"in": a},
+            new_shape,
+            a.dtype)
+
+
+def amax(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+    """
+    Sums array *a*'s elements along the *axis* axes.
+
+    :arg axis: The axes along which the elements are to be sum-reduced.
```
inducer, line 1925: > sum-reduced.

*cough*

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+
+    return indices, redn_bounds
+
+
+def sum(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+    """
+    Sums array *a*'s elements along the *axis* axes.
+
+    :arg axis: The axes along which the elements are to be sum-reduced.
+        Defaults to all axes of the input arrays.
```
inducer, line 1905:
```suggestion
        Defaults to all axes of the input array.
```

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+            new_shape,
+            a.dtype)
+
+
+def amax(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+    """
+    Sums array *a*'s elements along the *axis* axes.
+
+    :arg axis: The axes along which the elements are to be sum-reduced.
+        Defaults to all axes of the input arrays.
```
inducer, line 1926:
```suggestion
        Defaults to all axes of the input array.
```

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+            new_shape,
+            a.dtype)
+
+
+def amin(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+    """
+    Sums array *a*'s elements along the *axis* axes.
+
+    :arg axis: The axes along which the elements are to be sum-reduced.
+        Defaults to all axes of the input arrays.
```
inducer, line 1947:
```suggestion
        Defaults to all axes of the input array.
```

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+            bindings=bindings,
+            shape=shape,
+            dtype=dtype)
+
+# }}}
+
+
+# {{{ reductions
+
+def _preprocess_reduction_axes(
```
inducer, line 1849:
```suggestion
def _normalize_reduction_axes(
```
maybe?

pytato/array.py
```diff
@@ -1804,4 +1810,155 @@ def ones(namespace: Namespace, shape: ConvertibleToShape, dtype: Any = float,
...
+    indices, redn_bounds = _get_reduction_indices_bounds(a.shape, axes)
+
+    return make_index_lambda(a.namespace,
+            scalar_expr.Reduce(
+                prim.Subscript(prim.Variable("in"), tuple(indices)),
+                scalar_expr.ReductionOp.SUM,
+                redn_bounds),
+            {"in": a},
+            new_shape,
+            a.dtype)
```
inducer, line 1918: `sum`, `amax` and `amin` share a lot of code. Maybe reuse?

pytato/scalar_expr.py
```diff
@@ -155,4 +121,53 @@ def substitute(expression: Any, variable_assigments: Mapping[str, Any]) -> Any:
 # }}}
 
 
+class ReductionOp(Enum):
+    MAX = "max"
+    MIN = "min"
+    SUM = "sum"
+    PRODUCT = "product"
```
inducer, line 128: Product isn't supported by the rest of this PR.

pytato/scalar_expr.py
```diff
@@ -155,4 +121,53 @@ def substitute(expression: Any, variable_assigments: Mapping[str, Any]) -> Any:
...
+
+    def __post_init__(self) -> None:
+        self.neutral_element = self.neutral_element or self.op.neutral_element
+
+    def __hash__(self) -> int:
+        return hash((self.inner_expr,
+                self.op,
+                tuple(self.bounds.keys()),
+                tuple(self.bounds.values()),
+                self.neutral_element))
```
inducer, line 163: Overload `__getinitargs__` instead?

pytato/scalar_expr.py
```diff
@@ -155,4 +121,53 @@ def substitute(expression: Any, variable_assigments: Mapping[str, Any]) -> Any:
...
+
+    def __str__(self) -> str:
+        bounds_expr = " and ".join(f"{lb}<={key}<{ub}"
+                for key, (lb, ub) in self.bounds.items())
+        bounds_expr = "{" + bounds_expr + "}"
+        return (f"{self.op.value}({bounds_expr}, {self.inner_expr},"
+                f" {self.neutral_element})")
```
inducer, line 170: Why not implement this as part of the stringify-mapper?

pytato/scalar_expr.py
```diff
@@ -72,60 +74,24 @@ def parse(s: str) -> ScalarExpression:
 # {{{ mapper classes
 
 class WalkMapper(WalkMapperBase):
-
-    def map_reduction(self, expr: Reduction, *args: Any, **kwargs: Any) -> None:
-        if not self.visit(expr, *args, **kwargs):
-            return
-
-        self.rec(expr.expr, *args, **kwargs)
+    pass
```
inducer, line 77: Shouldn't these mappers support `Reduce`?

pytato/target/loopy/codegen.py
```diff
@@ -412,7 +369,8 @@ def map_index_lambda(self, expr: IndexLambda,
 
 # {{{ inlined expression gen mapper
 
-INDEX_RE = re.compile("_(0|([1-9][0-9]*))")
+ELWISE_INDEX_RE = re.compile("_(0|([1-9][0-9]*))")
+REDN_INDEX_RE = re.compile("_r(0|([1-9][0-9]*))")
```
inducer, line 373:
```suggestion
REDUCTION_INDEX_RE = re.compile("_r(0|([1-9][0-9]*))")
```

pytato/target/loopy/codegen.py
```diff
@@ -713,7 +714,7 @@ def generate_loopy(result: Union[Array, DictOfNamedArrays, Dict[str, Array]],
         state.results[expr] = StoredResult(name, expr.ndim, frozenset([insn_id]))
 
     return target.bind_program(
-            program=state.program,
+            program=lp.make_reduction_inames_unique(state.program),
```
inducer, line 717: It's easy to do that a priori? (e.g. via a or the `var_name_gen`) Why require another AST traversal?

pytato/target/loopy/codegen.py
```diff
@@ -470,6 +428,49 @@ def map_call(self, expr: prim.Call,
 
         return super().map_call(expr, expr_context)
 
+    def map_reduce(self, expr: scalar_expr.Reduce,
+            expr_context: LoopyExpressionContext) -> ScalarExpression:
+        from pymbolic.mapper.substitutor import make_subst_func
+        from loopy.symbolic import Reduction
```
inducer, line 434:
```suggestion
        from loopy.symbolic import Reduction as LoopyReduction
```
maybe, for clarity?

pytato/target/loopy/codegen.py
```diff
@@ -470,6 +428,49 @@ def map_call(self, expr: prim.Call,
...
+        state = expr_context.state
+
+        unique_names_mapping = {
+                old_name: prim.Variable(state.var_name_gen(f"_pt_{expr.op.value}")
+                                        + old_name)
+                for old_name in expr.bounds}
+        inner_expr = self.rec(expr.inner_expr, expr_context)
+        inner_expr = SubstitutionMapper(
+                make_subst_func(unique_names_mapping))(inner_expr)
+        inner_expr = Reduction(parse_reduction_op(expr.op.value),
```
inducer, line 445: > `expr.op.value`

Maybe better to maintain an explicit mapping?

pytato/target/loopy/codegen.py
```diff
@@ -470,6 +428,49 @@ def map_call(self, expr: prim.Call,
 
         return super().map_call(expr, expr_context)
 
+    def map_reduce(self, expr: scalar_expr.Reduce,
```
inducer, line 431: `Reduce` includes `neutral_element`, however this is ignored here?

pytato/scalar_expr.py
```diff
@@ -155,4 +121,53 @@ def substitute(expression: Any, variable_assigments: Mapping[str, Any]) -> Any:
...
+        # https://github.com/python/mypy/issues/3186
+        return neutral   # type: ignore
+
+
+@dataclass
+class Reduce(prim.Expression):
+    inner_expr: ScalarExpression
+    op: ReductionOp
+    bounds: Mapping[str, Tuple[ScalarExpression, ScalarExpression]]
+    neutral_element: Optional[ScalarExpression] = None
```
inducer, line 152: Right now, `neutral_element` is redundantly represented. Is that intentional?

inducer (commented):
Thanks!

inducer:

pytato/array.py
matthiasdiener: Done

pytato/array.py
matthiasdiener: Done

pytato/array.py
matthiasdiener: Fixed.

pytato/array.py
matthiasdiener: Fixed.

pytato/array.py
matthiasdiener: Fixed.

pytato/array.py
matthiasdiener: Fixed.

pytato/scalar_expr.py
matthiasdiener: Done in https://github.com/inducer/pytato/pull/38/commits/f2e952e91e2def97bd26978c833688c5c4ffbfbe (but not sure if thats what you had in mind)

pytato/scalar_expr.py
matthiasdiener: added in https://github.com/inducer/pytato/pull/38/commits/f2e952e91e2def97bd26978c833688c5c4ffbfbe

pytato/array.py
matthiasdiener: renamed to REDUCTION_INDEX_RE

pytato/array.py
matthiasdiener: refactored in https://github.com/inducer/pytato/pull/38/commits/f2e952e91e2def97bd26978c833688c5c4ffbfbe

pytato/array.py
matthiasdiener: Done.

pytato/scalar_expr.py
matthiasdiener: Hmm, not sure I understand.

kaushikcfd: pymbolic.primtives.Expression can have __getinitargs__ that takes care of the hash (https://github.com/inducer/pymbolic/blob/0fdfd8c8276cafc64bcc169c1b8c0a7fe2e4afe9/pymbolic/primitives.py#L550-L551).

matthiasdiener: Removing the `__hash__` function results in the following error:

```
  File "examples/demo.py", line 17, in <module>
    prg = pt.generate_loopy(result, cl_device=queue.device)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/target/loopy/codegen.py", line 745, in generate_loopy
    insn_id = add_store(name, expr, cg_mapper(expr, state), state, cg_mapper)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/transform.py", line 83, in __call__
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/transform.py", line 80, in rec
    return method(expr, *args, **kwargs)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/target/loopy/codegen.py", line 338, in map_index_lambda
    if expr in state.results:
  File "/Users/mdiener/Work/emirge_lazy/miniforge3/envs/ceesd_lazy/lib/python3.8/site-packages/pytools/__init__.py", line 698, in wrapper
    result = function(obj, *args, **kwargs)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/array.py", line 470, in __hash__
    return hash(tuple(attrs))
TypeError: unhashable type: 'Reduce'
```

kaushikcfd: Is it after defining `Reduce.__getinitargs__` method?

matthiasdiener: Yes

kaushikcfd: Hmm. not sure what's happening there. Looking at the definitions of `Expression.__getinitargs__`, `Expression.get_hash` it seemed like that should have worked, but doesn't seem to be the case.

pytato/scalar_expr.py
matthiasdiener: Implemented support for WalkMapper, no clue about the others.

pytato/target/loopy/codegen.py
kaushikcfd: Not quite. We cache array-expressions -> loopy-expressions in CodeGenMapper. Teaching the caching mechanism about not caching an array expression containing a reduce node is less intuitive than just calling `make_reduction_inames_unique`.

matthiasdiener:
@inducer @kaushikcfd For the remaining unresolved comments, I'm unsure what to do. Please add some more context. 

@inducer: Please take a look at the resolved comments.

pytato/scalar_expr.py
matthiasdiener: Which mapper is that @inducer?

pytato/target/loopy/codegen.py
matthiasdiener: I'm not sure what you mean here @inducer

pytato/array.py
kaushikcfd: This could be ignored. This codepath doesn't exist anymore as namespace doesn't exist anymore. Resolving.

pytato/target/loopy/codegen.py
kaushikcfd: We are invoking loopy's parse_reduction_op to get the reduction Op. I think @inducer suggests we shouldn't piggy back on that since that isn't quite a public interface.

pytato/target/loopy/codegen.py
kaushikcfd: @matthiasdiener: By this @inducer means we are generating incorrect loopy kernel for the non-default neutral element. I.e. If we had supported the `initial` argument of [np.sum](https://numpy.org/doc/stable/reference/generated/numpy.sum.html) we would get incorrect results.

pytato/target/loopy/codegen.py
matthiasdiener: Fixed in https://github.com/inducer/pytato/pull/38/commits/4ca5eb69fba8f4a8ebbf8851ef2c22a289dac68d, I think (the mapping between pytato reduction ops and loopy reduction ops seems to be 1:1, at least for the ops currently implemented in pytato)

kaushikcfd: Thanks, with 8644409 I think we can resolve this.

pytato/target/loopy/codegen.py
matthiasdiener: Is it worth supporting `neutral_element` at all (esp. considering we are not supporting a couple of other optional arguments to `np.sum` etc. as well)?

kaushikcfd: I also don't see an immediate need for neutral element, but not too difficult to generate the correct code. While generating the code we would have to overload the corresponding loopy reduction type and overload the neutral_element method to return the correct expression.

pytato/scalar_expr.py
```diff
@@ -161,51 +161,42 @@ def distribute(expr: Any, parameters: Set[Any] = set(),
 
 # }}}
 
+from loopy.library.reduction import (SumReductionOperation, MinReductionOperation, MaxReductionOperation, ProductReductionOperation, ReductionOperation)
```
kaushikcfd: @matthiasdiener: The plan here was that unless loopy codegen is triggered we shouldn't depend on :mod:`loopy` thereby introducing our reduction types.

matthiasdiener: Yeah, I was afraid of that :-(

kaushikcfd: I think what we can do is: in the map_reduce of CodeGenMapper, we derive from the loopy reduction type to emit the correct neutral element expression.

matthiasdiener: I removed this code.

pytato/target/loopy/codegen.py
matthiasdiener: I removed the neutral_element code in https://github.com/inducer/pytato/pull/38/commits/6268550bb740378e099dcc60a18cc8ece930e58a

pytato/scalar_expr.py
matthiasdiener: Removed this representation.

pytato/array.py
matthiasdiener: I'm not sure how to implement this.

matthiasdiener:
I think all comments have been addressed, or I left a specific question @inducer .

pytato/array.py
kaushikcfd: One way to do this would be to introduce a flag in pytato.scalar_expr.DependencyMapper.__init__ that would ignore deps. if they matched with the INDEX_RE. And then allowing the flag to be set by pytato.scalar_expr.get_dependencies.

matthiasdiener: Should it ignore deps that match INDEX_RE just in map_reduce, or in the other map_ functions too?

kaushikcfd: I would do it in map_variable, because the names of `pymbolic.primitives.Variable` is what is propagated upwards.

matthiasdiener: What do you think of https://github.com/inducer/pytato/pull/38/commits/8cf96b222ed42c5b9734108dbdb2233d122898aa ?

kaushikcfd: Yep, I think that resolves it. (Thank you!)

matthiasdiener:
> Probably better to use `super().map_variable(expr)`.

Good point, fixed.

matthiasdiener:
> Since this mapper is pytato-specific we could simplify it as taking `ignore_index_lambda_indices` which hardcodes to ignoring any dependencies that match `re.compile("_(r?)(0|([1-9][0-9]*))")` rather than taking in a regex. Less general than your version but for the pytato-verse should be sufficient.

My thinking here was that we might need to ignore other types of dependencies in the future, but I'm not sure how likely that is.

inducer:
An additional point on this is that inheriting a new `DependencyMapper` is really cheap. So IMO it's totally fine to leave things generic in the base `DependencyMapper` but then have a subclass with a really narrow use case.

matthiasdiener:
> Since this mapper is pytato-specific we could simplify it as taking ignore_index_lambda_indices which hardcodes to ignoring any dependencies that match re.compile("_(r?)(0|([1-9][0-9]*))") rather than taking in a regex. Less general than your version but for the pytato-verse should be sufficient.

> An additional point on this is that inheriting a new `DependencyMapper` is really cheap. So IMO it's totally fine to leave things generic in the base `DependencyMapper` but then have a subclass with a really narrow use case. (I was tempted to suggest, oh, pass in a callback, but I've consistently regretted similar design decisions in the past.)

I think these are good points and I've hardcoded the `re` in the most recent commit (https://github.com/inducer/pytato/pull/38/commits/4a3eb70a503268d5864fc6995889e3ac2101030d). This is ready for another review @inducer @kaushikcfd .

pytato/target/loopy/codegen.py
inducer: Hmm... I guess I'm not sure I understand why the reduction inames coming out of `InlinedExpressionGenMapper.map_reduce` aren't yet globally unique.

inducer (commented):
Thanks for working on this! A few comments below. @kaushikcfd, please also take a look at 5b382bd to see if you like that.

pytato/scalar_expr.py
```diff
@@ -191,4 +186,42 @@ def distribute(expr: Any, parameters: Set[Any] = set(),
...
+        A mapping from reduction inames to tuples ``(lower_bound, upper_bound)``
+        identifying half-open bounds intervals.  Must be hashable.
+    """
+    inner_expr: ScalarExpression
+    op: str
+    bounds: Mapping[str, Tuple[ScalarExpression, ScalarExpression]]
+
+    def __init__(self, inner_expr, op, bounds):
+        self.inner_expr = inner_expr
+        self.op = op
```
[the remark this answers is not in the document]
kaushikcfd: I am fine with this. But this is uglier than using Enums which used the type-system for giving us these sort of checks.

test/test_codegen.py
```diff
@@ -595,6 +595,25 @@ def _get_rand_with_nans(shape):
     np.testing.assert_allclose(y, np_func(x1_in, x2_in), rtol=1e-6)
 
 
+@pytest.mark.parametrize("axis", (None, 1, 0))
+@pytest.mark.parametrize("redn", ("sum", "amax", "amin"))
+def test_reductions(ctx_factory, axis, redn):
```
kaushikcfd, line 600: Note to self: Add a test with parametrized shapes.

pytato/array.py
```diff
@@ -84,6 +84,7 @@
 .. autofunction:: where
 .. autofunction:: maximum
 .. autofunction:: minimum
+.. autofunction:: sum
```
kaushikcfd, line 87: Note to self: add prod, max, min to docs.

pytato/array.py
```diff
@@ -839,7 +841,11 @@ def dtype(self) -> np.dtype[Any]:
 
     @property
     def shape(self) -> ShapeType:
-        common_axis_len = sum(ary.shape[self.axis] for ary in self.arrays)
+        from functools import reduce
+        import operator
+
+        common_axis_len = reduce(operator.add, (ary.shape[self.axis]
+                                                for ary in self.arrays))
```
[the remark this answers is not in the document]
kaushikcfd: Yep very weird, unsure why I did this :roll_eyes:. Should be equivalent. I'll revert these.

pytato/scalar_expr.py
```diff
@@ -80,60 +81,42 @@ def parse(s: str) -> ScalarExpression:
...
-                        f"reduction iname {iname} can only be renamed"
-                        " to another iname")
-            new_inames.append(new_iname.name)
 
-        return Reduction(expr.operation,
-                tuple(new_inames),
-                self.rec(expr.expr, *args, **kwargs),
-                allow_simultaneous=expr.allow_simultaneous)
+class IdentityMapper(IdentityMapperBase):
+    pass
```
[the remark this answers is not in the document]
kaushikcfd: These were mappers for handling loopy's reduction nodes. I will add the pytato ones here.

pytato/scalar_expr.py
```diff
@@ -80,60 +81,42 @@ def parse(s: str) -> ScalarExpression:
...
-                tuple(new_inames),
-                self.rec(expr.expr),
-                allow_simultaneous=expr.allow_simultaneous)
+    pass
 
 
 class DependencyMapper(DependencyMapperBase):
 
-    def map_reduction(self, expr: Reduction,
+    def __init__(self, **kwargs: bool) -> None:
```
kaushikcfd, line 103: Note to self: take in a kwarg for ignoring the deps.

kaushikcfd (commented):
@inducer, @matthiasdiener: Thanks for the help here. There are few minor things that I will handle. And then you guys can take a look.

@matthiasdiener: I think I will squash some of your commits (the authorship would be preserved) so that in the future git-blame, git-bisect are helpful. Hope you don't mind this.

pytato/array.py
matthiasdiener: done

test/test_codegen.py
matthiasdiener: Done.

pytato/array.py
kaushikcfd: Thank you! :)

pytato/target/loopy/codegen.py
kaushikcfd: Could you please take a look at 6dd3a2503c4c3c172e3505667c5e08edcc321169 and see if answers your concern.

pytato/target/loopy/codegen.py
```diff
@@ -463,6 +432,63 @@ def map_call(self, expr: prim.Call,
...
+        rule_name = state.var_name_gen("_pt_subst")
+        args = tuple(state.var_name_gen(f"i{i}")
+                     for i in range(expr_context.num_indices))
+
+        subst_rule = lp.SubstitutionRule(
+                rule_name,
+                args,
+                SubstitutionMapper(make_subst_func({
+                    f"_{i}": prim.Variable(args[i])
+                    for i in range(expr_context.num_indices)}))(inner_expr))
```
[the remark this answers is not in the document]
kaushikcfd: Does this help: c9bfb61cc5297cdaff5e811b0fc8578171be1f9a?

kaushikcfd (commented):
@inducer: This is good from my end too. For the tests to pass I *think* we might need https://github.com/inducer/ci-support/pull/17.

I almost had to do nothing (thanks @matthiasdiener :) ), just added some more not-implemented errors for things that we might need in the distant future, and added some more context in the docs.

pytato/scalar_expr.py
kaushikcfd: Done.

test/test_codegen.py
kaushikcfd: Sorry for being vague here, I was hinting towards having some tests with parametric shapes, but that can wait. Stepped over it via a NotImplementedError

pytato/target/loopy/codegen.py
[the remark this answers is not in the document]
kaushikcfd: Done.

pytato/array.py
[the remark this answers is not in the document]
kaushikcfd: Good news: There was actually a reason (which is better than, "I have started introducing rewriting things for no reason. :p") We have `def sum(x, ...)` in the scope now and hence polluting the namespace. Added a comment explaining this.

pytato/array.py
```diff
@@ -1887,8 +1887,8 @@ def _get_reduction_indices_bounds(shape: ShapeType,
             indices.append(prim.Variable(f"_{n_out_dims}"))
             n_out_dims += 1
 
-    from pyrsistent import pmap
-    return indices, pmap(redn_bounds)  # type: ignore
+    from pyrsistent import pmap  # type: ignore
```
kaushikcfd, line 1890: I was shooting for inducer/ci-support#17 to fix it.

kaushikcfd: (Or do you think that's not the right way, then please lmk)

matthiasdiener: oooh, sorry!

kaushikcfd: According to https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports, we could also solve it by passing `--install-types` instead of installing all the package. Anyone has a favorite?

matthiasdiener: I think https://github.com/inducer/ci-support/pull/17 is the better solution, if it fixes this issue.

pytato/array.py
```diff
@@ -840,7 +845,10 @@ def dtype(self) -> np.dtype[Any]:
 
     @property
     def shape(self) -> ShapeType:
-        common_axis_len = sum(ary.shape[self.axis] for ary in self.arrays)
+        from functools import reduce
+        # avoiding 'sum' to avoid collision with 'pt.sum'
+        common_axis_len = reduce(lambda a, b: a+b, (ary.shape[self.axis]
+                                                    for ary in self.arrays))
```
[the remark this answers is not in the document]
kaushikcfd: Thx!

pytato/array.py
inducer: https://github.com/inducer/pytato/pull/38, 71e6238

inducer: Hmm... it doesn't complain about missing types, but it seems to not be liking the types it *is* finding.

kaushikcfd: Yep, might need to revert this: https://github.com/inducer/pytato/pull/38/commits/79b4f65fd08fc355f508740b08b651554e883c2e

The reason might be that reduction_bounds cannot take in array expressions but mypy isn't able to simplify [this not implemented error.](https://github.com/kaushikcfd/pytato/blob/5c30ee5c0bca58786d85b0e815ad13049e94e162/pytato/array.py#L1875-L1879)

kaushikcfd: I had done the same mistake earlier. Will revert it and get resolve this.

kaushikcfd: :crossed_fingers: [This](85c239f) should pass all the tests.

inducer:
The subst rule was not necessary. 30fe0ab

inducer (commented):
Thanks for working on this. I'll apply these suggestions.

kaushikcfd:
Aah that's right. In the codegen mapper to "index" differently into the substitution we directly make changes to the "_0", "_1" expressions and should do it. Thanks for the simplification!

pytato/scalar_expr.py
```diff
@@ -80,72 +82,85 @@ def parse(s: str) -> ScalarExpression:
...
-                allow_simultaneous=expr.allow_simultaneous)
 
+IDX_LAMBDA_RE = re.compile("_r?(0|([1-9][0-9]*))")
 
-class DependencyMapper(DependencyMapperBase):
 
-    def map_reduction(self, expr: Reduction,
+class DependencyMapper(DependencyMapperBase):
+    def __init__(self, *,
+                 include_idx_lambda_indices: bool = False,
```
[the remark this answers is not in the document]
kaushikcfd: I vote to change to True.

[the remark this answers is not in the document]
kaushikcfd: We'll also need to fix the invocation in make_index_lambda.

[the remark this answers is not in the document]
inducer: e5fdb64

pytato/target/loopy/codegen.py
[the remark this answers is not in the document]
inducer: 30fe0ab

PR #43, opened 2021-03-10
Implements a naive data wrapper garbage collection

kaushikcfd:
/cc @xywei @inducer 

One thing we miss here is that we would no longer support `ns["x"]`, but I believe that was a bad interface to begin with.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -269,15 +269,16 @@ class Namespace(Mapping[str, "Array"]):
                 self, expr=var_ref, shape=value.shape,
                 dtype=value.dtype)
 
-    def remove_out_of_ref_data_wrappers(self) -> None:
+    def remove_out_of_scope_data_wrappers(self) -> None:
         import sys
         data_wrappers = {name
                          for name, ary in self.items()
-                         if isinstance(ary, DataWrapper)}
-        out_of_refs = {name
-                       for name in data_wrappers
-                       if sys.getrefcount(self[name]) <= 2}
-        for k in out_of_refs:
+                         if (isinstance(ary, DataWrapper)
+                             and name.startswith("_pt"))}
+        out_of_scope_dws = {name
+                            for name in data_wrappers
+                            if sys.getrefcount(self[name]) <= 2}
+        for k in out_of_scope_dws:
             del self._symbol_table[k]
 
 # }}}
@@ -1698,7 +1699,7 @@ def make_data_wrapper(namespace: Namespace,
     :param shape:      optional shape of the array, inferred from *data* if not given
     :param tags:       implementation tags
     """
-    namespace.remove_out_of_ref_data_wrappers()
+    namespace.remove_out_of_scope_data_wrappers()
 
     if name is None:
         name = namespace.name_gen("_pt_data")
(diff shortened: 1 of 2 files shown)


test/test_pytato.py
```diff
@@ -177,6 +177,17 @@ def test_reshape_input_validation():
...
 
+def test_gc():
+    def f(ns):
+        x = pt.make_data_wrapper(ns, np.random.rand(10, 10), name="x")
+        del x
+    ns = pt.Namespace()
+    f(ns)
+    pt.make_data_wrapper(ns, np.random.rand(10, 10), name="x1")
+    x2 = pt.make_data_wrapper(ns, np.random.rand(10, 10), name="x2")  # noqa: F841
+    assert list(ns.keys()) == ["x2"]
```
kaushikcfd, line 188: Python implementation specific regression. Is this a very bad idea?

pytato/array.py
```diff
@@ -269,6 +269,17 @@ def ref(self, name: str) -> Array:
...
+    def remove_out_of_ref_data_wrappers(self) -> None:
+        import sys
+        data_wrappers = {name
+                         for name, ary in self.items()
+                         if isinstance(ary, DataWrapper)}
+        out_of_refs = {name
+                       for name in data_wrappers
+                       if sys.getrefcount(self[name]) <= 2}
+        for k in out_of_refs:
+            del self._symbol_table[k]
```
xywei, line 281: How about running gc only on internally named data wrappers? This would not change the current interface of user-defined names.

kaushikcfd: I like that even better! (Thanks!) We could just drop all the entities in Namespace that begin with `__pt` and no-one else is referring to them.

pytato/array.py
```diff
@@ -1687,6 +1699,8 @@ def make_data_wrapper(namespace: Namespace,
     :param shape:      optional shape of the array, inferred from *data* if not given
     :param tags:       implementation tags
     """
+    namespace.remove_out_of_scope_data_wrappers()
```
inducer, line 1702: I'm not a fan of this being implicitly triggered, at least not from here. (Maybe `generate_loopy`? I could be talked into that.)

pytato/array.py
```diff
@@ -269,6 +269,18 @@ def ref(self, name: str) -> Array:
...
 
+    def remove_out_of_scope_data_wrappers(self) -> None:
+        import sys
+        data_wrappers = {name
+                         for name in self
+                         if (isinstance(self[name], DataWrapper)
+                             and name.startswith("_pt"))}
+        out_of_scope_dws = {name
+                            for name in data_wrappers
+                            if sys.getrefcount(self[name]) <= 2}
```
inducer, line 280: I'm not a fan of using the Python refcount. (Why "2" anyhow? The dict and the interpreter stack after evaluating `self[name]`?) Instead, we should just walk all expressions reachable from the namespace and, optionally, some user-supplied nodes. Whatever isn't referred to by that should get cleared.

inducer (commented):
Thanks! A few requests below.

inducer:
Unsubscribing... @-mention or request review once it's ready for another look or needs attention.

kaushikcfd:
The same feature is implemented via #56 and in a much cleaner way.

PR #49, opened 2021-03-12
Default to enabling return_dict when dictionary is passed to generate_loopy

inducer:
Draft because:

- [x] Requires/lives on top of #48 

cc @matthiasdiener


kaushikcfd (approved):
Thanks, LGTM!

PR #50, opened 2021-03-14
[Bug Fix]: Index lambda scoping principles

kaushikcfd:
The included test fails on `main`.

### DIFF
--- a/test/test_codegen.py
+++ b/test/test_codegen.py
@@ -518,6 +518,22 @@ def test_passsing_bound_arguments_raises(ctx_factory):
         evt, (out2,) = prg(queue, x=np.random.rand(10))
 
 
+def test_namespace_variables_in_idx_lambda_exprs(ctx_factory):
+    from pymbolic import parse
+    from numpy.random import default_rng
+
+    queue = cl.CommandQueue(ctx_factory())
+    rng = default_rng()
+    x_in = rng.random(size=(10, 4))
+
+    ns = pt.Namespace()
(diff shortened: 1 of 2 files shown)


test/test_codegen.py
```diff
@@ -518,6 +518,22 @@ def test_passsing_bound_arguments_raises(ctx_factory):
...
+    from pymbolic import parse
+    from numpy.random import default_rng
+
+    queue = cl.CommandQueue(ctx_factory())
+    rng = default_rng()
+    x_in = rng.random(size=(10, 4))
+
+    ns = pt.Namespace()
+    pt.make_placeholder(ns, name="x", shape=(10, 4), dtype=float)
+    y = pt.IndexLambda(ns, parse("2*x[_0, _1]"), (10, 4), float)
```
inducer, line 531: Technically, there are two ambiguities here:

- Using names from the namespace directly in `IndexLambda` expressions. (vs. forcing them into bindings)
- Referring to array expressions with their name (e.g. by using their name in an `IndexLambda` that uses them) vs. the `Array` object defining them.

I think canonicalizing them away is a good idea (I'm guessing at least part of this is what #51 is about... I've only seen the headline), but I think there's a legitimate discussion to be had over whether we should allow them in the first place. I'd be happy to hear your comments here, but I'll also put this on the agenda for Tuesday.

inducer (commented):
Thanks!

inducer:

kaushikcfd:
Addressing input arguments by their names in IndexLambda would be illegal once #56 goes in. Closing this in favor of that.

Issue #55, opened 2021-03-18
Communication nodes in pytato


matthiasdiener:
Form the discussion with @inducer (see also #54):

- Should we have separate send and receive nodes?
- Joined communication buffers?
- Is there overlap with other types of operations (e.g., visualizing or checkpointing)?

inducer:
If we make separate send and receive nodes (which feels like the right approach to me), a complication that arises is that a "send node" is a sink, i.e. it doesn't have a "result", and nobody (at least on the local node) depends on its completion. That means, it has nowhere obvious to go within the DAG. We *could* just randomly shove send nodes into the `Namespace` and be done with it, but that feels inelegant. Another option would be to attach (as "fake dependencies", in a way) to receive nodes (since, at least in DG, much of our communication is paired sends and receives, though I can think of counterexamples... think 1D advection, which only needs to communicate in one direction). I'm not enthusiastic about either of the two options.

@kaushikcfd @rwerger @vincentwells, I want to make sure that you're part of this discussion early on, to make sure we pick a direction that you're OK with.

kaushikcfd:
1. Receive nodes fit naturally within pytato as they could be thought of as data wrappers but initialized via callbacks instead of concrete data.
2. Send nodes in a namespace are weird: as "sends" are a characteristic of a computation and not the namespace (i.e. how would we realize which expression to tie the send to if multiple expressions are dangling from a namespace).
I propose realizing "send" nodes via tags:
```python
rank = comm.get_rank()
size = comm.get_size()
x = pt.make_placeholder(shape=(10,), dtype=float)
left_bnd = pt.tag(x[0], pt.MPISendTag(to=(rank-1)//size, tag="left")
rght_bnd = pt.tag(x[9], pt.MPISendTag(to=(rank+1)//size, tag="right")
left_halo = pt.mpi_recv(from=(rank-1)//size, tag="right", shape=(), dtype=float)
rght_halo = pt.mpi_recv(from=(rank+1)//size, tag="left", shape=(), dtype=float)
y = advect(x, left_halo, rght_halo)
pt.generate_loopy({"left_bnd": left_bnd, "rght_bnd": rght_bnd, "y": y})  # send nodes are "outputs"

```
The tags identifiers are a hint to the code-generator about the dependencies between the send and receive nodes.

kaushikcfd:
With the same "send via tags" interface an upwind scheme advection could be realized as:
```python
rank = comm.get_rank()
size = comm.get_size()
[...]

PR #59, opened 2021-04-01
Add some guidelines on tag use

inducer:


kaushikcfd (approved):
I generally agree with the tone here. However, I also wanted to shed light on the discussion in
#55. Wrt #55, the text is completely wrong.
1. Stripping off send-tags would be illegal.
2. And in the definition of "semantics" agreed upon, "sends" are a part of the semantic information.

inducer:
> Stripping off send-tags would be illegal.

@matthiasdiener pointed this out as well, in my discussion with him earlier today, and it I agree. (That discussion is, in part, what prompted this PR.) The problem is that I kind of like the way that [the current direction in #55](https://github.com/inducer/pytato/issues/55#issuecomment-808592686) allows the sends to be "along for the ride." What we could do is create an extra array attribute that would allow other nodes to "come along". `tag_along_nodes`? This seems awkward. Got any better ideas?

doc/design.rst
```diff
@@ -151,6 +151,27 @@ Reserved Identifiers
...
+Tags
+----
+
+In order to convey information about the computation from DAG construction
+time to processing/transformation/code generation time, each :class:`pytato.Array`
+node may be tagged (via the :attr:`pytato.Array.tags` attribute) with an arbitrary
+number of informational "tags". A tag is any subclass of :class:`pytools.tag.Tag`.
+Guidelines for tag use:
+
+- Tags *must not* carry semantic information; i.e. a computation must have the same
```
matthiasdiener, line 163: Any particular reason why it needs to be *must not* and not *should not* here? Imho, there are legitimate reasons for conveying semantic information, including the distributedsend stuff as well as tagging operations as unsuitable for GPU, for example (say, file I/O).

kaushikcfd:
> What we could do is create an extra array attribute that would allow other nodes to "come along". tag_along_nodes

Agree with this. From a user's POV I don't see a big difference with prescribing the dependency via a tag or via an attribute.

doc/design.rst
kaushikcfd: >  for example (say, file I/O).

Legitimate point, but I think the chances of allowing I/O into our DAG is slim.

inducer: IMO *must not* would provide the clearest delineation of what can and cannot go into a tag.

My point with this PR was precisely to provide a spot for this discussion.

inducer:
I propose that we move forward with this, and we find another way (in #55) to bring the sends "along for the ride". @matthiasdiener OK with you?

Also, unrelated: CI just seems to hang. I'm a little confused by that.

matthiasdiener (approved):

PR #60, opened 2021-04-02
Account for pure scalar types for user-facing routines

### LINKED ISSUE #57: Should math functions handle scalar values?
Minimal reproduer:
```python
In [1]: import numpy as np

In [2]: np.sin(3.0)
Out[2]: 0.1411200080598672

In [3]: import pytato as pt

In [4]: pt.sin(3.0)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-85cfcccd5e8f> in <module>
----> 1 pt.sin(3.0)

/shared/home/kgk2/pack/virtual_envs/emirge/pytato/pytato/array.py in sin(x)
   1756 
   1757 def sin(x: Array) -> IndexLambda:
-> 1758     return _apply_elem_wise_func(x, "sin")
   1759 
   1760 

AttributeError: 'float' object has no attribute 'dtype'
```

Numpy support scalar values for it's functions. Should we handle it too? I would vote yes, as handling this easy case would lead to better numpy compat.

/cc @inducer

kaushikcfd:
Fixes #57

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -356,7 +356,7 @@ def normalize_shape(
 
 SliceItem = Union[int, slice, None, EllipsisType]
 DtypeOrScalar = Union[_dtype_any, ScalarType]
-ArrayLike = Union["Array", ScalarType]
+ArrayOrScalar = Union["Array", ScalarType]
 
 
 def _truediv_result_type(arg1: DtypeOrScalar, arg2: DtypeOrScalar) -> np.dtype[Any]:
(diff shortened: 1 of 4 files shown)


pytato/array.py
```diff
@@ -355,7 +355,8 @@ def normalize_shape_component(
 # {{{ array inteface
 
 SliceItem = Union[int, slice, None, EllipsisType]
-DtypeOrScalar = Union[_dtype_any, Number]
+DtypeOrScalar = Union[_dtype_any, ScalarType]
+ArrayLike = Union["Array", ScalarType]
```
inducer, line 359:
```suggestion
ArrayOrScalar = Union["Array", ScalarType]
```
?
(Numpy defines the term array-like, and I think that definition disagrees. [ref](https://numpy.org/doc/stable/reference/typing.html#module-numpy.typing))

kaushikcfd: Indeed better not clobbering the variable name. 956b97f (Thanks!)

inducer:
Thanks!

PR #65, opened 2021-04-09
Distributed v2

inducer:
cc @matthiasdiener


examples/mpi.py
```diff
@@ -0,0 +1,131 @@
...
+from pytato.transform import Mapper, WalkMapper
+
+from pytato.array import Array
+
+
+def advect(*args):
+    return sum([a for a in args])
+
+
+class GraphToDictMapper(WalkMapper):
```
inducer, line 13: Are you benefiting from being a `WalkMapper` here? You need to override all `map_` methods here anyhow, and if you miss one, you get incorrect results.

examples/mpi.py
```diff
@@ -0,0 +1,131 @@
...
+    def __init__(self):
+        self.graph_dict = {}
+
+    def visit(self, expr, children):
+        self.graph_dict[expr] = children
+        return True
+
+    def map_placeholder(self, expr, children):
+        children = children | {expr}
+        super().map_placeholder(expr, children)
```
inducer, line 28: Wouldn't this construction end up gathering up *all* the children of a node, not just the immediate ones?

inducer (commented):
Some thoughts on your `GraphToDictMapper` implementation below.

examples/mpi.py
matthiasdiener: Yes, but I thought that was the goal :-/ Will fix.

examples/mpi.py
matthiasdiener: My impression when using the `Mapper` class was that I was just reimplementing a `WalkMapper`, but considering your other comment below a `Mapper` seems more appropiate.

examples/mpi.py
matthiasdiener: Should be fixed now, could you check again @inducer ?

inducer: > Yes, but I thought that was the goal :-/. Will fix.

It can't be: That would be a data structure of size O(n^2)!

examples/mpi.py
```diff
@@ -0,0 +1,145 @@
...
+    def map_placeholder(self, expr, *args):
+        children = set()
+
+        for dim in expr.shape:
+            if isinstance(dim, Array):
+                children = children | {dim}
+                self.rec(dim, *args)
+
+        for c in children:
+            self.graph_dict.setdefault(c, set()).add(expr)
```
inducer, line 32: This collects the graph as `{dependee: depender}`. There's in principle nothing wrong with that, though I wonder whether the code would look (even) more straightforward if it collected the `{depender: dependee}` form. Flipping the edges is easy once we have in a `dict` after all.

inducer:

examples/mpi.py
```diff
@@ -0,0 +1,233 @@
...
+def advect(*args):
+    return sum([a for a in args])
+
+
+class GraphToDictMapper(Mapper):
+    """
+    .. attribute:: graph_dict
+
+        :class:`dict`, maps each node in the graph to the set of directly connected
+        nodes.
```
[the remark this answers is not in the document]
matthiasdiener: Added

matthiasdiener:
Closed in favor of #148.

PR #67, opened 2021-04-17
mapper: small doc improvements

matthiasdiener:


pytato/transform.py
```diff
@@ -84,6 +84,7 @@ def __call__(self, expr: Array, *args: Any, **kwargs: Any) -> Any:
 
 
 class CopyMapper(Mapper):
+    """Performs a deep copy of a :class:`pytato.array.Array`."""
```
inducer, line 87:
```suggestion
    """Performs a deep copy of a :class:`pytato.array.Array`.
    The typical use of this mapper is to override individual ``map_`` methods
    in subclasses to permit term rewriting on an expression graph.
    """
    
```

inducer (approved):
Thanks! I added a small clarifying comment. LGTM!

PR #69, opened 2021-05-11
Pytato array context - merge with main branch

matthiasdiener:


inducer:
cc @kaushikcfd Is the software stack you're working with based on main or on a version that still has the namespace? If it no longer has the namespace, where can we find that version?

kaushikcfd:
> cc @kaushikcfd Is the software stack you're working with based on main or on a version that still has the namespace? If it no longer has the namespace, where can we find that version?

It still has the namespace in it. I am working on reviving #20 with rebase over main first. Almost done with that.

kaushikcfd:
I'm not sure I understand the need for pytato_array_context-copy branch, when inducer's tree already has a pytato_array_context branch. Is it because the earlier branch was trailing?

matthiasdiener:
> I'm not sure I understand the need for pytato_array_context-copy branch. `inducer`'s tree already has a pytato_array_context branch. Is it because the earlier branch was trailing?

I created `pytato_array_context-copy` in order not to mess up the `pytato_array_context` branch. Feel free to modify either one.

kaushikcfd:
> Also, there was no PR with pytato_array_context as far as I could see.

Fair point, but since the commit history is just a merge of 2 branches that have their own PRs #20 and #38, I think having a PR for this branch doesn't have much value.

I have pushed to inducer/pytato_array_context with the rebases and required features for mirgecom. The commit history of pytato_array_context is also more cleaner. It simply denotes what pending PRs does it depend on

Screenshot from 2021-05-12 16-51-38

matthiasdiener:
> > Also, there was no PR with pytato_array_context as far as I could see.
> 
> Fair point, but since the commit history is just a merge of 2 branches that have their own PRs #20 and #38, I think having a PR for this branch doesn't have much value.
> 
> I have pushed to inducer/pytato_array_context with the rebases and required features for mirgecom. The commit history of pytato_array_context is also more cleaner. It simply denotes what pending PRs does it depend on
> 
> Screenshot from 2021-05-12 16-51-38

Thanks, I think this PR can be closed now?

kaushikcfd:
> Thanks, I think this PR can be closed now?

I think so, unless someone foresees something wrong with this workflow.

PR #74, opened 2021-05-22
Drop autodoc_typehints from sphinx config

inducer:
https://github.com/inducer/modepy/pull/38#issuecomment-846472970


inducer:
Dupe of #74

PR #79, opened 2021-06-02
allow complex elwise functions

matthiasdiener:


pytato/array.py
```diff
@@ -1504,7 +1504,7 @@ def _apply_elem_wise_func(x: ArrayOrScalar, func_name: str,
 
     assert isinstance(x, Array)
 
-    if x.dtype.kind != "f":
+    if x.dtype.kind != "f" and x.dtype.kind != "c":
```
kaushikcfd, line 1507: For `abs` the return dtype isn't correct.

kaushikcfd (approved):
Thanks! Could we please also add these to the tests.

matthiasdiener:
> Thanks! Could we please also add these to the tests.

added test

kaushikcfd:
Hah. The tests pass :) even though we have a different result dtype for abs compared to numpy. Once you fix `abs` for complex dtypes, could you also add a check that pytato's result has the same dtype as that of numpy.

matthiasdiener:
> Hah. The tests pass :) even though we have a different result dtype for abs compared to numpy. Once you fix `abs` for complex dtypes, could you also add a check that pytato's result has the same dtype as that of numpy.

Fixed in https://github.com/inducer/pytato/pull/79/commits/62ee559242b141dc5a61e0b5d35ad4b10de12bd4

pytato/array.py
matthiasdiener: Fixed in https://github.com/inducer/pytato/pull/79/commits/62ee559242b141dc5a61e0b5d35ad4b10de12bd4

kaushikcfd:
Thanks a lot! Good to land from my end.

pytato/array.py
```diff
@@ -1517,7 +1517,16 @@ def _apply_elem_wise_func(x: ArrayOrScalar, func_name: str,
 
 
 def abs(x: Array) -> ArrayOrScalar:
-    return _apply_elem_wise_func(x, "abs")
+    if x.dtype == np.complex64:
```
kaushikcfd, line 1520: Could have been simplified as:
```python
if x.dtype.kind == "c":
    result_dtype = np.empty(0, dtype=x.dtype).real.dtype
else:
    result_dtype = x.dtype
```

kaushikcfd: (Optional)

matthiasdiener: Good idea, fixed.

matthiasdiener:
Is this good to go @inducer ?

inducer:
Thanks!

PR #80, opened 2021-06-03
einsum implementation

matthiasdiener:


pytato/array.py
```diff
@@ -717,7 +717,31 @@ def __init__(self,
 
 class Einsum(Array):
```
kaushikcfd, line 718: Is there a reason to have an `Einsum` node and not have `pt.einsum` that emits an IndexLambda with the appropriate sum nodes?

matthiasdiener: Not that I know of; in the initial disucssion with @inducer, we talked about keeping the `Einsum` class.

matthiasdiener:
Currently fails with 
```
$ python examples/einsum.py
<pytato.array.DataWrapper object at 0x10310e310>
  File "examples/einsum.py", line 19, in <module>
    prg = pt.generate_loopy(pt.make_einsum(spec, args), cl_device=queue.device)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/target/loopy/codegen.py", line 725, in generate_loopy
    input_name_recorder(expr)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/transform.py", line 83, in __call__
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/transform.py", line 80, in rec
    return method(expr, *args, **kwargs)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/transform.py", line 313, in map_einsum
    self.rec(child)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/transform.py", line 80, in rec
    return method(expr, *args, **kwargs)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/transform.py", line 271, in map_placeholder
    self.post_visit(expr)
  File "/Users/mdiener/Work/emirge_lazy/pytato/pytato/target/loopy/codegen.py", line 679, in post_visit
    assert expr.name is not None
AssertionError
```

(expr == DataWrapper in this error)

kaushikcfd:
@matthiasdiener: that's probably because the PreprocessMapper/CodegenMapper doesn't have a `map_einsum`.

kaushikcfd:
[This](https://github.com/inducer/pytato/blob/e25dc6cc8762af17b99eb0b4c367451a5fd4ee76/pytato/codegen.py#L102) is the place where names are assigned to the arrays.

pytato/array.py
matthiasdiener: > pt.einsum that emits an IndexLambda with the appropriate sum nodes

How would that look like?

kaushikcfd: For example "ij,j->j", would emit an IndexLambda like `IndexLambda(Reduce(ReductionOp.SUM, [_r0], in0[_r0, _0] *in1[_0]))` with the bindings that would map `in0`, `in1` to the inputs.

kaushikcfd: On thinking it a bit more, maybe this could stay as `Einsum`, as we have a node for MatrixMatrixProduct. If you have started toward working emitting the index-lambda (sorry!) but the work should translate almost completely in the PreprocessMapper.

matthiasdiener: Hehe, no worries. I'm still trying to digest the PreprocessMapper.

inducer: The one reason why I think it makes sense to keep `Einsum` around is because optimizations based on associativity become harder without it. (E.g. think `A@B@x` for matrices A and B. We should obviously pick the O(N^2) way to evaluate this! :)

matthiasdiener:
LGTM, thanks @kaushikcfd! I've manually tested a few of the einsum operations in grudge (current `main`), and they seem to work fine.

pytato/array.py
```diff
@@ -153,6 +154,13 @@
 .. autofunction:: make_size_param
 .. autofunction:: make_data_wrapper
 
+Internal API
+------------
+
+.. autoclass:: EinsumAxisDescriptor
+.. autoclass:: ElementwiseAxis
+.. autoclass:: ReductionAxis
```
matthiasdiener, line 162: Should this be documented if it is internal?

kaushikcfd: These notes could be something a developer might refer to it while working on other parts of the code. Was also needed to silence those sphinx warnings :).

matthiasdiener: Another option might be to put an underscore before the class name, sphinx doesn't complain then I think.

pytato/array.py
```diff
@@ -774,9 +785,250 @@ def __init__(self,
...
+    Einstein summation *subscripts* on *operands*.
+    """
+    if len(operands) == 0:
+        raise ValueError("must specify at least one operand")
+
+    if "->" not in subscripts:
+        # implicit-mode: output spec matched by alphabetical ordering of
+        # indices (ewwwww)
+        raise NotImplementedError("Implicit mode not supported. 'subscripts'"
+                                  " must contains '->', followed by the output's"
```
matthiasdiener, line 1006:
```suggestion
                                  " must contain '->', followed by the output's"
```

pytato/array.py
kaushikcfd: These classes are also needed by routines in `pytato.codegen`. That's why I decided to keep them public but document them as only suitable for a developer.

pytato/array.py
```diff
@@ -774,9 +785,249 @@ def __init__(self,
...
+    """
+    Einstein summation *subscripts* on *operands*.
     """
+    if len(operands) == 0:
+        raise ValueError("must specify at least one operand")
+
+    if "->" not in subscripts:
+        # implicit-mode: output spec matched by alphabetical ordering of
+        # indices (ewwwww)
+        raise NotImplementedError("implicit mode not supported.")
```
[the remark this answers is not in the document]
kaushikcfd: Thanks 2603409.

pytato/array.py
kaushikcfd: Oops and thanks :).

pytato/array.py
matthiasdiener: Ok, sounds good.

pytato/array.py
kaushikcfd: Yup agreed. From a user's perspective tagging the reduction/elementwise axes would be tricky. We went with this direction. Thanks for the clarifications!

pytato/array.py
```diff
@@ -774,9 +785,250 @@ def __init__(self,
...
+
+
+@dataclass(eq=True, frozen=True)
+class ReductionAxis(EinsumAxisDescriptor):
+    dim: int
+
+
 class Einsum(Array):
     """
+    An Einsum expression.
```
inducer, line 808:
```suggestion
    An array expression using the `Einstein summation convention
    <https://en.wikipedia.org/wiki/Einstein_notation>`__. See
    :func:`numpy.einsum` for a similar construct.
    
    .. note::

        Use :func:`einsum` to create this type of expression node in user code.
```

pytato/array.py
```diff
@@ -774,9 +785,250 @@ def __init__(self,
...
+    .. attribute:: access_descriptors
+
+        A :class:`tuple` of *access_descriptor* for each *arg* in
+        :attr:`Einsum.args`. An *access_descriptor* is a tuple of
+        :class:`EinsumAxisDescriptor` denoting how each axis of the
+        argument will be operated in the einstein summation.
+
+    .. attribute:: args
+
+       A :class:`tuple` of array over which the einstein summation is being
```
inducer, line 819:
```suggestion
       A :class:`tuple` of array over which the Einstein summation is being
```

pytato/array.py
```diff
@@ -774,9 +785,250 @@ def __init__(self,
...
 class Einsum(Array):
     """
+    An Einsum expression.
+
+    .. attribute:: access_descriptors
+
+        A :class:`tuple` of *access_descriptor* for each *arg* in
+        :attr:`Einsum.args`. An *access_descriptor* is a tuple of
+        :class:`EinsumAxisDescriptor` denoting how each axis of the
+        argument will be operated in the einstein summation.
```
inducer, line 815: I've got to admit, this representation doesn't make a great deal of sense to me. What is the `dim` in the `EinsumAxisDescriptor` for? How do you determine how many sums to generate? What's wrong with the original, text-based spec, or, said another way, what advantage does this offer over that? How would you turn this back into a string for presentation to the user, after you've thrown away the axis names that the user provided (which may be useful for readability)?

inducer (commented):
Thanks for working on this. I read up to the representation of the node and stopped there, you'll see my comments in that vicinity.

pytato/array.py
kaushikcfd: Thanks ea74db635e13ff395a5a8adc9a7570a80fb53024.

pytato/array.py
kaushikcfd: Thanks ea74db635e13ff395a5a8adc9a7570a80fb53024 :)

pytato/array.py
kaushikcfd: TL;DR version: I favored this format because all our future operations become trivial. As it can be seen, our codegen for einsums is only 50 lines.

kaushikcfd: I guess the other option could be to have a much "flatter", close to user description, that would lead to slight recomputation. But if people that's a better idea, please share your thoughts I'm open to rewiring this a bit.

inducer: I can live with the representation now that it explains itself better.

kaushikcfd: Also added that desciption as comments: 1f5f386de7cf32473ed3d5155160723bd09f420b.

pytato/array.py
```diff
@@ -774,9 +785,267 @@ def __init__(self,
...
+
+        def _scalar_placeholder_like(ary: Array) -> Placeholder:
+            return make_placeholder(shape=(),
+                                    dtype=ary.dtype)
+
+        # preferring this over invoking np.combine_types as pytato might
+        # choose to shift from numpy's binary ops' types
+        return reduce(operator.mul, (_scalar_placeholder_like(arg)
+                                     for arg in self.args[1:]),
+                      _scalar_placeholder_like(self.args[0])).dtype
```
inducer, line 884: This *should* use https://numpy.org/devdocs/reference/generated/numpy.find_common_type.html#numpy-find-common-type.

Some things to consider:

* `numpy` is boss. `pytato` doesn't have a great deal of freedom to stray from `numpy` in semantics. Any deviation is a bug.
* Typing rules in numpy are different for scalars and arrays.

pytato/array.py
```diff
@@ -774,9 +785,267 @@ def __init__(self,
...
+    @property  # type: ignore
+    @memoize_method
+    def shape(self) -> ShapeType:
+        iaxis_to_len = {}
+
+        for access_descr, arg in zip(self.access_descriptors,
+                                     self.args):
+            assert arg.ndim == len(access_descr)
+            for arg_axis_len, axis_op in zip(arg.shape, access_descr):
+                if isinstance(axis_op, ElementwiseAxis):
```
inducer, line 860: `elif isinstance(...): ... else: raise AssertionError()`.

pytato/transform.py
```diff
@@ -350,6 +369,7 @@ def map_named_array(self, expr: NamedArray) -> None:
             return
 
         self.rec(expr._container)
+
         self.post_visit(expr)
 
 # }}}
```
inducer, line 375: Does the visualizer support einsums?

inducer (commented):
Thanks! LGTM aside from the few points below.

pytato/transform.py
kaushikcfd: Yep, it didn't it just emitted a single node with `Einsum` written on it. But now it does: e40c127

pytato/array.py
kaushikcfd: Thanks: 71b827b04b1e75252e4db018ecebbaf5b785e00c

pytato/array.py
kaushikcfd: > numpy is boss.

:)

Fixed here: 
942f2d9

inducer (approved):
Thanks to the both of you for working on this! LGTM. It'll go in after #86.

PR #86, opened 2021-06-10
a few more operators

matthiasdiener:


pytato/array.py
```diff
@@ -537,6 +544,16 @@ def _unary_op(self, op: Any) -> Array:
...
 
+    __and__ = partialmethod(_binary_op, operator.and_)
+    __rand__ = partialmethod(_binary_op, operator.and_, reverse=True)
+    __or__ = partialmethod(_binary_op, operator.or_)
+    __ror__ = partialmethod(_binary_op, operator.or_, reverse=True)
+    __xor__ = partialmethod(_binary_op, operator.xor)
+    __rxor__ = partialmethod(_binary_op, operator.xor, reverse=True)
+
+    def __abs__(self) -> Array:
+        return abs(self)  # type: ignore
```
inducer, line 555: Won't this lead to infinite recursion?

pytato/array.py
```diff
@@ -537,6 +544,16 @@ def _unary_op(self, op: Any) -> Array:
...
+
```
inducer, line 556: Tests?

inducer (commented):
Thanks!

pytato/array.py
kaushikcfd: I think this would pick up https://documen.tician.de/pytato/array.html?highlight=abs#pytato.abs.

kaushikcfd (approved):
Except the tests pointed out by inducer, LGTM from my side as well. Thank you!

pytato/array.py
matthiasdiener: Yeah, it worked fine in my tests.

pytato/array.py
matthiasdiener: Added in https://github.com/inducer/pytato/pull/86/commits/e96cfc101c0b06ecef2508782ccbb0594ea3ed77

pytato/array.py
inducer: Added a comment to un-confuse future self in bb78065. Also, used a different solution to the typing problem.

inducer:
Thanks!

PR #90, opened 2021-06-13
release 2021.2

matthiasdiener:


inducer:
Could you explain the context? I'm a bit surprised given that there's tons of work in flight at the moment.

matthiasdiener:
> Could you explain the context? I'm a bit surprised given that there's tons of work in flight at the moment.

I thought this might make a good "checkpoint" for development - there have been a few major changes compared to the last release, e.g.:
- namespace removal
- many more operations
- kernel callables

Looking at the open PRs, there doesn't seem to be a lot of critical work open at the moment (imho).

inducer:
Hmm, you're right. There's not a lot pending for pytato, I must've been confused between all the different things going on. Fair, let's do it.

inducer:
https://pypi.org/project/pytato/2021.2/

PR #92, opened 2021-06-14
State that pytato arrays have undefined memory layout

inducer:
cc @matthiasdiener


kaushikcfd (approved):
Thanks, I agree with this. Downstream packages can take care of being consistent here.

PR #97, opened 2021-06-22
Enable input arguments to be tagged

kaushikcfd:
There is a genuine need for tagging input arguments (say attaching user
metadata to apply targetted transforms on the generated kernel).
Git-blame/code-comments before this commit doesn't quite shed light on
why this was disabled.


inducer:
I think those modified things in-place at some earlier point, but that's no longer the case. Thanks for proposing this!

PR #99, opened 2021-06-22
Provide pt.Array.reshape interface

kaushikcfd:


kaushikcfd:
I'll go ahead and merge this. Seems fairly non-contentious and I have a requirement for this with some urgency.

PR #102, opened 2021-06-23
Mimic array.reshape behavior to match that of numpy

kaushikcfd:
`array.reshape` is more flexible than  `np.reshape`. `array.reshape` can also
take a sequence of ints to be reshaped into along with a single-arg tuple newshape.


inducer:
Thanks!

PR #103, opened 2021-06-24
Implements CombineMapper

### LINKED ISSUE #96: DependencyMapper should not return every node in the graph
The one use case we have:
https://github.com/inducer/pytato/blob/6e65675341187c39631e57cb95e28e5d6efe75ce/pytato/array.py#L239-L253
only wants a specific node type. Building a set containing *all* nodes is expensive overkill. It'd be much better to have a base `CombineMapper` and then return only the nodes that are actually desired.

The same is true of one upcoming use case:
https://github.com/kaushikcfd/pytato/blob/274c51ec0e1c7f9f68f548b42797011b9befa866/pytato/loopy.py#L251-L268

kaushikcfd:
- DependencyMapper is now a sub-class of CombineMapper.
- Closes #96

### DIFF
--- a/pytato/codegen.py
+++ b/pytato/codegen.py
@@ -24,7 +24,7 @@ THE SOFTWARE.
 
 import dataclasses
 from functools import partialmethod
-from typing import Union, Dict, Tuple, Callable, List, Any
+from typing import Union, Dict, Tuple, Callable, List, Any, FrozenSet
 
 import pymbolic.primitives as prim
 from pymbolic import var
@@ -36,7 +36,7 @@ from pytato.array import (Array, DictOfNamedArrays, IndexLambda,
                           InputArgumentBase, MatrixProduct, Einsum)
 
 from pytato.scalar_expr import ScalarExpression, IntegralScalarExpression
-from pytato.transform import CopyMapper, CachedWalkMapper
+from pytato.transform import CopyMapper, CachedWalkMapper, DependencyMapper
 from pytato.target import Target
 from pytato.loopy import LoopyCall
 from pytools import UniqueNameGenerator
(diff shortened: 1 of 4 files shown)


kaushikcfd:
We should not get rid of DependencyMapper entirely. There are cases when we need to return all the nodes in the DAG. See the user of `pt.transform.get_dependencies` that uses these dependencies to decides the evaluation order for a `DictOfNamedArrays`.

inducer:
> We should not get rid of DependencyMapper entirely. There are cases when we need to return all the nodes in the DAG. See the (only) user of `pt.transform.get_dependencies` that uses these dependencies to decides the evaluation order for a `DictOfNamedArrays`.

Yikes! I'm surprised I missed that during review. This creates an intermediate data structure of size O(N^2) for a measly topsort (which should be linear-time). (IMO) we should fix that, or else it'll kill us before too long.

inducer:
(Two ways out IMO: Let this not close #96, or fix it now. Your pick.) Also, thanks for picking this up!

kaushikcfd:
I'll fix it now. Since we can solve it by sub-classing the CombineMapper.

kaushikcfd:
@inducer: How does 944f73 look?

pytato/codegen.py
```diff
@@ -482,9 +482,29 @@ class PreprocessResult:
     bound_arguments: Dict[str, DataInterface]
 
 
+# {{{ SubsetDependencyMapper
+
+class SubsetDependencyMapper(DependencyMapper):
```
inducer, line 487: Move to `transform`, add to docs.

pytato/transform.py
```diff
@@ -169,83 +170,147 @@ def map_dict_of_named_arrays(self,
                                   for key, val in expr.items()})
 
 
-class DependencyMapper(Mapper):
-    """
-    Maps a :class:`pytato.array.Array` to a :class:`frozenset` of
-    :class:`pytato.array.Array`'s it depends on.
-    """
-
+class CombineMapper(Mapper, Generic[CombineT]):
```
inducer, line 173: Add to docs.

pytato/transform.py
```diff
@@ -169,83 +170,147 @@ def map_dict_of_named_arrays(self,
...
+        return self.combine(*(self.rec(ary)
+                              for ary in expr.bindings.values()
+                              if isinstance(ary, Array)))
+
+
+class DependencyMapper(CombineMapper[R]):
+    """
+    Maps a :class:`pytato.array.Array` to a :class:`frozenset` of
+    :class:`pytato.array.Array`'s it depends on.
+    """
```
inducer, line 250:
```suggestion
    .. warning::
    
        This returns every node in the graph! Consider a custom :class:`CombineMapper`
        or a :class:`SubsetDependencyMapper` instead.
    """
```

inducer (approved):
> @inducer: How does 944f738 look?

I don't know. the The dangerous version of `DependencyMapper` is still around and ready to shoot anyone in the foot who will let it. I guess I can live with this, though I'm not elated about it. :)

Thanks again for taking care of this. A few minor things, after which this is ready to land.

pytato/codegen.py
kaushikcfd: Thanks, 4a7098d.

pytato/transform.py
kaushikcfd: Thanks,  4a7098d.

kaushikcfd: Done in 4a7098d.

inducer:
Thanks!

PR #104, opened 2021-06-24
Implement Array.__len__

kaushikcfd:


matthiasdiener (approved):

pytato/array.py
```diff
@@ -381,6 +381,8 @@ class Array(Taggable):
     # hashable. Dicts of hashable keys and values are also permitted.
     _fields: ClassVar[Tuple[str, ...]] = ("shape", "dtype", "tags")
 
+    __array_priority__ = 1  # disallow numpy arithmetic to take preference
```
kaushikcfd, line 384: Had to do this, otherwise numpy was considering `Array` to be array_like (as after this PR an array has both `__len__` and `__getitem__`) and try to perform arithmetic by taking the array's slices.

inducer:
```suggestion
    __array_priority__ = 1  # disallow numpy arithmetic to take precedence
```

inducer (commented):
Thanks! (I'll apply.)

PR #105, opened 2021-06-24
Fixes a bug in einsum' shape computation

kaushikcfd:
check that the argument axis length corresponding to the same output axis
matches with the previously recorded one.

The provided test fails on `main`.


inducer:
Thanks!

PR #106, opened 2021-06-24
Cache the results of a WalkMapper

kaushikcfd:
Array expressions are a DAG, if an expression is traversed do not re-traverse it.

### DIFF
--- a/pytato/transform.py
+++ b/pytato/transform.py
@@ -24,7 +24,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import Any, Callable, Dict, FrozenSet, Union, TypeVar
+from typing import Any, Callable, Dict, FrozenSet, Union, TypeVar, Set
 
 from pytato.array import (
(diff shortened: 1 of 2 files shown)


pytato/transform.py
```diff
@@ -259,6 +259,20 @@ class WalkMapper(Mapper):
     .. automethod:: visit
     .. automethod:: post_visit
     """
+    def __init__(self) -> None:
```
inducer, line 262: I think not every `WalkMapper` will want this. Make this a subclass?

pytato/transform.py
```diff
@@ -259,6 +259,20 @@ class WalkMapper(Mapper):
...
+        self._cache: Dict[ArrayOrNames, Any] = {}
```
inducer, line 263: I don't think `WalkMapper`s return anything... this can probably be a set?

inducer:

pytato/transform.py
kaushikcfd: Yes, thanks. There is a `CachedWalkMapper` now.

pytato/transform.py
kaushikcfd: Done. fc06fc2

inducer:
Thanks!

PR #116, opened 2021-07-01
Raise IndexError for OOB access

kaushikcfd:
Mimics numpy as:

```python
>>> a = np.random.rand(10)
>>> a[10]
  File "<stdin>", line 1, in <module>
IndexError: index 10 is out of bounds for axis 0 with size 10
>>> a[0, 0]
  File "<stdin>", line 1, in <module>
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```


inducer:
Thx!

PR #120, opened 2021-07-05
LoopyTarget: Reduction index lambda handle depends_on correctly

kaushikcfd:
The attached regression fails on `main`.

### DIFF
--- a/pytato/target/loopy/codegen.py
+++ b/pytato/target/loopy/codegen.py
@@ -33,9 +33,8 @@ import pytato.scalar_expr as scalar_expr
 import pymbolic.primitives as prim
 from pymbolic import var
 
-from typing import (
-        Union, Optional, Mapping, Dict, Tuple, FrozenSet, Set, Callable,
-        Any, List)
+from typing import (Union, Optional, Mapping, Dict, Tuple, FrozenSet, Set,
+                    Any, List)
 
 
 from pytato.array import (Array, DictOfNamedArrays, ShapeType, IndexLambda,
@@ -55,7 +54,7 @@ if getattr(sys, "PYTATO_BUILDING_SPHINX_DOCS", False):
     import pyopencl
 
 __doc__ = """
-.. autoclass:: LoopyExpressionContext
+.. autoclass:: SharedLoopyExpressionContext
 .. autoclass:: ImplementedResult
 .. autoclass:: StoredResult
 .. autoclass:: InlinedResult
@@ -68,7 +67,6 @@ __doc__ = """
 .. autofunction:: domain_for_shape
 .. autofunction:: get_loopy_temporary
 .. autofunction:: add_store
-.. autofunction:: rename_reductions
 .. autofunction:: normalize_outputs
 .. autofunction:: get_initial_codegen_state
 """
@@ -82,13 +80,13 @@ def loopy_substitute(expression: Any, variable_assigments: Mapping[str, Any]) ->
 
 # SymbolicIndex and ShapeType are semantically distinct but identical at the
 # type level.
-ReductionBounds = Dict[str, Tuple[ScalarExpression, ScalarExpression]]
+ReductionBounds = Mapping[str, Tuple[ScalarExpression, ScalarExpression]]
 
 
-# {{{ LoopyExpressionContext
+# {{{ LoopyExpressionContexts
 
 @dataclasses.dataclass(init=True, repr=False, eq=False)
-class LoopyExpressionContext(object):
+class SharedLoopyExpressionContext(object):
     """Mutable state used while generating :mod:`loopy` expressions.
     Wraps :class:`CodeGenState` with more expression-specific information.
 
@@ -129,8 +127,6 @@ class LoopyExpressionContext(object):
             dataclasses.field(default_factory=frozenset)
     local_namespace: Mapping[str, Array] = \
             dataclasses.field(default_factory=dict)
-    reduction_bounds: ReductionBounds = \
-            dataclasses.field(default_factory=dict)
 
     def lookup(self, name: str) -> Array:
         return self.local_namespace[name]
@@ -141,6 +137,22 @@ class LoopyExpressionContext(object):
 
     def update_depends_on(self, other: FrozenSet[str]) -> None:
         self._depends_on = self._depends_on | other
+
+
+@dataclasses.dataclass(frozen=True)
+class UpstreamLoopyExpressionContext:
+    """
+    Records context being to be conveyed from a parent expression to its
+    sub-expressions.
+    """
+    reduction_bounds: ReductionBounds = \
+            dataclasses.field(default_factory=dict)
+
+    def copy(self, *,
+             reduction_bounds: Optional[ReductionBounds] = None
+             ) -> UpstreamLoopyExpressionContext:
+        reduction_bounds = reduction_bounds or self.reduction_bounds
+        return UpstreamLoopyExpressionContext(reduction_bounds)
 
 # }}}
 
@@ -156,7 +168,7 @@ class ImplementedResult(ABC):
 
     @abstractmethod
     def to_loopy_expression(self, indices: SymbolicIndex,
-            expr_context: LoopyExpressionContext) -> ScalarExpression:
+            expr_context: SharedLoopyExpressionContext) -> ScalarExpression:
         """Return a :mod:`loopy` expression for this result.
 
         :param indices: symbolic expressions for the indices of the array
@@ -165,11 +177,6 @@ class ImplementedResult(ABC):
 
             - *depends_on* is populated with any dependencies needed for the
               generated expression.
-
-            - *reduction_bounds* is populated with reduction bounds for the
-              reduction inames in the returned expression. If
-              *reduction_bounds* is nonempty, then the returned inames are
-              ensured to be disjoint from those present.
         """
 
 # }}}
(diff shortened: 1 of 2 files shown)


pytato/target/loopy/codegen.py
```diff
@@ -573,13 +582,10 @@ def map_reduce(self, expr: scalar_expr.Reduce,
...
 
-        inner_expr = self.rec(expr.inner_expr,
-                              LoopyExpressionContext(
-                                  state=state,
-                                  _depends_on=expr_context.depends_on,
-                                  local_namespace=expr_context.local_namespace,
-                                  num_indices=expr_context.num_indices,
-                                  reduction_bounds=expr.bounds))  # type: ignore
+        # type-ignore-reason: passed 'Mapping', expected 'Dict'
+        with expr_context.swapped_reduction_bounds(expr.bounds):  # type: ignore
```
inducer, line 586: * This isn't correct. The body of an inner reduction is allowed to use reduction inames defined in an outer reduction, which this would prohibit.
* What I'd like to happen is that `LoopyExpressionContext` be split into two parts:
    * There are some parts (like `depends_on`) that gather information for instruction code gen (once recursion completes). That part is fine to imperatively update.
    * The other parts (like `reduction_bounds`) that communicate with lower levels of the recursion. Those parts should be done by making a modified copy of the context.
    
    Those two uses are mutually incompatible, therefore `LoopyExpressionContext` needs to be split.

pytato/target/loopy/codegen.py
```diff
@@ -140,6 +141,14 @@ def depends_on(self) -> FrozenSet[str]:
...
         self._depends_on = self._depends_on | other
 
+    @contextmanager
+    def swapped_reduction_bounds(self, bounds: ReductionBounds
+                                 ) -> Iterator[None]:
+        old_bounds = self.reduction_bounds
+        self.reduction_bounds = bounds
+        yield
+        self.reduction_bounds = old_bounds
+
```
inducer, line 151: Not loving this, it should go. (see more detailed comment below)

inducer (commented):
Thanks! Some thoughts below.

pytato/target/loopy/codegen.py
kaushikcfd: Removed.

pytato/target/loopy/codegen.py
kaushikcfd: Yep, agreed. Done. `LoopyExpressionContext` is now split into two parts:
- SharedLoopyExpressionContext
- UpstreamExpressionContext

pytato/target/loopy/codegen.py
```diff
@@ -573,13 +582,10 @@ def map_reduce(self, expr: scalar_expr.Reduce,
...
                 for old_name in expr.bounds}
 
-        inner_expr = self.rec(expr.inner_expr,
-                              LoopyExpressionContext(
-                                  state=state,
-                                  _depends_on=expr_context.depends_on,
-                                  local_namespace=expr_context.local_namespace,
-                                  num_indices=expr_context.num_indices,
-                                  reduction_bounds=expr.bounds))  # type: ignore
+        # type-ignore-reason: passed 'Mapping', expected 'Dict'
```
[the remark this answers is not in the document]
kaushikcfd: Done.

pytato/target/loopy/codegen.py
```diff
@@ -540,80 +534,89 @@ def __init__(self, codegen_mapper: CodeGenMapper):
         self.codegen_mapper = codegen_mapper
 
     def __call__(self, expr: ScalarExpression,
-            expr_context: LoopyExpressionContext) -> ScalarExpression:
-        return self.rec(expr, expr_context)
+                 shared_ctx: SharedLoopyExpressionContext,
+                 upstream_ctx: Optional[UpstreamLoopyExpressionContext] = None,
```
inducer, line 538: It doesn't look like that argument is used/supplied anywhere. (I was wondering what for anyway?) I tried without, and the tests still seem happy.

pytato/target/loopy/codegen.py
```diff
@@ -366,7 +360,7 @@ def map_index_lambda(self, expr: IndexLambda,
 
         # TODO: Respect tags.
 
-        loopy_expr_context = LoopyExpressionContext(state,
+        loopy_expr_context = SharedLoopyExpressionContext(state,
```
inducer, line 363: For being "shared"/"persistent", it's weird that this makes a new context. I suspect that that has the potential for some more missed dependencies. Maybe `num_indices` and the local namespace want to be in the "local" context as well?

pytato/target/loopy/codegen.py
```diff
@@ -479,7 +473,7 @@ def _get_sub_array_ref(array: Array, name: str) -> "lp.symbolic.SubArrayRef":
             else:
                 assert isinstance(arg, lp.ValueArg) and arg.is_input
                 pt_arg = expr.bindings[arg.name]
-                loopy_expr_context = LoopyExpressionContext(state,
+                loopy_expr_context = SharedLoopyExpressionContext(state,
```
inducer, line 476: As above. Is it correct that this makes a new one of these?

pytato/target/loopy/codegen.py
```diff
@@ -628,7 +631,7 @@ def shape_to_scalar_expression(shape: ShapeType,
                                cgen_mapper: CodeGenMapper,
                                state: CodeGenState
                                ) -> Tuple[ScalarExpression, ...]:
-    shape_context = LoopyExpressionContext(state, num_indices=0)
+    shape_context = SharedLoopyExpressionContext(state, num_indices=0)
```
inducer, line 634: If this has a non-empty `depends_on`, will those dependencies get respected?

inducer (commented):
Thanks! Some more fishing around in here for things I might not like, and asking questions about it... below. :)

inducer:

pytato/target/loopy/codegen.py
kaushikcfd: It gives some "extra" context for the variables downstream. Before emitting a loopy expression we verify whether every variable that we emit is "known" i.e. either it is an iname/placeholder name.

I tried removing the `upstream_ctx` and ended up in a bunch of errors, which essentially read that no context about `_pt_sum_r0` was known.

pytato/target/loopy/codegen.py
kaushikcfd: I don't think these should have a non-empty `depends_on` for now, added an `assert` to make sure that's the case.

pytato/target/loopy/codegen.py
```diff
@@ -82,13 +80,13 @@ def loopy_substitute(expression: Any, variable_assigments: Mapping[str, Any]) ->
...
-ReductionBounds = Dict[str, Tuple[ScalarExpression, ScalarExpression]]
+ReductionBounds = Mapping[str, Tuple[ScalarExpression, ScalarExpression]]
 
 
-# {{{ LoopyExpressionContext
+# {{{ LoopyExpressionContexts
 
 @dataclasses.dataclass(init=True, repr=False, eq=False)
-class LoopyExpressionContext(object):
+class SharedLoopyExpressionContext(object):
```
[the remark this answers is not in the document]
kaushikcfd: Agreed, went with PersistentExpressionContext, thanks!

pytato/target/loopy/codegen.py
```diff
@@ -142,6 +138,22 @@ def depends_on(self) -> FrozenSet[str]:
     def update_depends_on(self, other: FrozenSet[str]) -> None:
         self._depends_on = self._depends_on | other
 
+
+@dataclasses.dataclass(frozen=True)
+class UpstreamLoopyExpressionContext:
+    """
```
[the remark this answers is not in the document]
kaushikcfd: Went with LocalExpressionContext here.

pytato/target/loopy/codegen.py
```diff
@@ -142,6 +138,22 @@ def depends_on(self) -> FrozenSet[str]:
     def update_depends_on(self, other: FrozenSet[str]) -> None:
         self._depends_on = self._depends_on | other
 
+
+@dataclasses.dataclass(frozen=True)
+class UpstreamLoopyExpressionContext:
```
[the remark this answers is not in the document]
kaushikcfd: Thanks, added.

pytato/target/loopy/codegen.py
kaushikcfd: Yep, agreed that was ill-placed. Moved these attributes into `LocalExpressionContext`.

pytato/target/loopy/codegen.py
kaushikcfd: I have come to think `PersistentExpressionContext` as being local to a `ImplementedResult` and so that justifies this instantiation.

kaushikcfd: (Amended the docs to note this)

kaushikcfd:
@inducer: Thanks, I think the split b/w LocalExpressionContext and PersistentExpressionContext is cleaner now. Could you please give it a look?

pytato/target/loopy/codegen.py
```diff
@@ -542,80 +540,87 @@ def __init__(self, codegen_mapper: CodeGenMapper):
...
-                                          expr_context)
+                upstream_ctx.lookup(expr.aggregate.name), shared_ctx.state)
+        return result.to_loopy_expression(self.rec(expr.index, shared_ctx,
+                                                   upstream_ctx),
+                                          shared_ctx)
 
     def map_variable(self, expr: prim.Variable,
-            expr_context: LoopyExpressionContext) -> ScalarExpression:
+                     shared_ctx: PersistentExpressionContext,
+                     upstream_ctx: LocalExpressionContext,
```
inducer, line 561: These variable names are stale. (Here and in a few other places.) Could you fix them?

inducer (commented):
Thanks, functionality-wise this looks good now. Could you clean up the variable names? Or do you want to do that later?

pytato/target/loopy/codegen.py
kaushikcfd: Thanks, d10d949.

kaushikcfd:
@inducer: Thanks a lot for all the feedback. Looks like we have converged on the code here. I'll set this to automerge.

PR #121, opened 2021-07-05
Cleanups

inducer:
Best read commit-by-commit, in case you're watching. Uncontroversial IMO, so I'm merging right away.


kaushikcfd:
Thanks! Agree that `InputGatherer` is a cleaner approach.

PR #123, opened 2021-07-14
Make pt.(Placeholder|SizeParam) equality comparable

kaushikcfd:


pytato/array.py
```diff
@@ -223,12 +223,16 @@ class EllipsisType(Enum):
     Sequence[ShapeComponent]]
 
 
-def _check_identifier(s: Optional[str]) -> bool:
+def _check_identifier(s: Optional[str], optional: bool) -> bool:
+    err = ValueError(f"'{s}' is not a valid identifier")
```
inducer, line 227: This pays the construction of the string and the `ValueError` unconditionally. Not sure it's worth it.

pytato/array.py
```diff
@@ -1489,6 +1487,12 @@ def __init__(self,
         self.data = data
         self._shape = shape
 
+    def __hash__(self) -> int:
```
inducer, line 1490: Add a comment to the docstring saying that, since we may not be able to compare the wrapped arrays, each instance of  `DataWrapper` only compares equal to itself.

pytato/transform.py
```diff
@@ -536,18 +538,18 @@ class CachedWalkMapper(WalkMapper):
...
     def rec(self, expr: ArrayOrNames) -> None:  # type: ignore
-        if expr in self._visited_nodes:
+        if id(expr) in self._visited_ids:
             return
 
         # type-ignore reason: super().rec expects either 'Array' or
         # 'AbstractResultWithNamedArrays', passed 'ArrayOrNames'
         super().rec(expr)  # type: ignore
-        self._visited_nodes.add(expr)
+        self._visited_ids.add(id(expr))
```
inducer, line 552: Wouldn't it be essentially equivalent to *not* do this `id` business? If this is for efficiency, then, quite possibly, `Array.__eq__` should have a shortcut if `self is other`, and `Array.__hash__` should cache the hash value. That would help not just this but all sorts of other code.

inducer (commented):
Thanks! Just a few quick and hopefully easy things here, then this should be ready to roll.

inducer:

pytato/transform.py
kaushikcfd: > Wouldn't it be essentially equivalent to not do this id business?

Nope, we need the id as a cache key to catch illegals DAGs like: https://github.com/inducer/pytato/blob/ffc467acc3223804ba988e9624739929362da866/test/test_pytato.py#L187-L199.

pytato/array.py
kaushikcfd: Done, thanks!

pytato/array.py
kaushikcfd: Thanks,  5721867.

pytato/transform.py
inducer: I see what you're saying. Add a comment justifying the use of `id`?

kaushikcfd: Done, see 96e8071.

pytato/array.py
```diff
@@ -1495,6 +1492,11 @@ class DataWrapper(InputArgumentBase):
 
         Starting with the construction of the :class:`DataWrapper`,
         this array may not be updated in-place.
+
+    .. note::
+
+        Since we cannot compare instances of :class:`DataInterface` being
+        wrapped, a :class:`DataWrapper` can only be equal to itself.
```
inducer, line 1499:
```suggestion
        wrapped, a :class:`DataWrapper` instances compare equal to themselves
        (i.e. the very same instance).
```

pytato/transform.py
```diff
@@ -536,18 +538,22 @@ class CachedWalkMapper(WalkMapper):
...
-        self._visited_nodes: Set[ArrayOrNames] = set()
+        self._visited_ids: Set[int] = set()
 
     # type-ignore reason: CachedWalkMapper.rec's type does not match
     # WalkMapper.rec's type
     def rec(self, expr: ArrayOrNames) -> None:  # type: ignore
-        if expr in self._visited_nodes:
+        # Why choose id(x) as the cache key?
+        # - Some downstream users (NamesValidityChecker) of this mapper rely on
+        #   object-identity while implementing their logic during the walk.
```
inducer, line 548:
```suggestion
        # - Some downstream users (NamesValidityChecker) of this mapper rely on
        #   structurally equal objects being walked separately (e.g. to detect separate
        #   instances of Placeholder with the same name).
```

inducer (commented):
Thanks, LGTM! (I'll apply the two suggestions.)

PR #125, opened 2021-07-16
Define pt.Array.copy

kaushikcfd:
Test shipped with this patch fails on `main` with:
```
self = <pytato.array.Placeholder object at 0x7f5108c53640>, kwargs = {'tags': frozenset({test_tagging_array.<locals>.BestArrayTag()})}                
                                                                                                                                                      
    def copy(self: T_co, **kwargs: Any) -> T_co:                                                                                                      
        """                                                                                                                                           
        Returns of copy of *self* with the specified tags. This method                                                                                
        should be overridden by subclasses.                                                                                                           
        """                                                                                                                                           
>       raise NotImplementedError("The copy function is not implemented.")                                                                            
E       NotImplementedError: The copy function is not implemented.                                                                                    
                                                                                                                                                      
../../lib/python3.8/site-packages/pytools/tag.py:221: NotImplementedError 
```

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -179,7 +179,7 @@ import operator
 from dataclasses import dataclass
 from typing import (
         Optional, Callable, ClassVar, Dict, Any, Mapping, Tuple, Union,
-        Protocol, Sequence, cast, TYPE_CHECKING, List, Iterator)
+        Protocol, Sequence, cast, TYPE_CHECKING, List, Iterator, TypeVar)
 
 import numpy as np
 import pymbolic.primitives as prim
@@ -212,6 +212,8 @@ if TYPE_CHECKING:
     _dtype_any = np.dtype[Any]
 else:
     _dtype_any = np.dtype
+
+ArrayT = TypeVar("ArrayT", bound="Array")
 
 
 # {{{ shape
@@ -380,9 +382,15 @@ class Array(Taggable):
     _mapper_method: ClassVar[str]
     # A tuple of field names. Fields must be equality comparable and
     # hashable. Dicts of hashable keys and values are also permitted.
-    _fields: ClassVar[Tuple[str, ...]] = ("shape", "dtype", "tags")
+    _fields: ClassVar[Tuple[str, ...]] = ("tags",)
 
     __array_priority__ = 1  # disallow numpy arithmetic to take precedence
+
+    def copy(self: ArrayT, **kwargs: Any) -> ArrayT:
+        for field in self._fields:
+            if field not in kwargs:
+                kwargs[field] = getattr(self, field)
+        return type(self)(**kwargs)
 
     @property
     def shape(self) -> ShapeType:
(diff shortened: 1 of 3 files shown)


pytato/array.py
```diff
@@ -705,6 +714,7 @@ class NamedArray(Array):
     .. automethod:: __init__
     """
     _fields = Array._fields + ("_container", "name")
+    _input_fields: ClassVar[Tuple[str, ...]] = ("container", "name", "tags", )
```
inducer, line 717: Why not use the superclass's `_input_fields` here?

pytato/array.py
```diff
@@ -381,9 +383,16 @@ class Array(Taggable):
...
     _fields: ClassVar[Tuple[str, ...]] = ("shape", "dtype", "tags")
+    # A tuple of constructor's argument names
+    _input_fields: ClassVar[Tuple[str, ...]] = ("tags", )
 
     __array_priority__ = 1  # disallow numpy arithmetic to take precedence
 
+    def copy(self: ArrayT, **kwargs: Any) -> ArrayT:
+        copy_kwargs = {attr_name: kwargs.get(attr_name, getattr(self, attr_name))
+                       for attr_name in self._input_fields}
+        return type(self)(**copy_kwargs)
```
inducer, line 394: Should error on unused kwargs.

pytato/array.py
```diff
@@ -381,9 +383,16 @@ class Array(Taggable):
     # A tuple of field names. Fields must be equality comparable and
     # hashable. Dicts of hashable keys and values are also permitted.
     _fields: ClassVar[Tuple[str, ...]] = ("shape", "dtype", "tags")
+    # A tuple of constructor's argument names
+    _input_fields: ClassVar[Tuple[str, ...]] = ("tags", )
```
inducer, line 387: I'm not sure I understand why we have both and what the distinction is between the two.

inducer (commented):
Thanks!

inducer:

pytato/array.py
kaushikcfd: There were some issues while re-using `_fields`. For some classes the attribute name and the constructor argument name weren't agreeing, like for ex. in [NamedArray](https://github.com/inducer/pytato/blob/8f6a07fe041bc5a4bf35e770005f1b25f9b2efb0/pytato/array.py#L711-L715).

However after your comments, I agree this isn't a good enough reason to re-introduce another `_input_fields`. Handled it by special-casing NamedArray.copy: df7b57e.

pytato/array.py
kaushikcfd: Done, thanks!

pytato/array.py
kaushikcfd: See https://github.com/inducer/pytato/pull/125#discussion_r671326698, we don't use `_input_fields` anymore.

kaushikcfd:
@inducer: I think this is better now. For better reviewing experience please view the patch on a commit-by-commit basis.

inducer:
Thanks! LGTM, in it goes. Btw: Found #126 while reviewing this.

PR #127, opened 2021-07-16
Show a DAG manipulating example: materialization

kaushikcfd:
- Implemented `ImplStored` implementation strategy.
- Showed how tagging an array with `ImplStored` reduces the generated kernel's FlOps.
- I think the test which enables us to materialize the arrays is a helpful point to look at the motivation of the PR.

Draft, because:
- [ ] Depends on https://github.com/inducer/pytools/pull/97.


pytato/analysis.py
```diff
@@ -0,0 +1,90 @@
...
+        self.nusers[expr.x1] += 1
+        self.nusers[expr.x2] += 1
+        return super().map_matrix_product(expr)
+
+
+def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
+    """
+    For the DAG *outputs*, returns the mapping from each node to its number of
+    successors.
+    """
```
inducer, line 85: If this is a `DictOfNamedArrays`, what is actually being counted?

pytato/analysis.py
```diff
@@ -0,0 +1,90 @@
...
+
+        return super().map_einsum(expr)
+
+    def map_matrix_product(self, expr: MatrixProduct) -> None:
+        self.nusers[expr.x1] += 1
+        self.nusers[expr.x2] += 1
+        return super().map_matrix_product(expr)
+
+
+def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
```
inducer, line 81: Add to docs.

pytato/analysis.py
```diff
@@ -0,0 +1,90 @@
...
+from typing import Mapping, Dict, Union
+from .array import (Array, IndexLambda, Stack, Concatenate,
+                    Einsum, MatrixProduct, DictOfNamedArrays)
+from .transform import CachedWalkMapper
+from .loopy import LoopyCall
+
+
+class NUserCollector(CachedWalkMapper):
+    """
+    A WalkMapper that records the number of times an array expression is a
```
inducer, line 36: `WalkMapper`: Make proper ReST referenc.e

pytato/analysis.py
```diff
@@ -0,0 +1,90 @@
...
+    direct dependency of other nodes.
+    """
+    def __init__(self) -> None:
+        from collections import defaultdict
+        super().__init__()
+        self.nusers: Dict[Array, int] = defaultdict(lambda: 0)
+
+    def map_index_lambda(self, expr: IndexLambda) -> None:
```
inducer, line 44: Couldn't this be done generically in `visit`?

pytato/array.py
```diff
@@ -632,6 +623,15 @@ def reshape(self, *shape: Union[int, Sequence[int]], order: str = "C") -> Array:
...
 
+    @memoize_method
+    def tags_of_type(self, tag_t: type) -> FrozenSet[Tag]:
+        """
+        Returns *self*'s tags of type *tag_t*.
+        """
+        return frozenset(tag
+                         for tag in self.tags
+                         if isinstance(tag, tag_t))
+
```
inducer, line 634: Move to `pytools.tag`?

pytato/tags.py
```diff
@@ -0,0 +1,63 @@
...
+
+# }}}
+
+
+# {{{ pre-defined tag: CountNamed
+
+@tag_dataclass
+class CountNamed(UniqueTag):
+    """
+    .. attribute:: name
```
inducer, line 58: What is this tag supposed to mean?

pytato/tags.py
```diff
@@ -0,0 +1,63 @@
...
+.. autoclass:: ImplDefault
+"""
+
+
+from pytools.tag import Tag, UniqueTag, tag_dataclass
+
+
+# {{{ pre-defined tag: ImplementAs
+
+@tag_dataclass
```
inducer, line 22: Probably time to document these as well.

pytato/target/loopy/codegen.py
```diff
@@ -364,19 +365,34 @@ def map_index_lambda(self, expr: IndexLambda,
...
                 local_namespace=expr.bindings,
                 num_indices=expr.ndim)
         loopy_expr = self.exprgen_mapper(expr.expr, loopy_expr_context)
 
-        result = InlinedResult.from_loopy_expression(loopy_expr,
-                loopy_expr_context)
-        state.results[expr] = result
+        result: ImplementedResult = InlinedResult.from_loopy_expression(
+                                                    loopy_expr,
+                                                    loopy_expr_context)
```
inducer, line 375: Are you sure that `from_loopy_expression` will have no side effects that would need to be undone if this goes unused?

pytato/__init__.py
```diff
@@ -94,4 +97,7 @@
 
         "dot", "vdot",
 
+        # sub-modules
+        "analysis", "tags", "transform",
```
inducer, line 101: These aren't public, right? `tags` maybe yes, but the others?

pytato/transform.py
```diff
@@ -559,6 +567,30 @@ def rec(self, expr: ArrayOrNames) -> None:  # type: ignore
 # }}}
 
 
+# {{{ MapAndCopyMapper
+
+class MapAndCopyMapper(CopyMapper):
```
inducer, line 572: Document? Also document the caching behavior... maybe change the name to reflect it?

inducer (commented):
Thanks! A few comments below.

inducer:

pytato/analysis.py
kaushikcfd: It would return the mapping for each node reachable from the output.

pytato/analysis.py
kaushikcfd: Thanks, done!

pytato/analysis.py
kaushikcfd: Not quite. For `Union[pt.Stack, pt.Concatenate]` nodes the predecessor is determined by the `.arrays` attribute. For `Union[pt.IndexLambda, pt.loopy.LoopyCallResult]` nodes, the predecessor is determined by `.bindings` attribute. So having a single visit for the nodes doesn't quite work unless we insert `if-else`s in the `.visit` method.

pytato/array.py
kaushikcfd: Yep makes sense. https://github.com/inducer/pytools/pull/98

pytato/tags.py
kaushikcfd: TBH, I'm not sure. I just moved  it from [pytato/array.py](https://github.com/inducer/pytato/blob/23eb73c9616a44cc40fe6db294c8ba988fd5c7a4/pytato/array.py#L693-L703). I think we should get rid of it?

pytato/__init__.py
kaushikcfd: Chose to do it because pt.analysis, pt.transform also supply public-facing routines like `pt.analysis.get_nusers`, `pt.transform.get_dependencies`, and exporting all of them under top-level `pytato` namespace seemed to pollute the namespace.

pytato/transform.py
kaushikcfd: Thanks, done.

pytato/analysis.py
kaushikcfd: Thanks, done!

pytato/target/loopy/codegen.py
kaushikcfd: > no side effects 

I can't think of one Also for a `ImplStored`-tagged array, the result is not unused but passed to `add_store` which is a typical pattern in `pytato.target.loopy.codegen`, so I think we should be fine.

pytato/tags.py
kaushikcfd: Done!

pytato/tags.py
```diff
@@ -0,0 +1,73 @@
...
+    pass
+
+
+@tag_dataclass
+class ImplDefault(ImplementationStrategy):
+    pass
+
+
+@tag_dataclass
+class ImplementAs(UniqueTag):
```
inducer, line 47: This actually seems unnecessary. The way the `UniqueTag` machinery worked out, we could simply tag things with the `Impl*` tags and that should suffice. I don't see this wrapper serving a really useful purpose.

pytato/target/loopy/codegen.py
inducer: If this produces a new `depends_on`, will that get respected? How does that information travel?

inducer (commented):
Thanks! Some more thoughts below.

inducer:

pytato/tags.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Records metadata to be attached to :class:`pytato.Array` to convey
+    information to a :class:`pytato.target.Target` on how its supposed to be
+    lowered to.
```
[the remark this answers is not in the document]
kaushikcfd: Done, thanks!

pytato/tags.py
kaushikcfd: Removed, thanks!

pytato/tags.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+
+@tag_dataclass
+class ImplInlined(ImplementationStrategy):
+    pass
+
+
+@tag_dataclass
+class ImplDefault(ImplementationStrategy):
```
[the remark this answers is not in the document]
kaushikcfd: Documented `ImplInlined`.
Don't really know what `ImplDefault` stands for (missing the context since wasn't in the room while it was drafted).

pytato/target/loopy/codegen.py
kaushikcfd: > If this produces a new depends_on

I don't think new depdends_on will come up, as at this point all its predecessors have been traversed and accounted for.

The logic here is similar to using `StoredResult` instead of `InlinedResult` at other places, for example in `LoopyCallResult`.

pytato/tags.py
kaushikcfd: Aah thanks! I thought the name `CountNamed` was a bit restrictive and made it `Named` as other `ImplStored` arrays might also want  a name,

pytato/analysis/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    A :class:`pytato.transform.CachedWalkMapper` that records the number of
+    times an array expression is a direct dependency of other nodes.
+    """
+    def __init__(self) -> None:
+        from collections import defaultdict
+        super().__init__()
+        self.nusers: Dict[Array, int] = defaultdict(lambda: 0)
+
+    def map_index_lambda(self, expr: IndexLambda) -> None:
```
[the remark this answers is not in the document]
kaushikcfd: Yep, agreed. (Thanks!)

pytato/analysis/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    def map_matrix_product(self, expr: MatrixProduct) -> None:
+        self.nusers[expr.x1] += 1
+        self.nusers[expr.x2] += 1
+        return super().map_matrix_product(expr)
+
+
+def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
+    """
+    For the DAG *outputs*, returns the mapping from each node to its number of
+    successors.
```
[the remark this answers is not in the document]
kaushikcfd: Thanks!

pytato/target/loopy/codegen.py
```diff
@@ -368,19 +369,31 @@ def map_index_lambda(self, expr: IndexLambda,
...
+        # {{{ implementation tag
+
+        if expr.tags_of_type(ImplStored):
+            name = state.var_name_gen("_pt_temp")
+            result = StoredResult(name, expr.ndim,
+                                  frozenset([add_store(name, expr,
+                                                       result, state,
+                                                       self, True)]))
+        # TODO: Support more tags
+
```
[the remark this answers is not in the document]
kaushikcfd: Done, thanks

pytato/transform.py
```diff
@@ -632,6 +644,255 @@ def post_visit(self, expr: Any) -> None:
...
+    def __call__(self, expr: ArrayOrNames) -> ArrayOrNames:  # type: ignore[override]
+        return self.rec(expr)
+
+# }}}
+
+
+# {{{ MPMS materializer
+
+@dataclass(frozen=True, eq=True)
+class MPMSMaterializerContext:
```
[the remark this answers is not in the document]
kaushikcfd: Went with `MPMSMaterializerAccumulator` as the CombineMapper has a reduction-like pattern in its logic. Please lmk if you don't like it.

pytato/transform.py
```diff
@@ -632,6 +644,255 @@ def post_visit(self, expr: Any) -> None:
...
+        all_preds = rec_array.materialized_predecessors
+        new_expr = BasicIndex(rec_array.expr, expr.indices, expr.tags)
+
+        if self.nsuccessors[expr] > 1 and len(all_preds) > 1:
+            new_expr = new_expr.tagged(ImplStored())
+            return MPMSMaterializerContext(frozenset([new_expr]), new_expr)
+        else:
+            return MPMSMaterializerContext(all_preds, new_expr)
+
+    def map_contiguous_advanced_index(self,
```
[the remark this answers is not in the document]
kaushikcfd: Yep, all the MPMS related logic is at a single spot now.

Issue #52, opened 2021-03-14
pytato.array is too long


inducer:
It should be broken up into pieces, maybe

- namespace, "base" array and `DictOf`
- tags
- specific node types
- numpy-workalike

@kaushikcfd What do you think?

PR #133, opened 2021-08-11
Split pytato.array

kaushikcfd:
Moves
- math related interfaces into `pytato.cmath`
- reduction related interface into `pytato.reductions`.

For best reviewing experience please look at the patch on a commit-by-commit basis.

### DIFF
--- a/pytato/cmath.py
+++ b/pytato/cmath.py
@@ -1,0 +1,203 @@
+from __future__ import annotations
+
+__copyright__ = """
+Copyright (C) 2020 Andreas Kloeckner
+Copyright (C) 2020 Matt Wala
+Copyright (C) 2020 Xiaoyu Wei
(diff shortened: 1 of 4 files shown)


pytato/cmath.py
```diff
@@ -0,0 +1,203 @@
+from __future__ import annotations
```
inducer, line 1: Why `cmath` (as opposed to, say, `ufunc` or `elwise`)?

kaushikcfd: I thought a lot of routines would fall under `pt.unfuc` and was shooting for a finer-grained separation.

inducer: Sure. It's internal anyway, so why not. It's better than nothing. :)

inducer:
x-ref: #52

inducer:
LGTM. Thanks for working on this!

PR #134, opened 2021-08-12
Turn off implicit numpy-array handling in generated loopy

inducer:
For a moderate cost savings in the wrapper.

### DIFF
--- a/pytato/target/loopy/codegen.py
+++ b/pytato/target/loopy/codegen.py
@@ -859,7 +859,7 @@ def generate_loopy(result: Union[Array, DictOfNamedArrays, Dict[str, Array]],
     compute_order = preproc_result.compute_order
 
     if options is None and result_is_dict:
-        options = lp.Options(return_dict=True)
+        options = lp.Options(return_dict=True, no_numpy=True)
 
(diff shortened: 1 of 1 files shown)


kaushikcfd (commented):
I actually like `no_numpy=False` for testing/debugging as a default.

kaushikcfd:
Another thing being: I'm scared the amount of tests this will break. :)

inducer:
> I actually like `no_numpy=False` for testing/debugging as a default.

I'm torn. At best, it should be a user-option. (but then likely in the array context, which should set it to `False`)

> Another thing being: I'm scared the amount of tests this will break. :)

We'll find out together.

kaushikcfd:
> At best, it should be a user-option. (but then likely in the array context, which should set it to False)

I agree that we should set this option in `arraycontext`. `pt.generate_loopy` already takes in `options` argument. (cf [here](https://documen.tician.de/pytato/codegen.html?highlight=generate_loopy#pytato.generate_loopy))

inducer:
Superseded by https://github.com/inducer/arraycontext/pull/72 and https://github.com/inducer/pytato/pull/135.

PR #140, opened 2021-08-20
Add TopoSortMapper

matthiasdiener:


pytato/transform.py
```diff
@@ -559,6 +560,21 @@ def rec(self, expr: ArrayOrNames) -> None:  # type: ignore
 # }}}
 
 
+# {{{ TopoSortMapper
+
+class TopoSortMapper(CachedWalkMapper):
```
inducer, line 565: Add to docs.

inducer (approved):
LGTM. Just needs to be added to the docs, then ready to land!

pytato/transform.py
matthiasdiener: Added in https://github.com/inducer/pytato/pull/140/commits/50b33e14d0dc13ead4b69e16724749bf8821f16b

PR #141, opened 2021-08-20
show addr of node in visualization

matthiasdiener:
Makes it easier to match specific nodes to the dot graph


kaushikcfd (approved):
I can see that doing a Ctrl-F makes it easier to search for a node while debugging. So fine by me.

PR #154, opened 2021-09-20
Implements pt.broadcast_to

kaushikcfd:


kaushikcfd:
Setting this to "automerge" as its fairly trivial code and shouldn't break existing users. If anyone wishes to take a look later, feel free to put suggestions here. Those could be addressed in a separate PR.

inducer (commented):
LGTM

PR #158, opened 2021-09-30
Fix reshape on scalars/empty arrays

### LINKED ISSUE #157: Reshape on empty arrays
The following case doesn't work in `pytato` but `numpy` happily reshapes it to anything.

```python
x = pt.make_placeholder("x", shape=(0,), dtype=np.float64)
assert pt.reshape(x, (128, 0, 17)).shape == (128, 0, 17)
```

and 

```python
x = pt.make_placeholder("x", shape=(), dtype=np.float64)
pt.reshape(x, (1,))
```
seems to fail at code generation in [here](https://github.com/inducer/pytato/blob/7fcf1a65be0873f9eea991db74713e64685bf9f9/pytato/codegen.py#L411).

xref: inducer/arraycontext#91

kaushikcfd:
Closes #157

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -572,6 +572,8 @@ class Array(Taggable):
 
     def reshape(self, *shape: Union[int, Sequence[int]], order: str = "C") -> Array:
         import pytato as pt
+        if len(shape) == 0:
+            raise ValueError("reshape takes at least one argument (0 given)")
         if len(shape) == 1:
             # handle shape as single argument tuple
             return pt.reshape(self, shape[0], order=order)
@@ -1832,6 +1834,9 @@ def reshape(array: Array, newshape: Union[int, Sequence[int]],
     if newshape.count(-1) > 1:
         raise ValueError("can only specify one unknown dimension")
 
+    if newshape.count(-1) == 1 and newshape.count(0) > 0:
+        raise ValueError(f"cannot reshape {array.shape} into {newshape}")
+
     if not all(isinstance(axis_len, int) for axis_len in array.shape):
         raise ValueError("reshape of arrays with symbolic lengths not allowed")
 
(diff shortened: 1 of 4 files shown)


pytato/array.py
```diff
@@ -1844,7 +1847,7 @@ def reshape(array: Array, newshape: Union[int, Sequence[int]],
         if not isinstance(new_axislen, int):
             raise ValueError("Symbolic reshapes not allowed.")
 
-        if not(new_axislen > 0 or new_axislen == -1):
+        if not(new_axislen >= 0 or new_axislen == -1):
             raise ValueError("newshape should be either sequence of positive ints or"
```
alexfikl, line 1851:
```suggestion
            raise ValueError("newshape should be either sequence of non-negative ints or"
```
?

alexfikl (commented):
Looks good to me! Thanks for taking care of it so quickly!

pytato/array.py
```diff
@@ -572,6 +572,8 @@ def imag(self) -> ArrayOrScalar:
 
     def reshape(self, *shape: Union[int, Sequence[int]], order: str = "C") -> Array:
         import pytato as pt
+        if len(shape) == 0:
+            raise ValueError("reshape takes at least one argument (0 given)")
```
inducer, line 576:
```suggestion
            raise TypeError("reshape takes at least one argument (0 given)")
```

pytato/array.py
```diff
@@ -1844,9 +1849,9 @@ def reshape(array: Array, newshape: Union[int, Sequence[int]],
         if not isinstance(new_axislen, int):
             raise ValueError("Symbolic reshapes not allowed.")
 
-        if not(new_axislen > 0 or new_axislen == -1):
-            raise ValueError("newshape should be either sequence of positive ints or"
-                    " -1")
+        if new_axislen < -1:
+            raise ValueError("newshape should be either sequence of non-negative"
```
inducer, line 1853:
```suggestion
            raise ValueError("newshape should be either a sequence of non-negative"
```

inducer (commented):
Thanks, LGTM! I'll apply these two minor things myself.

PR #159, opened 2021-10-04
Loopy default offset

kaushikcfd:
- Minor: also adds an early substitution exit for identity substitutions.

/cc @thomasgibson


inducer:
Thanks!

thomasgibson (approved):
Thanks!!

PR #166, opened 2021-10-13
Equality comparison mapper

### LINKED ISSUE #163: Equality comparison of DAGs can have exponential complexity
Every time expressions are reused, they are re-traversed, since `__eq__` has no mechanism/means to memoize that expressions previously compared equal. (And we probably wouldn't want such a cache on the instance. Or maybe we do? Not sure.) I'm thinking we could introduce an `EqualityComparisonMapper` that would facilitate such memoization. To avoid having the cat biting its tail, it obviously couldn't use the arrays themselves as hash keys (since that would require equality comparison). But `id()` is available, and we could back that up with `a is b` as a check. An (IMO manageably) unfortunate consequence of this would be that we'd have to reimplement equality comparison of all types that can contain `Array` types, notably `dict`.

@mtcam is hitting this with cache retrievals upon freeze when exercising lazy evaluation for chemistry in https://github.com/illinois-ceesd/mirgecom.

cc @kaushikcfd

kaushikcfd:
Closes #163


pytato/transform.py
```diff
@@ -632,6 +634,187 @@ def post_visit(self, expr: Any) -> None:
 # }}}
 
 
+# {{{ EqualityComparisonMapper
```
inducer, line 637: Maybe move to `pytato.equality`? (Since it's not really much of a transform.)

pytato/transform.py
```diff
@@ -632,6 +634,187 @@ def post_visit(self, expr: Any) -> None:
...
+
+class EqualityComparisonMapper(Mapper):
```
inducer, line 639: Does this inherit anything of consequence from `Mapper`? Not inheriting would save you a bunch of type-ignores.

pytato/transform.py
```diff
@@ -632,6 +634,187 @@ def post_visit(self, expr: Any) -> None:
...
+    def handle_unsupported_array(self, expr1: Array,  # type: ignore[override]
+                                 expr2: Any) -> bool:
+        raise NotImplementedError
+
+    # type-ignored because method signature incompatible with the super class
+    def map_foreign(self, expr1: Any, expr2: Any) -> bool:  # type: ignore[override]
+        raise NotImplementedError
+
+    def map_placeholder(self, expr1: Placeholder, expr2: Any) -> bool:
+        return (isinstance(expr2, Placeholder)
```
inducer, line 682: `isinstance` isn't precise enough, since the object might compare equal to its subclass. I think the types need to match. (Yes, that was a bug in the original `__eq__`.)

pytato/transform.py
```diff
@@ -658,6 +841,14 @@ def get_dependencies(expr: DictOfNamedArrays) -> Dict[str, FrozenSet[Array]]:
 
     return {name: dep_mapper(val.expr) for name, val in expr.items()}
 
+
+def is_structurally_equivalent(expr1: ArrayOrNames, expr2: Any) -> bool:
```
inducer, line 845: Why have this wrapper?

pytato/transform.py
```diff
@@ -632,6 +634,187 @@ def post_visit(self, expr: Any) -> None:
...
+    def map_foreign(self, expr1: Any, expr2: Any) -> bool:  # type: ignore[override]
+        raise NotImplementedError
+
+    def map_placeholder(self, expr1: Placeholder, expr2: Any) -> bool:
+        return (isinstance(expr2, Placeholder)
+                and expr1.name == expr2.name
+                and expr1.shape == expr2.shape
+                and expr1.dtype == expr2.dtype
+                and expr1.tags == expr2.tags
+                )
```
inducer, line 687: What do you think? In principle, we could write that's as short and as generic as the original `__eq__`, but it would do a lot more `isinstance(..., Array)`, to see if it needs to recurse. Which version do you like better?

inducer (commented):
Thanks for working on this! A few questions below, but generally this looks good.

pytato/transform.py
inducer: Decision: Keep it type-specialized, the way it is.

pytato/transform.py
kaushikcfd: Went with `expr1.__class__ is expr2.__class__`

pytato/transform.py
kaushikcfd: Actually, no. Not inheriting anything. Removed along with all the `type-ignore`s. (thanks!)

pytato/transform.py
kaushikcfd: Agreed, Moved.

pytato/equality.py
```diff
@@ -0,0 +1,237 @@
...
+            self._cache[cache_key] = result
+            return result
+
+    def __call__(self, expr1: ArrayOrNames, expr2: Any
+                 ) -> bool:
+        return self.rec(expr1, expr2)
+
+    def handle_unsupported_array(self, expr1: Array,
+                                 expr2: Any) -> bool:
+        raise NotImplementedError
```
inducer, line 83: I think it'd be good to have an intelligible messages here.

pytato/equality.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def map_foreign(self, expr1: Any, expr2: Any) -> bool:
+        raise NotImplementedError
```
inducer, line 86: I think it'd be good to have an intelligible messages here.

pytato/equality.py
```diff
@@ -0,0 +1,237 @@
...
+                          BasicIndex, Concatenate, DataWrapper, Einsum,
+                          IndexBase, IndexLambda, MatrixProduct, NamedArray,
+                          Reshape, Roll, Stack, AbstractResultWithNamedArrays,
+                          Array, DictOfNamedArrays, Placeholder, SizeParam)
+
+if TYPE_CHECKING:
+    from pytato.loopy import LoopyCall
+
+__doc__ = """
+.. currentmodule:: pytato.equality
```
inducer, line 39:
```suggestion
```
(not needed)

pytato/equality.py
```diff
@@ -0,0 +1,237 @@
...
+                          AdvancedIndexInNoncontiguousAxes, AxisPermutation,
+                          BasicIndex, Concatenate, DataWrapper, Einsum,
+                          IndexBase, IndexLambda, MatrixProduct, NamedArray,
+                          Reshape, Roll, Stack, AbstractResultWithNamedArrays,
+                          Array, DictOfNamedArrays, Placeholder, SizeParam)
+
+if TYPE_CHECKING:
+    from pytato.loopy import LoopyCall
+
+__doc__ = """
```
inducer, line 38: Should this be hooked into the external docs somewhere?

pytato/equality.py
```diff
@@ -0,0 +1,237 @@
...
+    def map_dict_of_named_arrays(self, expr1: DictOfNamedArrays, expr2: Any) -> bool:
+        return (expr1.__class__ is expr2.__class__
+                and frozenset(expr1._data.keys()) == frozenset(expr2._data.keys())
+                and all(self.rec(expr1._data[name], expr2._data[name])
+                        for name in expr1._data))
+
+# }}}
+
+
+def is_structurally_equivalent(expr1: ArrayOrNames, expr2: Any) -> bool:
```
inducer, line 225: * Not loving the name. `equivalent` suggests something more restrictive than equal.
* Why not just call `EqualityComparisonMapper()(expr1, expr2)` in `__eq__`?

pytato/equality.py
```diff
@@ -0,0 +1,237 @@
...
+.. autofunction:: is_structurally_equivalent
+"""
+
+
+ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
+
+
+# {{{ EqualityComparisonMapper
+
+class EqualityComparisonMapper:
```
inducer, line 51:
```suggestion
class EqualityComparer:
```
?

pytato/equality.py
```diff
@@ -0,0 +1,237 @@
...
+
+def is_structurally_equivalent(expr1: ArrayOrNames, expr2: Any) -> bool:
+    """
+    Returns *True* only if *expr1* is equal to *expr2*.
+
+
+    .. note::
+
+        Complexity: :math:`O(N)`, where :math:`N` is the number of nodes in
+        *expr1*.
```
inducer, line 233: This could get moved to the `EqualityComparer` docstring.

pytato/equality.py
```diff
@@ -0,0 +1,237 @@
...
+"""
+
+
+ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
+
+
+# {{{ EqualityComparisonMapper
+
+class EqualityComparisonMapper:
+    def __init__(self) -> None:
```
inducer, line 52: Add a docstring justifying why this object exists (to host caches for equality comparison, and that comparison might be exponential without it).

pytato/equality.py
kaushikcfd: Added the type name. Thanks!

pytato/equality.py
kaushikcfd: Done.

pytato/equality.py
kaushikcfd: Done.

PR #170, opened 2021-10-19
Make `pt.Array`'s axes taggable types

### LINKED ISSUE #61: Array axes should allow free-form tags
x-ref:
- https://github.com/inducer/meshmode/pull/143
- https://github.com/inducer/loopy/issues/265

kaushikcfd:
This allows the axes of the arrays to be tagged.

Closes #61

### DIFF
--- a/doc/design.rst
+++ b/doc/design.rst
@@ -177,6 +177,31 @@ Guidelines for tag use:
   - **Good:** This array is the result of differentiation.
   - **Bad:** Unroll the loops in the code computing this result.
 
+Metadata Propagation
+^^^^^^^^^^^^^^^^^^^^
+
+Metadata (i.e. tags) is expected to be varied across application domains, not
+just in its information content, but also in the rules which might govern the
+way in which it propagates across the DAG. As an example, consider a Finite
+Element Method solver that uses :mod:`pytato`-arrays to store the DOFs.
+
+- A viable use of the tagging system in the solver could be to tag an array
+  ``u`` to describe the first axis of an array denotes the mesh's element
+  indices. In this case, after performing an operation ``u_squared = u * u`` is
+  semantically valid to tag ``u_squared`` also with the tag describing that it's
+  first denotes the mesh's element indices.
+- Another viable use of the tagging system by the solver could be to encode the
+  physical quantity that the array is storing. If ``u`` is an array tagged with
+  metadata to describe that it represents the fluid's velocity. Then, the
+  operation ``u_squared = u * u`` should not propagate that ``u_squared`` is
+  also an array storing the fluid's velocity.
+
+
+As a result, we choose to not propagate metadata (either upon construction, or
+via a dedicated, universally applied pass) and instead supply tools (such as
+:class:`~pytato.transform.Mapper`) to permit users to implement their own
+tag/metadata propagation rules.
+
 Memory layout
 -------------
 
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -41,6 +41,7 @@ Array Interface
 ---------------
 
 .. autoclass:: Array
+.. autoclass:: Axis
 .. autoclass:: NamedArray
 .. autoclass:: DictOfNamedArrays
 .. autoclass:: AbstractResultWithNamedArrays
@@ -155,13 +156,14 @@ import operator
 from dataclasses import dataclass
 from typing import (
         Optional, Callable, ClassVar, Dict, Any, Mapping, Tuple, Union,
-        Protocol, Sequence, cast, TYPE_CHECKING, List, Iterator, TypeVar)
+        Protocol, Sequence, cast, TYPE_CHECKING, List, Iterator, TypeVar,
+        FrozenSet)
 
 import numpy as np
 import pymbolic.primitives as prim
 from pymbolic import var
 from pytools import memoize_method
-from pytools.tag import Taggable, TagsType
+from pytools.tag import Tag, Taggable, TagsType
 
 from pytato.scalar_expr import (ScalarType, SCALAR_CLASSES,
                                 ScalarExpression)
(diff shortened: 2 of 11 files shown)


kaushikcfd:
The downstream failures can be addressed by a PR in `arraycontext`. See https://github.com/inducer/arraycontext/pull/111.

pytato/array.py
```diff
@@ -1678,6 +1736,21 @@ def dtype(self) -> np.dtype[Any]:
 
 # {{{ end-user facing
 
+def _get_default_axes(ndim: int) -> AxesT:
+    return tuple(Axis(_) for _ in range(ndim))
```
inducer, line 1740:
```suggestion
    return tuple(Axis(i) for i in range(ndim))
```
`_` is convention for "unused", so I find it strange to see an unused thing used.

pytato/array.py
```diff
@@ -513,7 +541,8 @@ def _unary_op(self, op: Any) -> Array:
                 expr,
                 shape=self.shape,
                 dtype=self.dtype,
-                bindings=bindings)
+                bindings=bindings,
+                axes=_get_default_axes(self.ndim))
```
inducer, line 545: I guess the big question here and elsewhere is whether this should preserve axis tags. If nothing else, we should have an entry in the design doc about why we're not preserving them.

pytato/array.py
```diff
@@ -312,6 +315,19 @@ class NormalizedSlice:
...
+class Axis(Taggable):
+    """
+    A type for recording the information about an :class:`~pytato.Array`'s
+    axis.
+    """
+    tags: FrozenSet[Tag]
+
+    def copy(self, **kwargs: Any) -> Axis:
+        from dataclasses import replace
+        return replace(self, **kwargs)
```
inducer, line 328: A thing we could debate is whether the axis lengths belong in here as well (making `.shape` a computed attribute). What do you think? (This doesn't have to happen in this PR. It could become an issue to be resolved later.) I feel like we talked about this, but I don't totally remember the resolution.

pytato/loopy.py
```diff
@@ -89,9 +89,16 @@ def __contains__(self, name: object) -> bool:
...
+
         if name not in self._result_names:
             raise KeyError(name)
-        return LoopyCallResult(self, name)
+
+        return LoopyCallResult(self, name,
+                               axes=_get_default_axes(len(self
+                                                          ._entry_kernel
+                                                          .arg_dict[name]
+                                                          .shape)))
```
inducer, line 101: This could use (perhaps mildly filtered) tags from the Loopy array axis tags.

pytato/array.py
```diff
@@ -1332,26 +1382,27 @@ class AxisPermutation(IndexRemappingBase):
 
     .. attribute:: array
 
-    .. attribute:: axes
+    .. attribute:: iaxes
```
inducer, line 1385: Not loving this name. `permutation` maybe? Or `axis_permutation`? Also, it's a compat break, but it's not that avoidable.

pytato/array.py
```diff
@@ -1639,13 +1694,15 @@ def __init__(self,
             name: str,
             shape: ShapeType,
             dtype: np.dtype[Any],
+            axes: AxesT,
```
inducer, line 1697: Looks like this is causing quite a few downstream failures in `arraycontext`. We should probably change those direct constructor calls to `make_placeholder`.

pytato/array.py
kaushikcfd: Oops. that was supposed to be `Axis(frozenset())`. But now, I'm confused if we should make the axis dimension an attribute of the `Axis` class. Wdyt?

pytato/array.py
kaushikcfd: I'm tempted to leave a statement in the design doc saying "Any metadata propagation logic isn't implemented in the frontend yet.". The reason being, the propagation implementation in the frontend won't be useful unless we thoroughly define the propagation system for every class of operation. (Added to the status in the docs here: 6f12dad7c38ffac23013342c431afaa6aad19451)

pytato/array.py
kaushikcfd: How do you feel about  making this go first and right after that https://github.com/inducer/arraycontext/pull/111 can go in?

pytato/array.py
kaushikcfd: Ok, thanks. Leaving as is.

pytato/array.py
kaushikcfd: Downstream failures should be address by https://github.com/inducer/arraycontext/pull/118.

pytato/array.py
kaushikcfd: @inducer: What do you think about  f80a1cc?

doc/design.rst
```diff
@@ -172,6 +172,16 @@ Guidelines for tag use:
...
 
+Metadata Propagation
+^^^^^^^^^^^^^^^^^^^^
+
+Given the user-specified nature of metadata associated with :mod:`pytato`
+arrays, it is almost impossible to classify how to correctly propagate
+:attr:`pytato.Array.tags`, :attr:`pytato.Array.axes` from inputs to outputs of
+primitive operations. Hence, the user must implement their own visitors (see
+:class:`~pytato.transform.Mapper`) to implement the desired domain-specific
+metadata propagation semantics.
```
inducer, line 183: https://notes.tiker.net/LBnyBLN9RFuhcIHlyZQmyw

kaushikcfd: Done, thanks!

inducer:
This seems to have a few CI failures still. To avoid downstream failures, please use a snippet like the following:

https://github.com/inducer/sumpy/blob/7bb8c8c070fc8e4f7317ee9dce96f96effbc633c/.github/workflows/ci.yml#L93-L97

kaushikcfd:
Sorry, this was lacking a rebase on top of `main`. Should be fixed now.

inducer:
LGTM, thanks!

PR #176, opened 2021-10-22
Explain why DataWrapper takes a separate shape

inducer:
cc @matthiasdiener

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1587,6 +1587,18 @@ class DataWrapper(InputArgumentBase):
         Starting with the construction of the :class:`DataWrapper`,
         this array may not be updated in-place.
 
+    .. attribute:: shape
+
+        The shape of the array is represented separately from array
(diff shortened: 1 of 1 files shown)


pytato/array.py
```diff
@@ -1598,10 +1612,13 @@ class DataWrapper(InputArgumentBase):
...
             name: Optional[str],
             data: DataInterface,
-            shape: ShapeType,
+            shape: ShapeType = None,
             tags: TagsType = frozenset()):
         super().__init__(name, tags=tags)
 
+        if shape is None:
+            shape = data.shape
+
```
kaushikcfd, line 1621: Not sure about this. We have generally tried to avoid input validation in the type's constructor.

kaushikcfd (changes requested):
Thanks! Looks good except one minor comment.

pytato/array.py
inducer: Fair point. I'll get rid of it.

inducer: 5795767

PR #177, opened 2021-10-25
Implements a routine for Index Lambda to High Level Op

kaushikcfd:
Q. Why is this useful?
A. Realizing how an IndexLambda came into existence is important for several reasons:
- Implementing metadata translation
- Allowing pattern matching on sub-expressions while exposing the high-level operations.
  - Raising to HLO also allows for richer pattern matching as we can match based on properties of the high level ops (like commutativity, associativity) 
- Implementing a Numpy-like target
- Implementing a smart `Array.__str__`.

### DIFF
--- a/doc/array.rst
+++ b/doc/array.rst
@@ -5,6 +5,12 @@ Array Expressions
 
 .. automodule:: pytato.array
 
+
+Raising :class:`~pytato.array.IndexLambda` nodes
+------------------------------------------------
+
+.. automodule:: pytato.raising
+
 Calling :mod:`loopy` kernels in an array expression
 ---------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


doc/array.rst
```diff
@@ -5,6 +5,12 @@ Array Expressions
 
 .. automodule:: pytato.array
 
+
+Raising :class:`~pytato.array.IndexLambda` nodes
+------------------------------------------------
+
+.. automodule:: pytato.normalization
```
inducer, line 12: Doc nomenclature and module name clash. Is it "normalization" or "raising"?

inducer (approved):
Thanks! This LGTM other than a quibble about naming.

doc/array.rst
kaushikcfd: I think "raising" makes more sense. I'll quickly change this.

kaushikcfd: Thanks, done: https://github.com/inducer/pytato/compare/0a8fb7ae1a0941c69ff1da12f7efbc80fb276de1..e113cc16d2b8b56763e47150cdee4fddeda8b85c

inducer:
LGTM. In it goes!

PR #179, opened 2021-10-28
DictOfNamedArrays: add __str__

matthiasdiener:


matthiasdiener:
Does this look useful @inducer?

pytato/array.py
```diff
@@ -784,6 +784,12 @@ def __eq__(self, other: Any) -> bool:
         from pytato.equality import EqualityComparer
         return EqualityComparer()(self, other)
 
+    def __repr__(self) -> str:
+        return "DictOfNamedArrays(" + str(self._data) + ")"
+
+    def __str__(self) -> str:
+        return self.__repr__()
```
inducer, line 791:
```suggestion
```
Not needed, see: https://docs.python.org/3/reference/datamodel.html#object.__repr__

inducer:
LGTM, thanks!

PR #186, opened 2021-11-01
Handle `Named` tags

kaushikcfd:


inducer:
I guess I'll need to roll a pytools release in order for this to pass? Let me do that.

kaushikcfd:
Aah sorry. This is weird, I thought we have [pytools as a requirement](https://github.com/inducer/pytato/blob/main/requirements.txt#L1)  so we should have been good.

inducer:
https://pypi.org/project/pytools/2021.2.9/#history

inducer:
(We'll still have to wait until it makes its way into conda forge. It's a pyopencl dep, that's why it's special.)

inducer (approved):
LGTM, thanks!

PR #188, opened 2021-11-01
Add random graph generation

matthiasdiener:
Closes #150.

- [x] Must wait for/contains #190.

PR #190, opened 2021-11-02
DataInterface: document dtype, shape as read-only

inducer:
This came out of adding type annotations for testlib in #188. I figured I'd put it up for discussion separately.


kaushikcfd (approved):
Thanks, TIL.

PR #196, opened 2021-11-05
Centralize shape/index handling in mappers

inducer:
cc @matthiasdiener

### DIFF
--- a/pytato/transform.py
+++ b/pytato/transform.py
@@ -174,8 +174,8 @@ class CopyMapper(CachedMapper[ArrayOrNames]):
        This does not copy the data of a :class:`pytato.array.DataWrapper`.
     """
 
-    def rec_idx_or_size_tuple(self, situp: Tuple[IndexOrShapeExpr, ...]) \
-            -> Tuple[IndexOrShapeExpr, ...]:
+    def rec_idx_or_size_tuple(self, situp: Tuple[IndexOrShapeExpr, ...]
+                              ) -> Tuple[IndexOrShapeExpr, ...]:
         return tuple(self.rec(s) if isinstance(s, Array) else s for s in situp)
 
     def map_index_lambda(self, expr: IndexLambda) -> Array:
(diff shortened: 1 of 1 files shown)


pytato/transform.py
```diff
@@ -173,22 +174,24 @@ class CopyMapper(CachedMapper[ArrayOrNames]):
        This does not copy the data of a :class:`pytato.array.DataWrapper`.
     """
 
+    def rec_idx_or_size_tuple(self, situp: Tuple[IndexOrShapeExpr, ...]) \
+            -> Tuple[IndexOrShapeExpr, ...]:
```
kaushikcfd, line 178:
```suggestion
    def rec_idx_or_size_tuple(self, situp: Tuple[IndexOrShapeExpr, ...]
                              ) -> Tuple[IndexOrShapeExpr, ...]:
```
?

pytato/transform.py
```diff
@@ -287,6 +284,10 @@ class CombineMapper(Mapper, Generic[CombineT]):
     def __init__(self) -> None:
         self.cache: Dict[ArrayOrNames, CombineT] = {}
 
+    def rec_idx_or_size_tuple(self, situp: Tuple[IndexOrShapeExpr, ...]) \
+            -> Tuple[CombineT, ...]:
```
kaushikcfd, line 288:
```suggestion
    def rec_idx_or_size_tuple(self, situp: Tuple[IndexOrShapeExpr, ...]
                              ) -> Tuple[CombineT, ...]:
```
?

pytato/transform.py
```diff
@@ -1185,10 +1179,14 @@ class EdgeCachedMapper(CachedMapper[ArrayOrNames], ABC):
     def handle_edge(self, expr: ArrayOrNames, child: ArrayOrNames) -> Any:
         pass
 
-    def _handle_shape(self, expr: Array, shape: Any, *args: Any) -> Tuple[Any, ...]:
+    def rec_idx_or_size_tuple(self,
+            expr: Array,
+            situp: Tuple[IndexOrShapeExpr, ...],
+            *args: Any) \
+            -> Tuple[IndexOrShapeExpr, ...]:
```
kaushikcfd, line 1186:
```suggestion
            *args: Any) -> Tuple[IndexOrShapeExpr, ...]:
```
?

kaushikcfd (approved):
Thanks, LGTM. Only one minor formatting issue from [PEP 8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length).
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

However, I am not sure if we want to abide to this in Pytato.

PR #199, opened 2021-11-05
Make `.name` mandatory in Placeholder and SizeParam

inducer:
cc @matthiasdiener

### DIFF
--- a/pytato/codegen.py
+++ b/pytato/codegen.py
@@ -605,20 +605,18 @@ class NamesValidityChecker(CachedWalkMapper):
         super().__init__()
 
     def post_visit(self, expr: Any) -> None:
-        if isinstance(expr, InputArgumentBase):
-            if expr.name is None:
-                # Name to be automatically assigned
-                return
-
-            try:
-                ary = self.name_to_input[expr.name]
-            except KeyError:
-                self.name_to_input[expr.name] = expr
-            else:
-                if ary is not expr:
-                    from pytato.diagnostic import NameClashError
-                    raise NameClashError("Received two separate instances of inputs "
-                                         f"named '{expr.name}'.")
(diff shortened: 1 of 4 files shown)


pytato/codegen.py
```diff
@@ -605,11 +605,7 @@ def __init__(self) -> None:
         super().__init__()
 
     def post_visit(self, expr: Any) -> None:
-        if isinstance(expr, InputArgumentBase):
-            if expr.name is None:
-                # Name to be automatically assigned
-                return
-
+        if isinstance(expr, (Placeholder, SizeParam)):
```
kaushikcfd, line 608: What if the user specifies 2 non-identical data wrappers with the same name? I think that's still a good reason to raise.

pytato/target/loopy/codegen.py
```diff
@@ -898,7 +898,7 @@ def generate_loopy(result: Union[Array, DictOfNamedArrays, Dict[str, Array]],
     state.var_name_gen.add_names({input_expr.name
             for name in compute_order
             for input_expr in ing(outputs[name].expr)
-            if input_expr.name is not None})
+            if isinstance(input_expr, (Placeholder, SizeParam))})
```
kaushikcfd, line 901: User specified data wrapper names are still something we don't want to collide with.

kaushikcfd (approved):
Thanks, the logic of an :attr:`Placeholder.name` having the type `Optional[str]` was flaky. It's great to not hit those.

pytato/codegen.py
inducer: cde46d1185dddc8e748dcc26d74dca5c923bd514

pytato/target/loopy/codegen.py
inducer: cde46d1185dddc8e748dcc26d74dca5c923bd514

PR #200, opened 2021-11-06
Expose some more classes under pytato module

kaushikcfd:
Wanted to write a mapper not within pytato module and `pt.TypeName` is quite handy for specifying type names.


matthiasdiener (approved):
Thanks!

kaushikcfd:
Seems pretty harmless. Let me merge it.

PR #202, opened 2021-11-07
Fix arc functions on scalars

alexfikl:
For scalars, it would call `numpy` directly, which uses arcsin not asin.


inducer:
Thanks!

PR #203, opened 2021-11-08
fix reverse_graph

matthiasdiener:


pytato/transform.py
```diff
@@ -1131,6 +1131,10 @@ def reverse_graph(graph: Dict[ArrayOrNames, Set[ArrayOrNames]]) \
     result: Dict[ArrayOrNames, Set[ArrayOrNames]] = {}
 
     for node_key, edges in graph.items():
+        # Make sure every node is in the result even if it has no users
+        if node_key not in result:
+            result[node_key] = set()
```
inducer, line 1136:
```suggestion
        result.setdefault(node_key, set())
```

pytato/transform.py
```diff
@@ -1131,6 +1131,10 @@ def reverse_graph(graph: Dict[ArrayOrNames, Set[ArrayOrNames]]) \
     result: Dict[ArrayOrNames, Set[ArrayOrNames]] = {}
 
     for node_key, edges in graph.items():
+        # Make sure every node is in the result even if it has no users
```
inducer, line 1134: Add a regression test for this.

inducer (commented):
LGTM with these two changes.

kaushikcfd (approved):
Looks good, thanks!

pytato/transform.py
matthiasdiener: bd8c336

PR #205, opened 2021-11-09
Implements DimToLoopyInameTag

kaushikcfd:
Draft because:
- [x] Requires array axes to be taggable via #170

### DIFF
--- a/pytato/tags.py
+++ b/pytato/tags.py
@@ -12,6 +12,7 @@ Pre-Defined Tags
 .. autoclass:: Named
 .. autoclass:: PrefixNamed
 .. autoclass:: AssumeNonNegative
+.. autoclass:: DimToLoopyInameTag
 """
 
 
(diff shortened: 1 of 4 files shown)


pytato/tags.py
```diff
@@ -104,3 +105,23 @@ class AssumeNonNegative(Tag):
...
+    whose axis is tagged with this tag is allocated, then the iname responsible
+    for indexing the axis in the allocation instruction would be tagged with
+    :attr:`iname_tag`.
+
+    .. attribute:: iname_tag
+
+        The tag that :class:`loopy.kernel.data.Iname` would be tagged with
+        for the iname indexing the axis.
+    """
+    iname_tag: Tag
```
inducer, line 125: Can you describe how this is used? If it's being used to apply iname implementation tags (like `g.0` or so), then that seems weird, because pytato/array metadata should be *descriptive* ("this is a FEM element index") IMO, not *prescriptive* ("parallelize over this axis"). If it is being used for descriptive things, why the additional packaging? Why can't we just generically copy over all tags (or at least ones that pytato isn't "digesting") from array axes to inames?

kaushikcfd: > because pytato/array metadata should be descriptive 

It's upto the user what they want to tag the iname with, my current use case is for the descriptive purposes. See [this in meshmode](https://github.com/kaushikcfd/meshmode/blob/d75f243eb93c0f1fc983faaa93907a939b819187/meshmode/array_context.py#L1006-L1020).

> Why can't we just generically copy over all tags (or at least ones that pytato isn't "digesting") from array axes to inames?

I felt this way of copying all the tags that aren't pytato-implementation tags isn't transparent enough. For ex. the user might tag an axis to notify the code-generator what the iname-name for the corresponding axis should be, and keeping this tag in the generated iname tag would simply amount to noise.

inducer: As discussed today, pending further insights, maybe we'll have this head towards propagating tags by default.

kaushikcfd: Yep that's slightly cleaner. Moved the discussion to #229.

PR #207, opened 2021-11-09
replace MatrixProduct with Einsum

### LINKED ISSUE #149: Kill `MatrixProduct` now that we have `Einsum`?

matthiasdiener:
Fixes #149 .

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1601,7 +1601,6 @@ def matmul(x1: Array, x2: Array) -> Array:
     :param x1: first argument
     :param x2: second argument
     """
-    from pytato.utils import are_shape_components_equal
     if (
             isinstance(x1, SCALAR_CLASSES)
             or x1.shape == ()
@@ -1612,12 +1611,6 @@ def matmul(x1: Array, x2: Array) -> Array:
     assert isinstance(x1, Array)
     assert isinstance(x2, Array)
 
-    if len(x1.shape) > 2 or len(x2.shape) > 2:
-        raise NotImplementedError("broadcasting not supported")
-
-    if not are_shape_components_equal(x1.shape[-1], x2.shape[0]):
-        raise ValueError("dimension mismatch")
-
     import pytato as pt
 
     if x1.ndim == x2.ndim == 1:
@@ -1626,12 +1619,12 @@ def matmul(x1: Array, x2: Array) -> Array:
         return pt.sum(x1 * x2, axis=(x1.ndim - 1))
 
     idx_stream = (chr(i) for i in range(ord("i"), ord("z")))
-    idx_gen: Callable[[], str] = lambda: next(idx_stream)  # noqa: E731
+    idx_gen: Callable[[], str] = lambda: next(idx_stream)
+
     x1_indices = "".join(idx_gen() for _ in range(x1.ndim))
-    x2_indices = "".join(idx_gen() for _ in range(x2.ndim))
-    # reduce over second-to-last axis of *b* and last axis of *a*
-    x2_indices = x2_indices[:-2] + x1_indices[-1] + x2_indices[-1]
-    result_indices = x1_indices[:-1] + x2_indices[:-2] + x2_indices[-1]
+    x2_indices = x1_indices[:-2] + x1_indices[-1] + idx_gen()
+    result_indices = x1_indices[:-1] + x2_indices[-1]
+
     return pt.einsum(f"{x1_indices}, {x2_indices} -> {result_indices}", x1, x2)
 
 
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1685,10 +1685,7 @@ def matmul(x1: Array, x2: Array) -> Array:
     if x1.ndim == x2.ndim == 1:
         return pt.sum(x1 * x2)
     elif x1.ndim == 1:
-        x1_index = index_names[0]
-        x2_indices = index_names[:x2.ndim]
-        result_indices = x2_indices[1:]
-        return pt.einsum(f"{x1_index}, {x2_indices} -> {result_indices}", x1, x2)
+        return pt.dot(x1, x2)
     elif x2.ndim == 1:
         return pt.sum(x1 * x2, axis=(x1.ndim - 1))
 
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1608,9 +1608,6 @@ def matmul(x1: Array, x2: Array) -> Array:
             or x2.shape == ()):
         raise ValueError("scalars not allowed as arguments to matmul")
 
-    assert isinstance(x1, Array)
-    assert isinstance(x2, Array)
-
     import pytato as pt
 
     if x1.ndim == x2.ndim == 1:
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1680,20 +1680,19 @@ def matmul(x1: Array, x2: Array) -> Array:
 
     import pytato as pt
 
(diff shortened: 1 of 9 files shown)


pytato/array.py
```diff
@@ -1660,13 +1609,30 @@ def matmul(x1: Array, x2: Array) -> Array:
...
+    elif x2.ndim == 1:
+        return pt.sum(x1 * x2, axis=(x1.ndim - 1))
+
+    idx_stream = (chr(i) for i in range(ord("i"), ord("z")))
+    idx_gen: Callable[[], str] = lambda: next(idx_stream)  # noqa: E731
+    x1_indices = "".join(idx_gen() for _ in range(x1.ndim))
+    x2_indices = "".join(idx_gen() for _ in range(x2.ndim))
+    # reduce over second-to-last axis of *b* and last axis of *a*
+    x2_indices = x2_indices[:-2] + x1_indices[-1] + x2_indices[-1]
+    result_indices = x1_indices[:-1] + x2_indices[:-2] + x2_indices[-1]
```
kaushikcfd, line 1634: This tells me that `result.ndim = a.ndim  + b.ndim - 2`, which is not quite true.
I think we want
```python
x1_indices = "".join(idx_gen() for _ in range(x1.ndim))
x2_indices = x1_indices[:-1] + idx_gen()
result_indices = x1_indices[:-1] + x2_indices[-1]
```

If this is wrong, could you please add a test with `x1.ndim > 2` and `x2.ndim > 2`.

kaushikcfd (commented):
This needs some higher ndim tests. Otherwise, LGTM. Thanks!

pytato/array.py
matthiasdiener: Should this be:

> ```python
> x2_indices = x1_indices[-1] + idx_gen()
> ```

?

kaushikcfd: Aah yep, that should have been `x2_indices = x1_indices[:-2] + x1_indices[-1] + idx_gen()`.

kaushikcfd: (Note: this doesn't consider the case when x1.ndim != x2.ndim, that might need stripping off some of the indices on the array with lower dimensionality)

pytato/array.py
```diff
@@ -1660,13 +1609,30 @@ def matmul(x1: Array, x2: Array) -> Array:
             or x2.shape == ()):
         raise ValueError("scalars not allowed as arguments to matmul")
 
+    assert isinstance(x1, Array)
+    assert isinstance(x2, Array)
```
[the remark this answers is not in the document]
matthiasdiener: removed in 88f3f51

pytato/array.py
matthiasdiener: Thanks, this is hopefully fixed in 395fe50

kaushikcfd: Thanks! I think it still misses the broadcasting logic for the following inputs:
```python
>>> import numpy as np
>>> a = np.random.rand(9, 3, 11, 4)
>>> b = np.random.rand(3, 4, 5)
>>> (a @ b).shape
(9, 3, 11, 5)
```

matthiasdiener: Hmm, I'm not sure what the [matmul documentation](https://numpy.org/doc/stable/reference/generated/numpy.matmul.html) is trying to say for this case: 

> If either argument is N-D, N > 2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.

kaushikcfd: The last 2 axes will undergo matmuls while the rest will be broadcasted. So in the above example the last 2 axes of A `(11, 4)` will get matmul-ed with last 2 axes of B i.e. `(4, 5)` and the rest of the starting axes will be broadcasted i.e. `(9, 3)` and `(3,)` will result in `(9, 3)`.

So I think what we need is:
```python
idx_stream = (chr(i) for i in range(ord("l"), ord("z")))
idx_gen: Callable[[], str] = lambda: next(idx_stream)
stack_indies = "".join(idx_gen() for _ in range(max(x1.ndim-2, x2.ndim-2)))
x1_indices = stack_indices[len(stack_indices)-x1.ndim+2:] + "ij"
x2_indices = stack_indices[len(stack_indices)-x2.ndim+2:] + "jk"
result_indices = stack_indices + "ik"
```

test/test_codegen.py
```diff
@@ -106,6 +106,25 @@ def get_array(ndim):
...
+    x2 = pt.make_data_wrapper(x2_in)
+
+    prog = pt.generate_loopy(x1 @ x2, cl_device=queue.device)
+    _, (out,) = prog(queue)
+
+    np_out = np.matmul(x1_in, x2_in)
+
+    assert np_out.shape == out.shape
+    assert (np_out == out).all()
+
```
kaushikcfd, line 126: Needs another test here with 1-dimensional 'A' and multi-dimensional 'B'. That's another "special" case in the numpy's matmul spec.

pytato/array.py
matthiasdiener: Thanks! 76e983d should fix this case.

kaushikcfd: Thanks!

kaushikcfd (approved):

pytato/array.py
```diff
@@ -1652,21 +1601,31 @@ def matmul(x1: Array, x2: Array) -> Array:
...
         raise ValueError("scalars not allowed as arguments to matmul")
 
-    if len(x1.shape) > 2 or len(x2.shape) > 2:
-        raise NotImplementedError("broadcasting not supported")
+    import pytato as pt
+
+    if x1.ndim == x2.ndim == 1:
+        return pt.sum(x1 * x2)
+    elif x1.ndim == 1:
+        return pt.einsum("i, ij -> j", x1, x2)
```
kaushikcfd, line 1616: Sorry had missed this earlier, will fail if `x2.ndim > 2`.

test/test_codegen.py
matthiasdiener: Added in 8e7056f

pytato/array.py
matthiasdiener: Thanks, 8e7056f should fix this case.

kaushikcfd (approved):
LGTM, thank you!

matthiasdiener:
Would you like to take a look as well @inducer ?

test/test_codegen.py
```diff
@@ -105,6 +105,48 @@ def get_array(ndim):
     assert (out == x1_in @ x2_in).all()
 
 
+@pytest.mark.parametrize("x1_shape", ([7],))
```
inducer, line 108: Shapes should be tuples.

test/test_codegen.py
```diff
@@ -105,6 +105,48 @@ def get_array(ndim):
...
+@pytest.mark.parametrize("x2_shape", ([7], [7, 3], [7, 7, 4], [7, 7, 7, 9]))
+def test_matmul_dimone(ctx_factory, x1_shape, x2_shape):
```
inducer, line 110: Merge with below test.

test/test_codegen.py
```diff
@@ -105,6 +105,48 @@ def get_array(ndim):
...
+    ctx = ctx_factory()
+    queue = cl.CommandQueue(ctx)
+
+    x1_in = np.ones(x1_shape)
```
inducer, line 114: Use `random.randn(*x1_shape)`

inducer (approved):
Thanks! Subject to these comments: LGTM!

PR #213, opened 2021-11-16
Use MRO to find Array mapper methods

matthiasdiener:

PR #214, opened 2021-11-16
Use a separate mapper method for LoopyCallResult

matthiasdiener:


inducer:
@kaushikcfd I'm mainly hoping to hear from you on this, as this might interfere with some of your ongoing work, though the combination of this and #213 shouldn't be too bad: if `map_loopy_call_result` isn't available, it dispatches to `map_named_array`.

kaushikcfd:
I agree, both these PRs should restore the back compat. I don't see any big red signals. Good to go from my end. Thanks!

inducer:
Thanks! In it goes.

PR #217, opened 2021-11-17
Preserve High-Level Info in the lowered pymbolic expressions

kaushikcfd:
Pymbolic's `Expression.(__(r?)add__|__(r?)sub__|__(r?)mul__)` can lose the
high-level information of expression as seen in the example below.

```python
>>> from pymbolic import var
>>> x = var("x")
>>> x + 0
Variable('x')
>>> x * 1
Variable('x')
```

---
Also includes 2 commits that help pass the CI.

### DIFF
--- a/doc/design.rst
+++ b/doc/design.rst
@@ -54,6 +54,11 @@ Computation and Results
     with an integral :attr:`Array.dtype` (i.e. having ``dtype.kind == "i"``).
     Such an expression marks the boundary between eager and lazy evaluation.
 
+-   :attr:`Array.shape` is required to be an affine expression in terms of the
+    instances of :class:`~pytato.array.SizeParam` in the computation graph. This
+    permits shape inference to use Presburger arithmetic, meaning that shape
+    inference is always decidable.
+
 -   There is (deliberate) overlap in what various expression nodes can
     express, e.g.
 
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -536,25 +536,37 @@ class Array(Taggable):
                 bindings=bindings,
                 axes=_get_default_axes(self.ndim))
 
-    __mul__ = partialmethod(_binary_op, operator.mul)
-    __rmul__ = partialmethod(_binary_op, operator.mul, reverse=True)
+    # NOTE: Initializing the expression to "prim.Product(expr1, expr2)" is
+    # essential as opposed to performing "expr1 * expr2". This is to account
+    # for pymbolic's implementation of the "*" operator which might not
+    # instantiate the node corresponding to the operation when one of
+    # the operands is the neutral element of the operation.
+    #
+    # For the same reason 'prim.(Sum|FloorDiv|Quotient)' is preferred over the
+    # python operators on the operands.
 
-    __add__ = partialmethod(_binary_op, operator.add)
-    __radd__ = partialmethod(_binary_op, operator.add, reverse=True)
+    __mul__ = partialmethod(_binary_op, lambda l, r: prim.Product((l, r)))
+    __rmul__ = partialmethod(_binary_op, lambda l, r: prim.Product((l, r)),
+                             reverse=True)
 
-    __sub__ = partialmethod(_binary_op, operator.sub)
-    __rsub__ = partialmethod(_binary_op, operator.sub, reverse=True)
+    __add__ = partialmethod(_binary_op, lambda l, r: prim.Sum((l, r)))
+    __radd__ = partialmethod(_binary_op, lambda l, r: prim.Sum((l, r)),
+                             reverse=True)
 
-    __floordiv__ = partialmethod(_binary_op, operator.floordiv)
-    __rfloordiv__ = partialmethod(_binary_op, operator.floordiv, reverse=True)
+    __sub__ = partialmethod(_binary_op, lambda l, r: prim.Sum((l, -r)))
+    __rsub__ = partialmethod(_binary_op, lambda l, r: prim.Sum((l, -r)),
+                             reverse=True)
 
-    __truediv__ = partialmethod(_binary_op, operator.truediv,
+    __floordiv__ = partialmethod(_binary_op, prim.FloorDiv)
+    __rfloordiv__ = partialmethod(_binary_op, prim.FloorDiv, reverse=True)
+
+    __truediv__ = partialmethod(_binary_op, prim.Quotient,
             get_result_type=_truediv_result_type)
-    __rtruediv__ = partialmethod(_binary_op, operator.truediv,
+    __rtruediv__ = partialmethod(_binary_op, prim.Quotient,
             get_result_type=_truediv_result_type, reverse=True)
 
-    __pow__ = partialmethod(_binary_op, operator.pow)
-    __rpow__ = partialmethod(_binary_op, operator.pow, reverse=True)
+    __pow__ = partialmethod(_binary_op, prim.Power)
+    __rpow__ = partialmethod(_binary_op, prim.Power, reverse=True)
 
     __neg__ = partialmethod(_unary_op, operator.neg)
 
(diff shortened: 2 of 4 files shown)


pytato/array.py
```diff
@@ -536,25 +536,28 @@ def _unary_op(self, op: Any) -> Array:
...
+    __truediv__ = partialmethod(_binary_op, prim.Quotient,
             get_result_type=_truediv_result_type)
-    __rtruediv__ = partialmethod(_binary_op, operator.truediv,
+    __rtruediv__ = partialmethod(_binary_op, prim.Quotient,
             get_result_type=_truediv_result_type, reverse=True)
 
-    __pow__ = partialmethod(_binary_op, operator.pow)
-    __rpow__ = partialmethod(_binary_op, operator.pow, reverse=True)
+    __pow__ = partialmethod(_binary_op, prim.Power)
+    __rpow__ = partialmethod(_binary_op, prim.Power, reverse=True)
```
inducer, line 560: For context: why is it important that this information be preserved? (Reply with a comment in the code.)

pytato/utils.py
```diff
@@ -265,19 +265,20 @@ def are_shape_components_equal(dim1: ShapeComponent, dim2: ShapeComponent) -> bo
...
+    from pytato.transform import InputGatherer
 
-    dim1_expr = to_expr(dim1)
-    dim2_expr = to_expr(dim2)
-    # ScalarExpression.__eq__  returns Any
-    return (distribute(dim1_expr-dim2_expr) == 0)  # type: ignore
+    # type-ignore reason: not all InputArgumentBase have a name attr.
+    space = _create_size_param_space({expr.name  # type: ignore[attr-defined]
+                                      for expr in InputGatherer()(dim1_minus_dim2)})
+    aff = ShapeToISLExpressionMapper(space)(dim1_minus_dim2)
```
inducer, line 279: Add a comment explaining why these are all expected to be affine.

inducer (commented):
Thanks for working on this! Two requests for improved comments below.

pytato/utils.py
kaushikcfd: I just wrote that pytato requires it, although I don't see mention about it in the design docs.

The only current functioning target (loopy) requires the shapes to be affine expressions and if we intend to have a guarantee that the shape inference solver would result in a unique solution or no solution I think this becomes a must.

pytato/array.py
kaushikcfd: Almost wrote what's in the description of this PR.

pytato/utils.py
inducer: Then let's document that requirement!

pytato/array.py
inducer: I still don't understand, other than it might have something to do with shape inference. Could you try and explain the actual reason in the comment?

kaushikcfd: Oh this doesn't have anything to do with shape inference (sorry for squashing 2 things in the same PR, I thought the PR's description might help untangling it). We need to retain the ability to raise index lambdas to their numpy ops. However, pymbolic is slightly smarter when it comes to identities in binary ops, making it difficult to raise index lambdas cleanly. This PR tries to avoid pymbolic from acting smartly.

```python
>>> import pymbolic.primitives as prim
>>> x = prim.Variable("x")
>>> x + 0
Variable('x')  # cannot infer expression came from "Plus" op.
>>> prim.Sum((x, 0))
prim.Sum((Variable('x'), 0)) # CAN infer expression came from "Plus" op.
```

So I think the explanation already provided in the comments was fine, please let me know if it wasn't.

pytato/utils.py
kaushikcfd: Yep!
Done in 8961569.

kaushikcfd:
@inducer: The CI is failing here at `rm -rf .miniforge3/envs/testing/x86_64-conda-linux-gnu/sysroot`. The return code seems like a permission issue (`EPERM`), any idea?

kaushikcfd:
Tis' not permission issues but the numpy version replacement[[1]] in the ci-script is incorrect. Because `grep` returns 1 when it sees no match.

[1]: https://github.com/inducer/ci-support/commit/0c2b9c0f7b2923d469c2725769f1c23273686752

pytato/array.py
```diff
@@ -536,25 +536,37 @@ def _unary_op(self, op: Any) -> Array:
...
-
-    __floordiv__ = partialmethod(_binary_op, operator.floordiv)
-    __rfloordiv__ = partialmethod(_binary_op, operator.floordiv, reverse=True)
-
-    __truediv__ = partialmethod(_binary_op, operator.truediv,
+    # NOTE: Initializing the expression to "prim.Product(expr1, expr2)" is
+    # essential as opposed to performing "expr1 * expr2". This is to account
+    # for pymbolic's implementation of the "*" operator which loses high level
+    # information in the output expression when one of the operands is
+    # identity.
```
inducer, line 543:
```suggestion
    # for pymbolic's implementation of the "*" operator which might not
    # instantiate the node corresponding to the operation when one of 
    # the operands is the neutral element of the operation.
```

pytato/array.py
inducer: Sorry for being dense here. Why is it important that a `+ 0` be preserved?

doc/design.rst
```diff
@@ -54,6 +54,11 @@ Computation and Results
     with an integral :attr:`Array.dtype` (i.e. having ``dtype.kind == "i"``).
     Such an expression marks the boundary between eager and lazy evaluation.
 
+-   :attr:`Array.shape` is required to be an affine expression in terms of the
+    instances of :class:`~pytato.array.SizeParam` in the computation graph. This
+    was done to enjoy an exact shape inference solver during the code-generation
+    of the computation graph.
```
inducer, line 60:
```suggestion
    permits shape inference to use Presburger arithmetic, meaning that shape inference
    is always decidable.
```

inducer (commented):
Thanks! Two more phrasing things and a question on rationale.

pytato/array.py
kaushikcfd: This was done to tackle some of the issues that come up in #177. So the options we have are:
1. Avoid introducing ambiguity and keep the high-level-op to be interpreted back to the NumPy-like ops used to build the DAG, or,
2. Don't promise that raising to the high-level ops in #177 would be what the user provided, instead these binary operations with identities should have been interpreted as [`BroadcastOp`](https://github.com/kaushikcfd/pytato/blob/64b90ddfeeab9f8542b71dbca46db34d3b35a32f/pytato/normalization.py#L82).

I'm shooting for (1) here, but happy here if this might hurt us in the future.

inducer: I think the promise of exactly preserving what the user gave us is (for now) flimsy at best, at least without additional changes to pymbolic. For example, the default `map_sum` and `map_product` in `IdentityMapper` [will flatten](https://github.com/inducer/pymbolic/blob/7434c58242542a4f84e334062ed8de7eb79a9145/pymbolic/mapper/__init__.py#L422-L438) the expression. I'm happy to discuss getting rid of that that, but without that, I think (1) is unattainable.

kaushikcfd: Looking at the [high-level op nodes](https://github.com/kaushikcfd/pytato/blob/64b90ddfeeab9f8542b71dbca46db34d3b35a32f/pytato/normalization.py#L39-L103), I don't think flattening the sums/products would have any effect on the index lambda as we won't have more than 2 children in a pymbolic expression that came through the pytato frontend.

doc/design.rst
kaushikcfd: Thanks, see fece1eb.

pytato/array.py
kaushikcfd: Thanks, done in 544cc71.

kaushikcfd:
After discussing with @inducer in a personal chat, it was concluded that the current approach is a bit too hacky and must be enforced on the pymbolic end and so closing this.

(The commits regarding shapes being affine are still valuable, and I'll open them as a separate PR. (Sorry for the noise!))

pytato/array.py
kaushikcfd: See https://github.com/inducer/pytato/pull/217#issuecomment-1018584244 for conclusion.

inducer:
It's not that the approach was hacky, it's that pymbolic currently can't currently be trusted to preserve operations as well as this claims, so the extra effort that this goes to seems unnecessary.

PR #222, opened 2021-11-24
Defines `pt.tags.AssumeNonNegative`

### LINKED ISSUE #221: [RFC] Adding metadata to avoid modulo in indirect addressing
## Problem
For the advanced indexing `a[b]` in the following pytato code,
```python
a = pt.make_placeholder("a", 10, np.float64)
b = pt.make_placeholder("b", 10, np.int32)
print(pt.generate_loopy(a[b]).program)
```

we generate the following code:
```python
---------------------------------------------------------------------------
INSTRUCTIONS:
for _pt_out_dim0
  _pt_out[_pt_out_dim0] = a[b[_pt_out_dim0] % 10]  {id=_pt_out_store}
end _pt_out_dim0
---------------------------------------------------------------------------
```
This is inefficient and in certain cases is ~saturating the modulo unit~ the temporaries for the modulo are occupying significant register space on the device costing us 50% of the runtime costs for computing the RHS in https://github.com/illinois-ceesd/drivers_y1-nozzle.

## Proposed Solution
A tag `AssumeNonNegative()` that would inform the code generator to assume all entries in the array to be positive, which can then be used to generate code without the modulo op. The original code would now look as:

```python
a = pt.make_placeholder("a", 10, np.float64)
b = pt.make_placeholder("b", 10, np.int32).tagged(pt.tags.AssumeNonNegative())
print(pt.generate_loopy(a[b]).program)
```
and should generate code without any modulos in it.

kaushikcfd:
Implementation to go along with #221.

Closes #221.

### DIFF
--- a/pytato/codegen.py
+++ b/pytato/codegen.py
@@ -42,6 +42,7 @@ from pytato.scalar_expr import ScalarExpression, IntegralScalarExpression
 from pytato.transform import CopyMapper, CachedWalkMapper, SubsetDependencyMapper
 from pytato.target import Target
 from pytato.loopy import LoopyCall
+from pytato.tags import AssumeNonNegative
 from pytools import UniqueNameGenerator
 import loopy as lp
 SymbolicIndex = Tuple[IntegralScalarExpression, ...]
@@ -499,12 +500,16 @@ class CodeGenPreprocessor(CopyMapper):
                 if isinstance(axis_len, int):
                     bnd_name = vng("in")
                     bindings[bnd_name] = self.rec(idx)
-                    indices.append(prim.Subscript(
-                                        prim.Variable(bnd_name),
-                                        get_indexing_expression(
-                                                idx.shape,
-                                                (1,)*i_adv_indices[0]+adv_idx_shape))
-                                   % axis_len)
+                    indirect_idx_expr = prim.Subscript(
+                        prim.Variable(bnd_name),
+                        get_indexing_expression(
+                            idx.shape,
+                            (1,)*i_adv_indices[0]+adv_idx_shape))
+
+                    if not idx.tags_of_type(AssumeNonNegative):
+                        indirect_idx_expr = indirect_idx_expr % axis_len
+
+                    indices.append(indirect_idx_expr)
                 else:
                     raise NotImplementedError("Advanced indexing over"
                                               " parametric axis lengths.")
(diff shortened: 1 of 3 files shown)


pytato/codegen.py
```diff
@@ -499,12 +500,20 @@ def map_contiguous_advanced_index(self,
...
+                            prim.Variable(bnd_name),
+                            get_indexing_expression(idx.shape,
+                                                    (1,)*i_adv_indices[0]
+                                                    + adv_idx_shape))
+                    else:
+                        indirect_idx_expr = (prim.Subscript(
+                            prim.Variable(bnd_name),
+                            get_indexing_expression(idx.shape,
+                                                    (1,)*i_adv_indices[0]
+                                                    + adv_idx_shape))
```
inducer, line 514: Could you avoid the code duplication here, so that only the modulo sits behind the `if`?

pytato/codegen.py
```diff
@@ -555,11 +564,19 @@ def map_non_contiguous_advanced_index(self,
...
+                        indirect_idx_expr = prim.Subscript(
+                            prim.Variable(bnd_name),
+                            get_indexing_expression(idx.shape,
+                                                    adv_idx_shape))
+                    else:
+                        indirect_idx_expr = (prim.Subscript(
+                            prim.Variable(bnd_name),
+                            get_indexing_expression(idx.shape,
+                                                    adv_idx_shape))
+                                             % axis_len)
```
inducer, line 578: Could you avoid the code duplication here, so that only the modulo sits behind the `if`?

pytato/tags.py
```diff
@@ -94,3 +95,12 @@ class PrefixNamed(_BaseNameTag):
     prefix: str
 
 # }}}
+
+
+@tag_dataclass
+class AssumeNonNegative(Tag):
+    """
+    Tagged to an :class:`~pytato.Array` to indicate the
```
inducer, line 103: "Tagged to" is not grammatical in English AFAIK. How about "A tag used on..."?

test/test_codegen.py
```diff
@@ -1381,6 +1381,35 @@ def get_partition_id(topo_list, expr) -> MyPartitionId:
...
 
 
+def test_assume_non_negative_indirect_address(ctx_factory):
+    from numpy.random import default_rng
+    from pytato.scalar_expr import WalkMapper
+
+    ctx = ctx_factory()
+    cq = cl.CommandQueue(ctx)
+
+    class RemainderCatcher(WalkMapper):
```
inducer, line 1391:
```suggestion
    class OnRemainderRaiser(WalkMapper):
```
?

inducer (commented):
Thanks! LGTM, aside from very minor wrinkles.

inducer:

test/test_codegen.py
kaushikcfd: Thanks, see 635cdcf42232e7700e396f82d086c7b17094245b.

pytato/tags.py
kaushikcfd: Thanks, fixed in 7c90c29.

pytato/codegen.py
kaushikcfd: Yep, thanks, much cleaner now.

pytato/codegen.py
kaushikcfd: Thanks, fixed.

inducer:
Thx!

PR #224, opened 2021-11-30
Introduce attributes to store information about reduction dimensions

kaushikcfd:
Draft because:
- [x] Depends on https://github.com/inducer/pytato/pull/217
- [x] Depends on #205

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -960,18 +960,35 @@ class Einsum(Array):
 
        A :class:`tuple` of array over which the Einstein summation is being
        performed.
+
+    .. attribute:: access_descr_to_index
+
+       Mapping from the access descriptors to the index used by the user during
+       the instantiation of the :class:`Einsum` node. This is a strictly
+       non-semantic attribute and only present to support a friendlier
+       :meth:`with_tagged_reduction`.
+
+    .. automethod:: with_tagged_reduction
     """
-    _fields = Array._fields + ("access_descriptors", "args")
+    _fields = Array._fields + ("access_descriptors",
+                               "args",
+                               "redn_axis_to_redn_descr",
+                               "index_to_access_descr")
     _mapper_method = "map_einsum"
 
     def __init__(self,
                  access_descriptors: Tuple[Tuple[EinsumAxisDescriptor, ...], ...],
                  args: Tuple[Array, ...],
                  axes: AxesT,
+                 redn_axis_to_redn_descr: PMap[EinsumReductionAxis,
+                                              ReductionDescriptor],
+                 index_to_access_descr: PMap[str, EinsumAxisDescriptor],
                  tags: FrozenSet[Tag] = frozenset()):
         super().__init__(axes=axes, tags=tags)
         self.access_descriptors = access_descriptors
         self.args = args
+        self.redn_axis_to_redn_descr = redn_axis_to_redn_descr
+        self.index_to_access_descr = index_to_access_descr
 
     @memoize_method
     def _access_descr_to_axis_len(self
@@ -1023,6 +1040,52 @@ class Einsum(Array):
     def dtype(self) -> np.dtype[Any]:
         return np.find_common_type(array_types=[arg.dtype for arg in self.args],
                                     scalar_types=[])
+
+    def with_tagged_reduction(self,
+                              redn_axis: Union[EinsumReductionAxis, str],
+                              tag: Tag) -> Einsum:
+        """
+        Returns a copy of *self* with the :class:`ReductionDescriptor`
+        associated with *redn_axis* tagged with *tag*.
+        """
+        from pytato.diagnostic import InvalidEinsumIndex, NotAReductionAxis
+        # {{{ sanity checks
+
+        if isinstance(redn_axis, str):
+            try:
+                redn_axis_ = self.index_to_access_descr[redn_axis]
+            except KeyError:
+                raise InvalidEinsumIndex(f"'{redn_axis}': not a valid axis index.")
+            if isinstance(redn_axis_, EinsumReductionAxis):
+                redn_axis = redn_axis_
+            else:
+                raise NotAReductionAxis(f"'{redn_axis}' is not"
+                                        " a reduction axis.")
+        elif isinstance(redn_axis, EinsumReductionAxis):
+            pass
+        else:
+            raise TypeError("Argument 'redn_axis' expected to be"
+                            f" EinsumReductionAxis, got {type(redn_axis)}")
+
+        if redn_axis in self.redn_axis_to_redn_descr:
+            assert any(redn_axis in access_descrs
+                       for access_descrs in self.access_descriptors)
+        else:
+            raise ValueError(f"{redn_axis}: does not appear as a"
+                             " reduction access descriptor.")
+
+        # }}}
+
+        new_redn_axis_to_redn_descr = self.redn_axis_to_redn_descr.set(
+            redn_axis, self.redn_axis_to_redn_descr[redn_axis].tagged(tag))
+
+        return type(self)(access_descriptors=self.access_descriptors,
+                          args=self.args,
+                          axes=self.axes,
+                          redn_axis_to_redn_descr=new_redn_axis_to_redn_descr,
+                          tags=self.tags,
+                          index_to_access_descr=self.index_to_access_descr,
+                          )
 
 
 EINSUM_FIRST_INDEX = re.compile(r"^\s*((?P<alpha>[a-zA-Z])|(?P<ellipsis>\.\.\.))\s*")
@@ -1159,12 +1222,17 @@ def _normalize_einsum_in_subscript(subscript: str,
             index_to_descr, index_to_axis_length)
 
 
-def einsum(subscripts: str, *operands: Array) -> Einsum:
+def einsum(subscripts: str, *operands: Array,
+           index_to_redn_descr: Optional[Mapping[str, ReductionDescriptor]] = None
+           ) -> Einsum:
     """
     Einstein summation *subscripts* on *operands*.
     """
     if len(operands) == 0:
         raise ValueError("must specify at least one operand")
+
+    if index_to_redn_descr is None:
+        index_to_redn_descr = {}
 
     if "->" not in subscripts:
         # implicit-mode: output spec matched by alphabetical ordering of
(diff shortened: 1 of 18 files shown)


test/test_pytato.py
```diff
@@ -392,6 +392,78 @@ def construct_intestine_graph(depth=100, seed=0):
...
+
+    x_sum = x_sum.with_tagged_redn_dim(1, FooRednTag())
+    hlo = index_lambda_to_high_level_op(x_sum)
+    assert x_sum.reduction_dims[hlo.axes[1]].tags_of_type(FooRednTag)
+    assert not x_sum.reduction_dims[hlo.axes[0]].tags_of_type(FooRednTag)
+
+    x_trace = pt.einsum("ii->i", x)
+    x_colsum = pt.einsum("ij->j", x)
+
+    with pytest.raises(ValueError):
```
inducer, line 448: Maybe make a custom exception type for this? (`ValueError` is fairly unspecific.)

pytato/array.py
```diff
@@ -884,18 +954,32 @@ class Einsum(Array):
 
        A :class:`tuple` of array over which the Einstein summation is being
        performed.
+
+    .. attribute:: access_descr_to_index
+
+       Mapping from the access descriptors to the index used by the user during the
+       instantiation of the :class:`Einsum` node.
```
inducer, line 961: Maybe turn these into attributes of `EinsumAxisDescriptor`? State there that this is strictly non-semantic feature, and that they are optional.

test/test_pytato.py
kaushikcfd: Yep, done. Thanks!

inducer:
It seems this has picked up a bunch of conflicts...

kaushikcfd:
>  It seems this has picked up a bunch of conflicts...

Thanks, fixed!

pytato/array.py
```diff
@@ -1046,27 +1054,51 @@ def dtype(self) -> np.dtype[Any]:
         return np.find_common_type(array_types=[arg.dtype for arg in self.args],
                                     scalar_types=[])
 
-    def with_tagged_redn_dim(self, redn_descr: ReductionDim, tag: Tag) -> Einsum:
+    def with_tagged_redn_dim(self,
```
inducer, line 1057: I think we should always spell this out as `reduction`, not `redn`. `with_tagged_reduction`?

pytato/array.py
```diff
@@ -322,6 +323,19 @@ def copy(self, **kwargs: Any) -> Axis:
         return replace(self, **kwargs)
 
 
+@dataclass(eq=True, frozen=True)
+class ReductionDim(Taggable):
```
inducer, line 329: I'm not sure I love the name. Maybe `ReductionDescriptor`?

pytato/array.py
```diff
@@ -851,12 +861,65 @@ def __init__(self,
...
         self.bindings = bindings
+        self.reduction_dims = reduction_dims
+
+    def with_tagged_redn_dim(self,
+                             redn_axis: Union[str, int],
+                             tag: Tag) -> IndexLambda:
+        """
+        Returns a copy of *self* with the :class:`ReductionDim` associated with
+        *redn_variable* tagged with *tag*.
+        """
```
inducer, line 879: Docs speak of `redn_variable` and do not describe the `int` case.

pytato/array.py
```diff
@@ -840,9 +843,16 @@ class IndexLambda(_SuppliedShapeAndDtypeMixin, Array):
         expressions available for use in
         :attr:`expr`.
 
+    .. attribute:: reduction_dims
```
inducer, line 846: `var_to_reduction_descr`? (Also elsewhere for consistency.)

pytato/array.py
```diff
@@ -851,12 +861,65 @@ def __init__(self,
...
+
+    def with_tagged_redn_dim(self,
+                             redn_axis: Union[str, int],
+                             tag: Tag) -> IndexLambda:
+        """
+        Returns a copy of *self* with the :class:`ReductionDim` associated with
+        *redn_variable* tagged with *tag*.
+        """
+        from pytato.diagnostic import NotAReductionDim
+        if isinstance(redn_axis, int):
```
inducer, line 881: The handling of the `int` case feels like it wants to be a function. I don't think it belongs here.

pytato/reductions.py
```diff
@@ -178,9 +179,14 @@ def _normalize_reduction_axes(
...
-                Sequence[ScalarExpression],
-                Dict[str, Tuple[ScalarExpression, ScalarExpression]]]:
+                                  axes: Tuple[int, ...],
+                                  axis_to_reduction_dim: Optional[
+                                      Mapping[int,
+                                              ReductionDim]] = None
+                                  ) -> Tuple[Sequence[ScalarExpression],
+                                             Dict[str, Tuple[ScalarExpression,
+                                                             ScalarExpression]],
+                                             PMap[str, ReductionDim]]:
```
inducer, line 189: Docstring does not describe what `axis_to_reduction_dim` adds here.

pytato/array.py
```diff
@@ -976,18 +976,24 @@ class Einsum(Array):
...
     def __init__(self,
                  access_descriptors: Tuple[Tuple[EinsumAxisDescriptor, ...], ...],
                  args: Tuple[Array, ...],
                  axes: AxesT,
+                 redn_descr_to_redn_dim: PMap[EinsumReductionAxis, ReductionDim],
                  tags: FrozenSet[Tag] = frozenset()):
         super().__init__(axes=axes, tags=tags)
         self.access_descriptors = access_descriptors
         self.args = args
+        self.redn_descr_to_redn_dim = redn_descr_to_redn_dim
```
inducer, line 1003: * `reduction_axis_to_reduction_descr`?
* `reduction_axis_to_descr`?

pytato/array.py
```diff
@@ -851,12 +861,65 @@ def __init__(self,
...
+            reduction_dims: PMap[str, ReductionDim],
             tags: FrozenSet[Tag] = frozenset()):
 
         super().__init__(shape=shape, dtype=dtype, axes=axes, tags=tags)
 
         self.expr = expr
         self.bindings = bindings
+        self.reduction_dims = reduction_dims
+
+    def with_tagged_redn_dim(self,
```
inducer, line 873: `with_tagged_reduction`?

pytato/diagnostic.py
```diff
@@ -52,3 +52,16 @@ class UnknownIndexLambdaExpr(ValueError):
...
+    """
+    Raised when an einsum index was referred by an invalid value.
+    """
+
+
+class NotAReductionDim(ValueError):
+    """
+    Raised when a :class:`pytato.ReductionDim` was referred by an invalid
+    value.
+    """
```
inducer, line 67: Are these used anywhere?

inducer (commented):
Thanks! A few comments below.

inducer:

pytato/array.py
kaushikcfd: Went with `redn_axis_to_redn_descr`.

pytato/diagnostic.py
kaushikcfd: Yes, these are used, but not in the same commit they were defined in.

pytato/array.py
kaushikcfd: Yep, thanks!

pytato/array.py
kaushikcfd: Yep, that's more explicit. Done, thanks!

pytato/array.py
kaushikcfd: Done, thanks!

pytato/array.py
kaushikcfd: On thinking a bit more, I think the `int` was presumptuous in expecting the reduction in the index lambda to be one of `pt.sum`, `pt.prod`, etc. Got rid of it. Now we only accept the name of the reduction variable i.e. the `str` case.

pytato/reductions.py
kaushikcfd: Done!

pytato/array.py
kaushikcfd: Yes, agreed very much. See https://github.com/inducer/pytato/pull/224#discussion_r880843037.

pytato/array.py
kaushikcfd: Making them attributes of `EinsumAxisDescriptor` demands extra synchronization while instantiating a `pt.Einsum` as they appear times depending on the index usage. Added a note in the docs about the non-semantic feature.

pytato/array.py
```diff
@@ -963,6 +1041,53 @@ def dtype(self) -> np.dtype[Any]:
...
+        elif isinstance(redn_axis, EinsumReductionAxis):
+            pass
+        else:
+            raise TypeError("Argument 'redn_axis' expected to be"
+                            f" EinsumReductionAxis, got {type(redn_axis)}")
+
+        if redn_axis in self.redn_axis_to_redn_descr:
+            assert any(any(access_descr == redn_axis
+                           for access_descr in access_descrs)
+                       for access_descrs in self.access_descriptors)
```
inducer, line 1073: :green_circle: No need to nest the `any` here.

pytato/array.py
```diff
@@ -899,18 +960,35 @@ class Einsum(Array):
...
+                               "index_to_access_descr")
     _mapper_method = "map_einsum"
 
     def __init__(self,
                  access_descriptors: Tuple[Tuple[EinsumAxisDescriptor, ...], ...],
                  args: Tuple[Array, ...],
                  axes: AxesT,
+                 redn_axis_to_redn_descr: PMap[EinsumReductionAxis,
+                                              ReductionDescriptor],
+                 index_to_access_descr: PMap[str, EinsumAxisDescriptor],
```
inducer, line 985: :green_circle: Conceivably, we should just say `Mapping` here. (`MutableMapping` is distinct.) It'd be nice if there were an `ImmutableMapping`, because we'd like to enforce immutability here...

inducer, line 985: :green_circle: As a somewhat related point, these are possibly better off being `immutables.Map` in concrete use. I'm a bit puzzled why we're introducing new uses of `PMap`.

pytato/reductions.py
```diff
@@ -178,18 +179,44 @@ def _normalize_reduction_axes(
 
 
 def _get_reduction_indices_bounds(shape: ShapeType,
```
inducer, line 181: :yellow_circle: `var_to_redn_descr` ostensibly has nothing to do with "bounds" as this function claims to return. I'd prefer if this were done outside this function.

inducer (commented):
Thanks! Just few minor things here, then this is good to go.

Question: Do you want this squashed or rebased?

Legend for my PR comments:

- :red_circle: Must fix, no two ways about it.
- :yellow_circle: Kinda annoyed by this, but I can probably be talked into looking past it.
- :green_circle: Minor nit, doesn't matter one way or the other.

pytato/diagnostic.py
```diff
@@ -52,3 +52,16 @@ class UnknownIndexLambdaExpr(ValueError):
...
     pass
+
+
+class InvalidEinsumIndex(ValueError):
+    """
+    Raised when an einsum index was referred by an invalid value.
+    """
+
+
+class NotAReductionDescriptor(ValueError):
```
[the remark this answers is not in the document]
kaushikcfd: Done, thanks!

pytato/array.py
kaushikcfd: Done, thanks!

kaushikcfd: Introduced a commit to replace most of these occurrences with `immutables.Map`. Thanks!

> I'm a bit puzzled why we're introducing new uses of PMap.

This PR was opened before I got to know about `immutables.Map` :)

pytato/array.py
kaushikcfd: Done, replaced the inner `any` with a  ` redn_axis in access_descrs`. Thank you!

pytato/reductions.py
kaushikcfd: Yep, agreed, that was misplaced. Made it a new function of its own.

kaushikcfd:
> Question: Do you want this squashed or rebased?

The commit messages are still clean enough, however, the weird dependency between the commits would make them unusable for git-bisect. I wouldn't mind squashing them.

inducer:
Thanks! I was not a huge fan of importing `immutables.Map` as `PMap`, I thought that made the code confusing to read. I've backed that out, and I've also added a dependency on `immutables` in `setup.py`. LGTM now, thanks for working on this. Setting to merge+squash.

kaushikcfd:
> Thanks! I was not a huge fan of importing immutables.Map as PMap,

No complaints there. Thanks!

PR #225, opened 2021-12-03
Fix Downstream MIRGE-Com CI (?)

kaushikcfd:
See: https://github.com/illinois-ceesd/mirgecom/pull/316

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,7 +129,7 @@ jobs:
                 if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
                     # can't turn off MPI in mirgecom
                     sudo apt-get update
-                    sudo apt-get install openmpi-bin libopenmpi-dev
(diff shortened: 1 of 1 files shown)


kaushikcfd:
Hmm.. the same link error exists here too. Looks like the linker cannot find `libc`'s shared objects.

matthiasdiener:
We'll need to apply the same workaround as here (using OpenMPI should be fine): https://github.com/illinois-ceesd/mirgecom/pull/525/files#diff-464a640b719a59f4671d01f644ed750894b34669d94035fbe89038c3d7018510R25

matthiasdiener:
> We'll need to apply the same workaround as here (using OpenMPI should be fine): https://github.com/illinois-ceesd/mirgecom/pull/525/files#diff-464a640b719a59f4671d01f644ed750894b34669d94035fbe89038c3d7018510R25

This should probably be done in ci-support-v0

kaushikcfd:
@matthiasdiener: Thanks for taking a look here! I see, will open a PR over there then.

PR #229, opened 2021-12-07
Propagate axis tags to loopy iname tags

kaushikcfd:
Propagate all `pt.Array.axis` tags as iname tags.


inducer:
LGTM, thanks for working on this!

PR #230, opened 2021-12-08
Adds a helper routine to check einsum subscripts

kaushikcfd:
/cc @matthiasdiener

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -24,7 +24,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import Mapping, Dict, Union, Set
+from typing import Mapping, Dict, Union, Set, Tuple
 from pytato.array import (Array, IndexLambda, Stack, Concatenate, Einsum,
                           DictOfNamedArrays, NamedArray,
                           IndexBase, IndexRemappingBase, InputArgumentBase)
@@ -35,6 +35,8 @@ __doc__ = """
 .. currentmodule:: pytato.analysis
 
 .. autofunction:: get_nusers
+
+.. autofunction:: is_einsum_similar_to_subscript
 """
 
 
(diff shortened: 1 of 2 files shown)


pytato/analysis/__init__.py
```diff
@@ -147,3 +149,74 @@ def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
...
+
+        # }}}
+
+    return tuple(normalized_indices)
+
+
+def is_einsum_similar_to_subscript(expr: Einsum, subscript: str) -> bool:
+    """
+    Returns *True* only if the *expr* is an einsum expression that was
+    instantiated via a subscript that's alpha-equivalent to *subscript*.
```
inducer, line 185: Spell out what "alpha-equivalent" means.

kaushikcfd: How does the patch https://github.com/inducer/pytato/compare/062cd34096518062b3f2453836520e92f0bb8e4a..3b19c45a798e3e9ebf2fa5841609908de02e0ac8 look?

matthiasdiener (approved):
LGTM, thanks!

pytato/analysis/__init__.py
```diff
@@ -147,3 +149,78 @@ def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
...
+                             f" '{subscript}'.")
+
+        # }}}
+
+    return tuple(normalized_indices)
+
+
+def is_einsum_similar_to_subscript(expr: Einsum, subscript: str) -> bool:
+    """
+    Returns *True* only if *expr* is an einsum expression that was instantiated
```
inducer, line 184: * Did you write "only if" on purpose here? What's missing from "if and only if"?
* To make this easier to think about, I would like to this to say, "an einsum with the subscript descriptor string *subscript_descr* would compute the same result as *expr*`." Is that something that we can claim? That would allow us to get rid of the clumsy concept of "alpha-equivalence."

pytato/analysis/__init__.py
```diff
@@ -147,3 +149,78 @@ def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
...
+    if not isinstance(expr, Einsum):
+        raise TypeError(f"{expr} expected to be Einsum, got {type(expr)}.")
+
+    if "->" not in subscript:
+        raise NotImplementedError("Comparing against implicit mode einsums:"
+                                  " not supported.")
+
+    in_spec, out_spec = subscript.split("->")
+
+    index_to_descrs: Dict[str, EinsumAxisDescriptor] = {}
```
inducer, line 203:
```suggestion
    # build up a mapping from index names to axis descriptors
    index_to_descrs: Dict[str, EinsumAxisDescriptor] = {}
```

pytato/analysis/__init__.py
```diff
@@ -147,3 +149,78 @@ def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
...
+
+    for idim, idx in enumerate(_get_indices_from_input_subscript(out_spec)):
+        index_to_descrs[idx] = ElementwiseAxis(idim)
```
inducer, line 206: Check that there are no repetitions in the `out_spec`? There shouldn't be in order for it to be a meaningful `einsum`, but it can't hurt to check.

pytato/analysis/__init__.py
```diff
@@ -147,3 +149,78 @@ def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
...
+            return False
+
+        for idx, access_descr in zip(indices, access_descrs):
+            try:
+                if index_to_descrs[idx] != access_descr:
+                    return False
+            except KeyError:
+                if not isinstance(access_descr, ReductionAxis):
+                    return False
+                index_to_descrs[idx] = access_descr
```
inducer, line 224: Maybe capture this logic in a (local) function `try_add_subscript_mapping`?

inducer (commented):
Looks good generally, just a few things below.

inducer:
Also, thanks for working on this!

pytato/analysis/__init__.py
```diff
@@ -147,3 +149,78 @@ def get_nusers(outputs: Union[Array, DictOfNamedArrays]) -> Mapping[Array, int]:
...
+            raise ValueError(f"Cannot parse '{acc}' in provided einsum"
+                             f" '{subscript}'.")
+
+        # }}}
+
+    return tuple(normalized_indices)
+
+
+def is_einsum_similar_to_subscript(expr: Einsum, subscript: str) -> bool:
+    """
```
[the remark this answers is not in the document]
kaushikcfd: `np.einsum` calls it `subscripts` (yep was a typo, fixed) so I am hesitant to call it something else.

pytato/analysis/__init__.py
kaushikcfd: Yep we can claim the definition you propose. It's cleaner and I put that in here. Thanks!

pytato/analysis/__init__.py
kaushikcfd: Thanks, done!

pytato/analysis/__init__.py
kaushikcfd: Sure, added that.

pytato/analysis/__init__.py
kaushikcfd: Sorry wasn't quite liking it. It was making seem the 10-line logic more complicated than it is. Added some comments, hopefully that helps?

kaushikcfd:
@inducer: A bit more context on why this is needed. If we want to apply special transformations to einsums with particular `subscripts` (say materialization strategy specific to particular einisum types), this utility might come handy.

inducer:
LGTM! Thanks for working on this, it'll land once CI passes.

PR #238, opened 2021-12-28
Introduce `ReductionOperation` class, accept 'initial' in reductions

inducer:
This replaces the strings we used to use to identify reductions with something a bit more structured. The impulse for this came from https://github.com/inducer/arraycontext/pull/129, which needs to supply `initial` to array reductions to match the `numpy` interface for empty min/max reductions. This only adds very minimal support for supplying `initial`, in that it allows the neutral element to be passed (and obtaining that is what led to the structured reduction op types), and it better mimics numpy's behavior for empty reductions.

cc @majosm

### DIFF
--- a/pytato/reductions.py
+++ b/pytato/reductions.py
@@ -27,7 +27,11 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import Optional, Tuple, Union, Sequence, Dict, List
+from typing import Any, Optional, Tuple, Union, Sequence, Dict, List
+from abc import ABC, abstractmethod
+
+import numpy as np
+
 from pytato.array import ShapeType, Array, make_index_lambda
 from pytato.scalar_expr import ScalarExpression, Reduce, INT_CLASSES
 import pymbolic.primitives as prim
@@ -43,7 +47,93 @@ __doc__ = """
 .. autofunction:: prod
 .. autofunction:: all
 .. autofunction:: any
+
+.. currentmodule:: pytato.reductions
+
+.. autoclass:: ReductionOperation
+.. autoclass:: SumReductionOperation
+.. autoclass:: ProductReductionOperation
+.. autoclass:: MaxReductionOperation
+.. autoclass:: MinReductionOperation
+.. autoclass:: AllReductionOperation
+.. autoclass:: AnyReductionOperation
 """
+
+# }}}
+
+
+class _NoValue:
+    pass
+
+
+# {{{ reduction operations
+
+class ReductionOperation(ABC):
+    """
+    .. automethod:: neutral_element
+    .. automethod:: __hash__
+    .. automethod:: __eq__
+    """
+
+    @abstractmethod
+    def neutral_element(self, dtype: np.dtype[Any]) -> Any:
+        pass
+
+    @abstractmethod
+    def __hash__(self) -> int:
+        pass
+
+    @abstractmethod
+    def __eq__(self, other: Any) -> bool:
+        pass
+
+
+class _StatelessReductionOperation(ReductionOperation):
+    def __hash__(self) -> int:
+        return hash(type(self))
+
+    def __eq__(self, other: Any) -> bool:
+        return type(self) is type(other)
+
+
+class SumReductionOperation(_StatelessReductionOperation):
+    def neutral_element(self, dtype: np.dtype[Any]) -> Any:
+        return 0
+
+
+class ProductReductionOperation(_StatelessReductionOperation):
+    def neutral_element(self, dtype: np.dtype[Any]) -> Any:
+        return 1
+
+
+class MaxReductionOperation(_StatelessReductionOperation):
+    def neutral_element(self, dtype: np.dtype[Any]) -> Any:
+        if dtype.kind == "f":
+            return dtype.type(float("-inf"))
+        elif dtype.kind == "i":
+            return np.iinfo(dtype).min
+        else:
+            raise TypeError(f"unknown neutral element for max and {dtype}")
+
+
+class MinReductionOperation(_StatelessReductionOperation):
+    def neutral_element(self, dtype: np.dtype[Any]) -> Any:
+        if dtype.kind == "f":
+            return dtype.type(float("inf"))
+        elif dtype.kind == "i":
+            return np.iinfo(dtype).max
+        else:
+            raise TypeError(f"unknown neutral element for min and {dtype}")
+
+
+class AllReductionOperation(_StatelessReductionOperation):
+    def neutral_element(self, dtype: np.dtype[Any]) -> Any:
+        return np.bool_(True)
+
+
+class AnyReductionOperation(_StatelessReductionOperation):
+    def neutral_element(self, dtype: np.dtype[Any]) -> Any:
+        return np.bool_(False)
 
 # }}}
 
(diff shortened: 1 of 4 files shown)


pytato/reductions.py
```diff
@@ -151,76 +262,106 @@ def _make_reduction_lambda(op: str, a: Array,
...
+        if the reduction is empty.
+        In that case, the reduction size must not be symbolic.
     """
-    return _make_reduction_lambda("min", a, axis)
+    return _make_reduction_lambda(MinReductionOperation(), a, axis, initial)
 
 
-def prod(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+def prod(a: Array, axis: Optional[Union[int, Tuple[int]]] = None,
+        initial: Any = 1) -> Array:
```
kaushikcfd, line 317: Should this be `_NoValue` as well? Numpy does it, see https://numpy.org/doc/stable/reference/generated/numpy.prod.html#numpy.prod.

pytato/reductions.py
```diff
@@ -151,76 +262,106 @@ def _make_reduction_lambda(op: str, a: Array,
...
+        if the reduction is empty.
+        In that case, the reduction size must not be symbolic.
     """
-    return _make_reduction_lambda("product", a, axis)
+    return _make_reduction_lambda(ProductReductionOperation(), a, axis, initial)
 
 
-def all(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+def all(a: Array, axis: Optional[Union[int, Tuple[int]]] = None,
+        initial: Any = True) -> Array:
```
kaushikcfd, line 334: `numpy.all` doesn't take any initial, see https://numpy.org/doc/stable/reference/generated/numpy.all.html?highlight=all#numpy.all

pytato/reductions.py
```diff
@@ -151,76 +262,106 @@ def _make_reduction_lambda(op: str, a: Array,
...
+        if the reduction is empty.
+        In that case, the reduction size must not be symbolic.
     """
-    return _make_reduction_lambda("all", a, axis)
+    return _make_reduction_lambda(AllReductionOperation(), a, axis, initial)
 
 
-def any(a: Array, axis: Optional[Union[int, Tuple[int]]] = None) -> Array:
+def any(a: Array, axis: Optional[Union[int, Tuple[int]]] = None,
+        initial: Any = False) -> Array:
```
kaushikcfd, line 351: `numpy.any` doesn't take any initial argument.

pytato/reductions.py
```diff
@@ -151,76 +262,106 @@ def _make_reduction_lambda(op: str, a: Array,
...
     """
     Returns the logical-or of array *a*'s elements along the *axis* axes.
 
     :arg a: The :class:`pytato.Array` on which to perform the reduction.
 
     :arg axis: The axes along which the elements are to be product-reduced.
         Defaults to all axes of the input array.
+    :arg initial: The value returned for an empty array, if supplied.
+        If not supplied, an :exc:`ValueError` will be raised
+        if the reduction is empty.
```
kaushikcfd, line 361: Not quite true:
```python
>>> import numpy as np
>>> np.all([])
True
```

kaushikcfd (approved):
This almost looks great! Found some minor numpy compat issues, feel free to merge after that.

Thanks!

pytato/reductions.py
inducer: I think this is OK. Numpy will return 1 for an empty `prod`. I think for an immutable object like `1` it does not matter whether that's done via `_NoValue` and then setting internally, or directly like this.

pytato/reductions.py
inducer: Addressed in latest push: https://github.com/inducer/pytato/compare/01f5f0fe18220bd21cb9f239dafa9fa7877fe1c2..8c022221397a03ec19f74d8418e3d5a690584b5b

pytato/reductions.py
inducer: Addressed in latest push: https://github.com/inducer/pytato/compare/01f5f0fe18220bd21cb9f239dafa9fa7877fe1c2..8c022221397a03ec19f74d8418e3d5a690584b5b

pytato/reductions.py
inducer: Addressed in latest push: https://github.com/inducer/pytato/compare/01f5f0fe18220bd21cb9f239dafa9fa7877fe1c2..8c022221397a03ec19f74d8418e3d5a690584b5b

pytato/reductions.py
```diff
@@ -137,9 +228,29 @@ def _make_reduction_lambda(op: str, a: Array,
...
+        for iax in reduction_axes:
+            shape_iax = a.shape[iax]
+            if get_dependencies(shape_iax):
+                raise NotImplementedError(
+                        "cannot statically determine emptiness of "
+                        f"reduction axis {iax} (0-based)")
+
+            assert isinstance(shape_iax, int)
+
+            if shape_iax == 0:
```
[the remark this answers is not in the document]
inducer: Addressed in latest push: https://github.com/inducer/pytato/compare/01f5f0fe18220bd21cb9f239dafa9fa7877fe1c2..8c022221397a03ec19f74d8418e3d5a690584b5b

pytato/reductions.py
```diff
@@ -137,9 +228,29 @@ def _make_reduction_lambda(op: str, a: Array,
...
+    new_shape, reduction_axes = _normalize_reduction_axes(a.shape, axis)
     del axis
-    indices, redn_bounds = _get_reduction_indices_bounds(a.shape, axes)
+    indices, redn_bounds = _get_reduction_indices_bounds(a.shape, reduction_axes)
+
+    if initial is _NoValue:
+        from pytato.scalar_expr import get_dependencies
+        for iax in reduction_axes:
+            shape_iax = a.shape[iax]
+            if get_dependencies(shape_iax):
```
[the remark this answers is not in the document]
inducer: Addressed in latest push: https://github.com/inducer/pytato/compare/01f5f0fe18220bd21cb9f239dafa9fa7877fe1c2..8c022221397a03ec19f74d8418e3d5a690584b5b

pytato/reductions.py
```diff
@@ -151,52 +261,72 @@ def _make_reduction_lambda(op: str, a: Array,
...
     """
     Sums array *a*'s elements along the *axis* axes.
 
     :arg a: The :class:`pytato.Array` on which to perform the reduction.
 
     :arg axis: The axes along which the elements are to be sum-reduced.
         Defaults to all axes of the input array.
+    :arg initial: The value returned for an empty array, if supplied.
+        If not supplied, an :exc:`ValueError` will be raised
+        if the reduction is empty.
```
kaushikcfd, line 275: Sorry had missed this:
```python
>>> np.sum([])
0.0
```

pytato/reductions.py
```diff
@@ -151,52 +261,72 @@ def _make_reduction_lambda(op: str, a: Array,
...
+def sum(a: Array, axis: Optional[Union[int, Tuple[int]]] = None,
+        initial: Any = _NoValue) -> Array:
     """
     Sums array *a*'s elements along the *axis* axes.
 
     :arg a: The :class:`pytato.Array` on which to perform the reduction.
 
     :arg axis: The axes along which the elements are to be sum-reduced.
         Defaults to all axes of the input array.
+    :arg initial: The value returned for an empty array, if supplied.
```
kaushikcfd, line 273: The description misses that this is the initial value of the reduction, i.e. `pt.sum([1], initial=3)` would return 4 isn't clear here. (Same for `amin`, `prod`, `amax`)

kaushikcfd (approved):
Found out some more places where the implementation disagrees Numpy, good to go after that.

pytato/reductions.py
inducer: Good catch! Addressed in this force push: https://github.com/inducer/pytato/compare/acf72ae24d972f2da3243d755970f0ebf32b3113..b3e359c1859600b2d0b735fbf5484e30366e2f08

pytato/reductions.py
inducer: https://github.com/inducer/pytato/compare/acf72ae24d972f2da3243d755970f0ebf32b3113..b3e359c1859600b2d0b735fbf5484e30366e2f08

PR #247, opened 2022-01-18
Cast scalars to sized-dtypes

### LINKED ISSUE #246: Scalars should be promoted to numpy-dtypes so that their size is known to the downstream targets
The following code fails:
```python
x = pt.make_placeholder("x", shape=(10, 4), dtype=np.float64)
pt_prg = pt.generate_loopy(x * 3.14j)
print(lp.generate_code_v2(pt_prg.program).device_code())
```

with the error msg:
```
loopy.diagnostic.TypeInferenceFailure: Complex constant '3.14j' needs to be sized (i.e. as numpy.complex64/128) for type inference 
```

However, the numpy-equivalent interprets the complex literal as `complex128`.

kaushikcfd:
Closes #246.

### DIFF
--- a/pytato/utils.py
+++ b/pytato/utils.py
@@ -171,6 +171,12 @@ def broadcast_binary_op(a1: ArrayOrScalar, a2: ArrayOrScalar,
                         ) -> ArrayOrScalar:
     from pytato.array import _get_default_axes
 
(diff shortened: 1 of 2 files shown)


kaushikcfd:
Seems pretty trivial, setting to auto-merge.

PR #248, opened 2022-01-20
add NodeCountMapper

matthiasdiener:


test/test_pytato.py
```diff
@@ -580,6 +580,25 @@ def test_repr_array_is_deterministic():
...
+    for i in range(10):
+        rdagc = RandomDAGContext(np.random.default_rng(seed=i),
+                                 axis_len=axis_len, use_numpy=False)
+        dag = make_random_dag(rdagc)
+        ncm = NodeCountMapper()
+        ncm(dag)
+
+        counts.append(ncm.count)
+
+    assert counts == [32, 32, 1, 1, 17, 15, 47, 18, 1, 13]
```
kaushikcfd, line 599: This might be slightly annoying in the future to re-tune these values if our random generator changes.
How about `assert ncm(dag) == len(pt.transform.DependencyMapper()(dag))`?
This should be fine as long as the dags are small (less than a couple hundred nodes.)

pytato/transform.py
```diff
@@ -1335,4 +1336,26 @@ def map_loopy_call(self, expr: LoopyCall) -> LoopyCall:
...
+    def __init__(self) -> None:
+        super().__init__()
+        self.count = 0
+
+    def post_visit(self, expr: Any) -> None:
+        self.count += 1
+
+# }}}
+
+
```
kaushikcfd, line 1360: I think `pt.analysis` is a better place for this. Since this is not technically a transformation.

kaushikcfd (approved):
LGTM, thanks!

pytato/transform.py
kaushikcfd: And an interface like `pt.analysis.get_number_of_nodes` would be pretty helpful.

test/test_pytato.py
matthiasdiener: Thanks, done in 89dc13e.

pytato/transform.py
matthiasdiener: done in 89dc13e

test/test_pytato.py
```diff
@@ -580,6 +580,20 @@ def test_repr_array_is_deterministic():
...
+    from pytato.analysis import get_num_nodes
+
+    axis_len = 5
+
+    for i in range(10):
+        rdagc = RandomDAGContext(np.random.default_rng(seed=i),
+                                 axis_len=axis_len, use_numpy=False)
+        dag = make_random_dag(rdagc)
+
+        assert get_num_nodes(dag)-1 == len(pt.transform.DependencyMapper()(dag))
```
kaushikcfd, line 594: Why do we need to subtract by '1' here?

test/test_pytato.py
kaushikcfd: Oh didn't see this. Moving discussion to https://github.com/inducer/pytato/pull/248#discussion_r789225495.

test/test_pytato.py
kaushikcfd: Is it because of the `NamedArray` being included in one and not in the other?

matthiasdiener: > Is it because of the `NamedArray` being included in one and not in the other?

I'm suspecting something like this.

kaushikcfd: Yep, that seems to be the case, `CombineMapper` steps over the `NamedArray`s. https://github.com/inducer/pytato/blob/191857a92f311f22e4b2d43aa6f9682a8c14ceeb/pytato/transform.py#L389-L392. Ok, this is good from my end. Thanks!

matthiasdiener: Great, thanks!

kaushikcfd: Aah slightly incorrect, both the mappers "skip" NamedArray, but WalkMapper counts an extra one for `DictOfNamedArrays` as well.

https://github.com/inducer/pytato/blob/191857a92f311f22e4b2d43aa6f9682a8c14ceeb/pytato/transform.py#L640-L647

matthiasdiener: Thanks, should be fixed now!

PR #249, opened 2022-01-20
distributed: limit size of created partitions

matthiasdiener:
This does not work yet since it causes a PartitionInducedCycleError.

Some statistics for y2-isolator:
- total number of nodes in DAG (unpartitioned): 6728

### DIFF
--- a/pytato/equality.py
+++ b/pytato/equality.py
@@ -231,7 +231,8 @@ class EqualityComparer:
                 )
 
     def map_loopy_call_result(self, expr1: LoopyCallResult, expr2: Any) -> bool:
-        return (self.rec(expr1._container, expr2._container)
+        return (expr1.__class__ is expr2.__class__
+                and self.rec(expr1._container, expr2._container)
                 and expr1.tags == expr2.tags)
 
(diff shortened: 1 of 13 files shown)


pytato/equality.py
```diff
@@ -231,7 +231,8 @@ def map_loopy_call(self, expr1: LoopyCall, expr2: Any) -> bool:
                 )
 
     def map_loopy_call_result(self, expr1: LoopyCallResult, expr2: Any) -> bool:
-        return (self.rec(expr1._container, expr2._container)
+        return (expr1.__class__ is expr2.__class__
+                and self.rec(expr1._container, expr2._container)
```
matthiasdiener, line 235: @kaushikcfd - Does this change look reasonable? If so, I could open a separate PR with just this change (expr2 does not always have a `_container` attribute)

kaushikcfd: Yep, that is very reasonable. Thanks for fixing this! (A different PR would be great.)

Looks like we also missed this condition in `map_named_array`.

matthiasdiener: Done in #250.

matthiasdiener:
Closing as unnecessary

PR #250, opened 2022-01-21
fix equality mapper functions

matthiasdiener:


kaushikcfd (approved):
LGTM, thanks!

pytato/equality.py
```diff
@@ -230,7 +231,8 @@ def map_loopy_call(self, expr1: LoopyCall, expr2: Any) -> bool:
                 )
 
     def map_loopy_call_result(self, expr1: LoopyCallResult, expr2: Any) -> bool:
-        return (self.rec(expr1._container, expr2._container)
+        return (expr1.__class__ is expr2.__class__
+                and self.rec(expr1._container, expr2._container)
                 and expr1.tags == expr2.tags)
```
matthiasdiener, line 236: Should we also compare the other fields here (name, axes)?

kaushikcfd: Yep!

matthiasdiener: Done in f72bb2d

PR #251, opened 2022-01-21
Allow only affine array shapes

kaushikcfd:
Using `pymbolic.distribute` to check equivalence of shape expressions is dubious, to say the least. (Modulos not supported)


kaushikcfd:
This was already reviewed as part of #217. So I'm setting this to auto-merge.

PR #253, opened 2022-01-27
Implements `rec_get_user_nodes`.

### LINKED ISSUE #252: pt.transform.tag_child_nodes is O(2^(|V|))
```python
import pytato as pt
from numpy.random import default_rng


def construct_intestine_graph(depth=100, seed=0):
    rng = default_rng(seed)
    x = pt.make_placeholder("x", shape=(10,), dtype=float)
    y = x

    for _ in range(depth):
        coeff1, coeff2 = rng.integers(0, 10, 2)
        y = coeff1 * y + coeff2 * y

    return y, x


expr, inp = construct_intestine_graph()
user_collector = pt.transform.UsersCollector()
user_collector(expr)
result = pt.transform.tag_child_nodes(user_collector.node_to_users, 1, inp)
print(result)
```
keep going on and on. (Note: won't work on main as the argument order [here](https://github.com/inducer/pytato/blob/4975cfd89d784e82063b196307be587fad862cdc/pytato/transform.py#L1177-L1178) is incorrect, needs to reverse the 2nd and 3rd argument order)

Other than that, I think `tag_child_nodes` is very specific and I don't think this needs to be an abstraction on the graphs we should support.

kaushikcfd:
Resolves #252.

### DIFF
--- a/pytato/transform.py
+++ b/pytato/transform.py
@@ -27,6 +27,7 @@ THE SOFTWARE.
 from abc import ABC, abstractmethod
 from typing import (Any, Callable, Dict, FrozenSet, Union, TypeVar, Set, Generic,
                     List, Mapping, Iterable, Optional, Tuple)
+from pyrsistent.typing import PMap as PMapT
 
 from pytato.array import (
         Array, IndexLambda, Placeholder, Stack, Roll,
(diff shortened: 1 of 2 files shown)


kaushikcfd:
@matthiasdiener: How do you feel about this? One can construct trivially construct `tag_to_nodes` from `rec_get_user_nodes`, please let me know if you disagree.

matthiasdiener (approved):
LGTM, thanks!

inducer:
Thanks for working on this!

PR #257, opened 2022-02-01
Fix IndexBase equality comparison

kaushikcfd:
Provided test fails on `main`.


inducer:
Yikes! Thanks for finding that. :+1:

PR #262, opened 2022-02-21
Use `np.errstate` to raise on numpy errors, instead of a warning filter

inducer:
The previous way broke https://github.com/inducer/pytools/pull/120, which (deliberately) introduced a warning to be raised, which the code being removed turned into an exception.


inducer:
Low-impact and simple, proceeding with merge.

PR #264, opened 2022-02-22
Add `deduplicate_data_wrappers`

inducer:
Supports https://github.com/inducer/grudge/pull/218.

cc @matthiasdiener

### DIFF
--- a/pytato/transform.py
+++ b/pytato/transform.py
@@ -26,9 +26,12 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+import logging
+import numpy as np
 from abc import abstractmethod
 from typing import (Any, Callable, Dict, FrozenSet, Union, TypeVar, Set, Generic,
-                    List, Mapping, Iterable, Tuple, Optional, TYPE_CHECKING)
+                    List, Mapping, Iterable, Tuple, Optional, TYPE_CHECKING,
+                    Hashable)
 
 from pytato.array import (
         Array, IndexLambda, Placeholder, Stack, Roll,
@@ -36,7 +39,7 @@ from pytato.array import (
         AbstractResultWithNamedArrays, Reshape, Concatenate, NamedArray,
         IndexRemappingBase, Einsum, InputArgumentBase,
         BasicIndex, AdvancedIndexInContiguousAxes, AdvancedIndexInNoncontiguousAxes,
-        IndexBase)
+        IndexBase, DataInterface)
 
 from pytato.loopy import LoopyCall, LoopyCallResult
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


pytato/transform.py
```diff
@@ -1474,4 +1481,102 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
...
+                    ary.shape,
+                    ary.strides,
+                    ary.dtype,
+                    )
+    if isinstance(ary, np.ndarray):
+        return (
+                ary.shape,
+                ary.strides,
+                ary.dtype,
+                hash(ary.data.tobytes())
```
kaushikcfd, line 1504: Maybe we can avoid this? These quite likely aren't the data that the user expected to be equal but just happened to be equal.

pytato/transform.py
```diff
@@ -1474,4 +1481,102 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
...
+
+def _check_array_equality(ary1: DataInterface, ary2: DataInterface) -> bool:
+    assert type(ary1) is type(ary2)
+
+    import sys
+    if "pyopencl" in sys.modules:
+        from pyopencl.array import Array as CLArray
+        if isinstance(ary1, CLArray):
+            # FIXME: We'll accept hash equality as equality.
+            return True
```
kaushikcfd, line 1518: The return type of this routine is `bool`, which doesn't quite match in the else branch.

pytato/transform.py
```diff
@@ -1474,4 +1481,102 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
...
+
+    if isinstance(ary1, np.ndarray):
+        assert isinstance(ary2, np.ndarray)
+        return np.array_equal(ary1, ary2)
```
kaushikcfd, line 1522: Same concern as above. I.e. can we avoid this equality check?

kaushikcfd (approved):
Thanks, this is a frequent problem, and shipping it as a pytato transform makes a lot of sense. 
Left some minor implementation nits, otherwise LGTM!

pytato/transform.py
matthiasdiener: But wouldn't it be good to avoid the data duplication even if the equality is just conincidental?

pytato/transform.py
matthiasdiener: Sorry, which else branch are you referring to? The one in line 1524?

kaushikcfd: Aah I see, my bad, it's caught in the `isisntance(ary1, np.ndarry)`. Resolving this.

pytato/transform.py
kaushikcfd: I thought it was a bit expensive to traverse all the elements in the array.

pytato/transform.py
inducer: https://github.com/inducer/pytato/compare/42cf0e41b689dd61a7bb7269273652be982dc8f1..990d1739f5213bf5be48fbe7418806ec568900d3

pytato/transform.py
inducer: I thought I had convinced myself that a scenario existed where concurrent allocation in a separate thread could drive this to be incorrect. But I no longer buy that scenario.

https://github.com/inducer/pytato/compare/42cf0e41b689dd61a7bb7269273652be982dc8f1..990d1739f5213bf5be48fbe7418806ec568900d3

inducer:
Thanks for taking a look!

PR #267, opened 2022-02-28
Override Array._with_new_tags for new Taggable interface

inducer:
- [x] Needs https://github.com/inducer/pytools/pull/120, it turns out


inducer:
Turns out https://github.com/inducer/pytools/pull/120 broke CI (removed `Taggable.copy`). This is fixed in 891e29b8cf6c2dacd9fed60b83bc0cf6b10c8c04, and this overall change is simple enough that I think I'll go ahead with it.

PR #268, opened 2022-03-03
Add a test application: Trace an FFT

inducer:
cc @isuruf 

This also fixes an actual bug: You could tag multiple distinct values as `Named` the same thing, and nothing would complain.

Read commit-by-commit.


inducer:
> Merge branch 'main' into trace-fft

Please don't, on PRs that I created. Not all PRs are intended to be squashed.

inducer:
I feel like there's ultimately very little here:

- A fix for something that's clearly a bug (allowing clashing `Named` tags), plus a regression test
- An additional "application-y" test

Given the simplicity, I'll go ahead and merge this.

PR #282, opened 2022-03-22
Add comments that upper OOB access is UB in codegen

inducer:


kaushikcfd (approved):
Yep, LGTM. Thank you!

PR #284, opened 2022-03-23
Add tag to store array creation traceback

matthiasdiener:
_Please squash_

#### TODOs as of 11/14/2023:

- [x] Make sure we do not create multiple CreatedAt tags (or overwrite existing ones) for the same node ([4ec3cbf](https://github.com/inducer/pytato/pull/284/commits/4ec3cbf60dffa9a38e1518269db31dc9c7daa6fc))
- [x] Make sure tags propagate correctly through mappers ([4ec3cbf](https://github.com/inducer/pytato/pull/284/commits/4ec3cbf60dffa9a38e1518269db31dc9c7daa6fc))
- [x] More comprehensive performance testing (see below)

---

#### Simple performance test as of 11/14/2023:

- M1

```python
import pytato as pt
from time import time
import numpy as np

n=50

start = time()
for i in range(100000):
    pt.make_placeholder(name='a', shape=(n, n), dtype=np.float64)
stop = time()
print(stop-start)
```

| Config         | Time |
|----------------------|-------|
| w/ `CreatedAt` | 1.48s |
| w/o `CreatedAt`         | 0.35s |

---

#### `smoke_test_ks_3d` performance as of 11/15/2023:

- M1, 1 rank, CPU, production, `MPIFusionContractorArrayContext`, `LOOPY_NO_CACHE=1`

| Config          | Total time |
|----------------------|-------|
| w/ `CreatedAt` | 463s |
| w/o `CreatedAt`         | 452s |

---

Needs:
- [x] https://github.com/inducer/meshmode/pull/329
- [x] https://github.com/inducer/pytools/pull/134


matthiasdiener:
This is ready for a first look @inducer @kaushikcfd

inducer:
- Could you look at the CI failures?
- Could you try to quantify the runtime impact on complex DAGs and add that to the PR description?

pytato/array.py
```diff
@@ -419,8 +419,12 @@ class Array(Taggable):
     __array_priority__ = 1  # disallow numpy arithmetic to take precedence
 
     def __init__(self, axes: AxesT, tags: FrozenSet[Tag]) -> None:
+        import traceback
+        v = "".join(traceback.format_stack())
+        from pytato.tags import CreatedAt
+        c = CreatedAt(v)
```
kaushikcfd, line 425: 1. I think putting it in `Array.__init__` would lead to several of these after a  sequence of `.map_and_copy`'s.
3. Having many of these would also affect our DAG visualization.
4. We could probably only attach the first call-frame from outside pytato, but to know these we might have to start attaching them only in the user-exposed function.

pytato/array.py
```diff
@@ -419,8 +419,12 @@ class Array(Taggable):
     __array_priority__ = 1  # disallow numpy arithmetic to take precedence
 
     def __init__(self, axes: AxesT, tags: FrozenSet[Tag]) -> None:
+        import traceback
+        v = "".join(traceback.format_stack())
```
inducer, line 423: Don't splat this out to a string. I'd like to keep this machine-readable. (But don't use the full traceback object either... that contains all the call frames, which we emphatically don't want to keep alive.)

kaushikcfd: Probably something like [clang's SourceLocation](https://clang.llvm.org/doxygen/classclang_1_1SourceLocation.html) is what we want here?

pytato/array.py
inducer: Good points.

I think an important tweak is to only add one if one isn't already there (in `tags`). We're not interested in call stacks of where these get rewritten by mappers (I think).

> We could probably only attach the first call-frame from outside pytato

We could, though I'd expect that that might be a bunch of engineering that, ultimately, might not be buying us much.

kaushikcfd: > I think an important tweak is to only add one if one isn't already there (in tags).

I would be cautious in spending time in the constructor.

> We could, though I'd expect that that might be a bunch of engineering that, ultimately, might not be buying us much.

I think at least dropping the frames within pytato would be helpful. Otherwise, they would all be polluted with `pytato.array.Array.__add__`, `pytato.cmath.sin`.

pytato/array.py
inducer: I think we can start by storing what's returned from `extract_tb`. 

https://docs.python.org/3/library/traceback.html#traceback.extract_tb

inducer: - https://github.com/python/cpython/blob/ff619c7dfe8dcb0e4c8dc655abc3acc7dc586d0d/Lib/traceback.py#L372
- https://github.com/python/cpython/blob/ff619c7dfe8dcb0e4c8dc655abc3acc7dc586d0d/Lib/traceback.py#L247

pytato/array.py
inducer: > I would be cautious in spending time in the constructor.

Argh, that's a very fair point. We shouldn't. The alternative is to splat this into a zillion other places, but I agree that's preferable.

> I think at least dropping the frames within pytato would be helpful. Otherwise, they would all be polluted with `pytato.array.Array.__add__`, `pytato.cmath.sin`.

Fair, but we can also do that when rendering, which I think will be *way* less frequent than construction.

kaushikcfd: > The alternative is to splat this into a zillion other places

https://github.com/inducer/pytato/pull/285

> Fair, but we can also do that when rendering, which I think will be way less frequent than construction.

Agreed, I'm completely fine with this.

pytato/array.py
matthiasdiener: > I think we can start by storing what's returned from `extract_tb`.
> 
> https://docs.python.org/3/library/traceback.html#traceback.extract_tb

`extract_tb` returns a `StackSummary` object which isn't hashable, and I'm not sure what to do about it.

kaushikcfd: One option: we could have our version of hashable `StackSummary` and `FrameSummary`.

kaushikcfd: > https://docs.python.org/3/library/traceback.html#traceback.StackSummary.from_list

Each FrameSummary is essentially a 4-tuple  (filename, lineno, name, line). A stack_summary is a tuple of FrameSummary i.e. all hashable.

matthiasdiener: Can we just convert the StackSummary into a tuple (and then use `StackSummary.from_list` if we need the StackSummary)?

kaushikcfd: Yep, I guess that works too. (Note that `FrameSummary` is still unhashable)

matthiasdiener: > Can we just convert the StackSummary into a tuple (and then use `StackSummary.from_list` if we need the StackSummary)?

[2037423](https://github.com/inducer/pytato/pull/284/commits/20374238db313c5173f4c5f6d30a31da07a8944a) implements this idea (and admittedly does not win prizes for elegance).

pytato/tags.py
```diff
@@ -101,3 +103,16 @@ class AssumeNonNegative(Tag):
     :class:`~pytato.target.Target` that all entries of the tagged array are
     non-negative.
     """
+
+
+# See
+# https://mypy.readthedocs.io/en/stable/additional_features.html#caveats-known-issues
+# on why this can not be '@tag_dataclass'.
+@dataclass(init=True, eq=True, frozen=True, repr=True)
+class CreatedAt(UniqueTag):
```
inducer, line 112: I wonder whether `CreatedAt` should impact equality of the tagged object. I.e. if you have, say, two `IndexLambda`s that only differ in their `CreatedAt` tags, should they be considered different?

- Upside: There's clearly no need to generate code once for every version if the only difference is a `CreatedAt`. (But maybe that's unlikely?)
- Downside: Potentially confusing when looking at `CreatedAt` metadata.

kaushikcfd: Yep, I think it is unlikely that the nodes would "diverge". So I would vote against ignoring these during codegen. On the upside, if we preserve the information we'll also get source code information on the loopy end.

inducer: OK, sounds good. Let's keep including them in equality checking until there's a problem that results from that.

matthiasdiener: I believe most of the pytest failures currently are due to the equality checking.

pytato/array.py
```diff
@@ -1675,8 +1676,36 @@ def _get_default_axes(ndim: int) -> AxesT:
...
 
+@dataclass(frozen=True, eq=True)
+class _PytatoFrameSummary:
+    filename: str
+    lineno: int
+    name: str
+    line: str
+
+
+class _PytatoStackSummary(Tag):
```
inducer, line 1687: This needn't be a `Tag`.

pytato/array.py
```diff
@@ -1675,8 +1676,36 @@ def _get_default_axes(ndim: int) -> AxesT:
...
+    def __init__(self, stack_summary: StackSummary) -> None:
+        self.frames: List[_PytatoFrameSummary] = []
+        for s in stack_summary:
+            pfs = _PytatoFrameSummary(s.filename, s.lineno, s.name, s.line)
+            self.frames.append(pfs)
```
inducer, line 1692: List comprehension

pytato/array.py
```diff
@@ -1675,8 +1676,36 @@ def _get_default_axes(ndim: int) -> AxesT:
...
+        return StackSummary.from_list(frames)  # type: ignore[arg-type]
+
+
 def _get_default_tags() -> TagsType:
-    return frozenset()
+    import traceback
+    from pytato.tags import CreatedAt
+
+    c = CreatedAt(_PytatoStackSummary(traceback.extract_stack()))
+    return frozenset((c,))
```
inducer, line 1708:
```suggestion
    if __debug__:
        c = CreatedAt(_PytatoStackSummary(traceback.extract_stack()))
        return frozenset((c,))
```

test/test_pytato.py
```diff
@@ -790,6 +790,28 @@ def test_einsum_dot_axes_has_correct_dim():
...
+
+    res = a+b
+
+    from pytato.tags import CreatedAt
+
+    found = False
+
+    # Make sure the function name appears in the traceback
+    for tag in res.tags:
+        if isinstance(tag, CreatedAt):
```
inducer, line 805:
```
created_tag, = res.tags_of_type(CreatedAt)
```

pytato/array.py
kaushikcfd, line 1692: Move this out of `__init__` to a classmethod `_PytatoStackSummary.from_stack_summary`?

test/test_pytato.py
```diff
@@ -462,7 +464,12 @@ def test_array_dot_repr():
     x = pt.make_placeholder("x", (10, 4), np.int64)
     y = pt.make_placeholder("y", (10, 4), np.int64)
 
+    from pytato.transform import remove_tags_of_type
+    from pytato.tags import CreatedAt
+
     def _assert_stripped_repr(ary: pt.Array, expected_repr: str):
+        ary = cast(pt.Array, remove_tags_of_type(CreatedAt, ary))
```
kaushikcfd, line 471: I wonder if we need to strip these tags in the stringifier itself? Either that or we might need to define `CreatedAt.__repr__` to avoid seeing multiple lines for a single tag.

pytato/array.py
matthiasdiener: Done in [02362bf](https://github.com/inducer/pytato/pull/284/commits/02362bfaf8f6d6da00e2b2e639b9b6773866aa14)

pytato/array.py
matthiasdiener: Fixed in [02362bf](https://github.com/inducer/pytato/pull/284/commits/02362bfaf8f6d6da00e2b2e639b9b6773866aa14)

test/test_pytato.py
matthiasdiener: Fixed in [1c275fd](https://github.com/inducer/pytato/pull/284/commits/1c275fd80cca82a6f584f7b20b33549c3ff977be)

pytato/array.py
matthiasdiener: `_PytatoStackSummary` was changed to a frozen dataclass, so I think this can't be done.

test/test_pytato.py
matthiasdiener: Yeah, the CreatedAt tags take a lot of space in the visualization now...

matthiasdiener:
Not that in the current implementation, the following code will not compare the two placeholders as equal (as they are created on two different lines), without removing the CreatedAt tags or modifying the EqualityComparer. 

```
a = pt.make_placeholder(name="a", shape=(10, 10), dtype=np.float64)
b = pt.make_placeholder(name="a", shape=(10, 10), dtype=np.float64)

print(a==b)
```

Furthermore, the result of this comparison changes depending on whether python is running with optimizations or not. This seems not ideal. My initial thought was to add this traceback info not as a normal tag, but as an extra field in the `Array` class, which I think would avoid these issues. What do you think @inducer @kaushikcfd ?

kaushikcfd:
> Note that in the current implementation, the following code will not compare the two placeholders as equal (as they are created on two different lines), without removing the CreatedAt tags or modifying the EqualityComparer.
>
> a = pt.make_placeholder(name="a", shape=(10, 10), dtype=np.float64)
> b = pt.make_placeholder(name="a", shape=(10, 10), dtype=np.float64)
>
> print(a==b)

I think these placeholders not being equal is OK. We already require
that within a DAG we cannot have more than placeholder objects having
the same name. But this also hints at a "new developer guideline" that
optimization passes should never invoke creation routines that are
directed towards the users as they would force divergence.

>
> Furthermore, the result of this comparison changes depending on whether python is running with optimizations or not. This seems not ideal. My initial thought was to add this traceback info not as a normal tag, but as an extra field in the Array class, which I think would avoid these issues.

Yep, this is annoying. I vote we keep these tags in `-O` mode as well.
This would cut down our debug-times.

inducer:
> I think these placeholders not being equal is OK.

One "suboptimal" consequence of this is that our "common subexpression elimination" (really just "canonicalization of nodes via hash") is defeated by this, and we'll keep many more versions of the same computation than is necessary. If we stick with non-euqality of these nodes, I think we'll want an explicit "canonicalization" pass. As a result, `CreatedAt` actually *shouldn't* be a `UniqueTag`, because a node can be "created at" more than one backtrace.

>  I vote we keep these tags in `-O` mode as well.

I'd like to know more about the cost of these changes in practice before I'm OK with them being included unconditionally.

> But this also hints at a "new developer guideline" that optimization passes should never invoke creation routines that are directed towards the users as they would force divergence.

Agreed. @matthiasdiener, could you add this?

kaushikcfd:
I guess we should also acknowledge that there are many places in the CI where we use `==` to mean structural equality, i.e. we'll never get `2*a + 3*b == 2*a +3*b`. That makes me uncomfortable, but I don't think this will be of much consequence in a large-scale application.

matthiasdiener:
> I think these placeholders not being equal is OK. 

Just to clarify, it is not just the Placeholders that have this problem, but at least all of the `make_` operations. I just chose Placeholders to illustrate this issue.

Here is how it looks like currently:
image

---

I'm struggling to resolve the pytest errors that use `construct_intestine_graph`, removing the CreatedAt tags crashes with a `RecursionError: maximum recursion depth exceeded`.  See [437954c](https://github.com/inducer/pytato/pull/284/commits/437954cbb203793890c16616bbcabab0757f04a2) for an illustration of this issue.

It also leads to awkward code like:

```python
    node_to_users = {}

    for k in user_collector.node_to_users.keys():
        new_key = remove_tags_of_type(CreatedAt, k)
        new_values = set({remove_tags_of_type(CreatedAt, v) for v in user_collector.node_to_users[k]})

        node_to_users[new_key] = new_values
```

kaushikcfd:
> Here is how it looks like currently:

Thanks, this already looks pretty good. However, The stringification of the tag could just limit to `<file:function:lineno>` where these properties correspond to the topmost call-frame outside of pytato.

matthiasdiener:
> > Here is how it looks like currently:
> 
> I don't think the `CreatedAt` tags are useful to have in the visualized DAG. Please take them out. Relatedly, `dot` supports a `tooltip` property that we could use to show additional, more detailed information on mouse-hover.

I implemented Kaushik's suggestion of just showing the first non-pytato file in the traceback in [f05592e](https://github.com/inducer/pytato/pull/284/commits/f05592e87cea001c195714fe4a7c59deac0fb28f). Here is how it looks like:

image

Imho, this could be quite useful. What do you think?

matthiasdiener:
> I think that's unusable. Utility of these graphs depends on fitting many nodes (with readable details) onto your screen. If a node is that big, the resulting graph visualization is not useful.

Ok, final try ;-) What do you think of this way ([d040996](https://github.com/inducer/pytato/pull/284/commits/d0409968304c553f7c62a28463f74881e4e8a4c3)) to visualize (with a separate created_at field, and limiting the width of the information)?

image

inducer:
> Some initial timing measurements:

One reason I asked you to do a realistic measurement in mirgecom is because not all stacktraces their will be as shallow as in the example you show here. So I wonder whether your measurement is meaningful.

inducer:
> Ok, final try ;-)

That node is still unreasonably big IMO.

matthiasdiener:
> That node is still unreasonably big IMO. What of that `tooltip` suggestion?

d80066f4 implements the tooltip suggestion. Here is how it looks like:

image

test/test_pytato.py
matthiasdiener: I removed them from the visualization and put them into a tooltip

inducer:
I would love to merge this. But in order to do so, I would like to know how much of a cost increase there is from this when using it in mirgecom, perhaps for a range of operators, including one that exercises chemistry.

matthiasdiener:
> I would love to merge this. But in order to do so, I would like to know how much of a cost increase there is from this when using it in mirgecom, perhaps for a range of operators, including one that exercises chemistry.

I will continue to work on this. However, this PR needs a bit more thinking as the behavior differs between when using this PR and not using it. It might just be a matter of modifying the EqualityMapper to ignore `CreatedAt` tags.

inducer:
> I will continue to work on this.

Sounds good. Would it be fair then to bump this back to draft? (This helps me understand where review attention is needed.)

> It might just be a matter of modifying the EqualityMapper to ignore `CreatedAt` tags.

Good point, the impact on this on equality is an important, if irritating, detail. I wonder...

- whether we should exclude certain types of tags from equality comparison, or whether we should have a different tag set that's not considered for equality?
- whether we should seek to collect these tags from all nodes that compare "equal", perhaps with a special mapper?

kaushikcfd:
I think getting rid of the `CreatedAt` tags during equality makes sense. Here's a patch -- https://gist.github.com/kaushikcfd/ccfe08185f1b9eabee0534f790544c5f.

inducer:
I find it kind of unappealing to specifically exclude `CreatedAt`. This is a generic problem, and it should be solved generically at the level of `pytools.tag`.

matthiasdiener:
[0794bdb](https://github.com/inducer/pytato/pull/284/commits/0794bdb4fd986549bce4ed6f463e31f34f995dec) combines @kaushikcfd's proposed code (thanks!) with a new Tag type to filter out the "superfluous" tags. Is this what you had in mind @inducer ?

matthiasdiener:
> I find it kind of unappealing to specifically exclude `CreatedAt`. This is a generic problem, and it should be solved generically at the level of `pytools.tag`.

What o you think of https://github.com/inducer/pytools/pull/134 ?

matthiasdiener:
I think this is ready for another look @inducer @kaushikcfd

matthiasdiener:
Performance data for autoignition-mpi (7 species) on my Mac M1, single rank (total startup wall time):

| Config             | Time   |
|--------------------|--------|
| CreatedAt disabled | 81.8 s |
| CreatedAt enabled  | 82.3 s |

I also verified that due to the EqualityMapper changes, the performance anomaly I had mentioned to @inducer previously is avoided.

pytato/array.py
```diff
@@ -1589,7 +1596,9 @@ def __init__(self,
         self._shape = shape
 
     def __hash__(self) -> int:
```
inducer, line 1598: - This should now rely on the infrastructure in pytools.

inducer, line 1598: Make sure to add a version constraint on pytools in `setup.py`.

inducer:

pytato/array.py
matthiasdiener: I don't understand what you mean, could you please clarify? We can't `return super().__hash__()` here due to `self.data`.

matthiasdiener: Added in [26c3590](https://github.com/inducer/pytato/pull/284/commits/26c3590e07313788ad2a1b856c5f7e1474ea07db)

pytato/array.py
```diff
@@ -1589,7 +1596,9 @@ def __init__(self,
...
-        return id(self)
+        from pytato.equality import preprocess_tags_for_equality
+        return hash((self.name, id(self.data), self._shape, self.axes,
+                     preprocess_tags_for_equality(self.tags)))
```
inducer, line 1601:
```suggestion
        return hash((self.name, id(self.data), self._shape, self.axes,
                     super().__hash__()))
```

pytato/array.py
inducer: See below on hash of hash.

pytato/equality.py
```diff
@@ -103,14 +119,14 @@ def map_placeholder(self, expr1: Placeholder, expr2: Any) -> bool:
                 and expr1.name == expr2.name
                 and expr1.shape == expr2.shape
                 and expr1.dtype == expr2.dtype
-                and expr1.tags == expr2.tags
+                and self.are_tags_equal(expr1.tags, expr2.tags)
```
inducer, line 122:
```suggestion
                and Taggable.__eq__(expr1, expr2)
```
?

pytato/array.py
matthiasdiener: Slightly rewritten in 87606ed.

pytato/equality.py
matthiasdiener: Done in 87606ed

pytato/array.py
```diff
@@ -479,9 +483,12 @@ def T(self) -> Array:
 
     @memoize_method
     def __hash__(self) -> int:
+        from pytato.equality import preprocess_tags_for_equality
         attrs = []
         for field in self._fields:
             attr = getattr(self, field)
+            if field == "tags":
+                attr = preprocess_tags_for_equality(attr)
```
inducer, line 491: This should use `Taggable.__hash__` instead.

pytato/array.py
```diff
@@ -141,6 +142,9 @@
 .. autoclass:: ReductionAxis
 .. autoclass:: NormalizedSlice
 
+.. autoclass:: _PytatoFrameSummary
+.. autoclass:: _PytatoStackSummary
```
inducer, line 146: I don't want to expose much about the internal structure of these things. Maybe change these to `.. class` and state that they should be considered undocumented?

pytato/equality.py
```diff
@@ -24,26 +24,38 @@
...
 __doc__ = """
+.. autofunction:: preprocess_tags_for_equality
 .. autoclass:: EqualityComparer
 """
 
 
 ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
 
 
+def preprocess_tags_for_equality(tags: FrozenSet[Tag]) -> FrozenSet[Tag]:
```
inducer, line 51: It's up to `Taggable` to decide equality of taggables. This shouldn't be here. I see you're using this in the tests. It's OK to have this there, but only there.

pytato/transform/__init__.py
```diff
@@ -1242,6 +1243,21 @@ def map_and_copy(expr: ArrayOrNames,
     return CachedMapAndCopyMapper(map_fn)(expr)
 
 
+def remove_tags_of_type(tag_types: Union[type, Tuple[type]], expr: ArrayOrNames
```
inducer, line 1246: What's the use case for this?

pytato/array.py
```diff
@@ -1677,8 +1685,94 @@ def _get_default_axes(ndim: int) -> AxesT:
...
+
+        # Fallback in case we don't find any file that is not in the pytato/
+        # directory (should be unlikely).
+        return self.__repr__()
+
+    def __repr__(self) -> str:
+        return "\n  " + "\n  ".join([str(f) for f in self.frames])
+
+
+def _get_default_tags(existing_tags: Optional[FrozenSet[Tag]] = None) \
```
inducer, line 1762: I don't love that this takes `existing_tags`. Why is that necessary?

inducer (commented):
Thanks! Some comments and questions below.

pytato/array.py
matthiasdiener: What do you think of 6c653bf?

pytato/array.py
matthiasdiener: Done in 02dd6f5

pytato/transform/__init__.py
matthiasdiener: We could probably replace this with `preprocess_tags_for_equality` if you prefer.

pytato/array.py
matthiasdiener: In the current implementation, we must manually avoid adding multiple `CreatedAt` tags.

pytato/array.py
```diff
@@ -506,6 +515,18 @@ def T(self) -> Array:
...
+    def __hash__(self) -> int:
+        attrs_filtered: List[Any] = []
+        for field in attrs.fields(type(self)):  # type: ignore[misc]
+            attr = getattr(self, field)
+            if field == "tags":
+                attr = Taggable.__hash__(self)
+            if isinstance(attr, dict):
+                attr = frozenset(attr.items())
+            attrs_filtered.append(attr)
+        return hash(tuple(attrs_filtered))
```
inducer, line 528: I'm not loving this, it will be slower in a major way than the existing `__hash__` made by `attrs`. Maybe we've taken a wrong turn with the filtering in `Taggable`. Possibly the non-equality tags want to be in a separate attribute? (attrs can be told to exclude fields from the generated `__eq__`, `__hash__`, but deeper customization is hard.)

pytato/array.py
```diff
@@ -506,6 +515,18 @@ def T(self) -> Array:
                                tags=_get_default_tags(),
                                axes=_get_default_axes(self.ndim))
 
+    @memoize_method
+    def __hash__(self) -> int:
+        attrs_filtered: List[Any] = []
+        for field in attrs.fields(type(self)):  # type: ignore[misc]
```
inducer, line 521: mypy 1.7 is due out in a few days with a fix for this, let's wait for that.

pytato/array.py
```diff
@@ -660,6 +681,12 @@ def __repr__(self) -> str:
         from pytato.stringifier import Reprifier
         return Reprifier()(self)
 
+    def tagged(self, tags: ToTagSetConvertible) -> Array:
+        from pytato.equality import preprocess_tags_for_equality
+        from pytools.tag import normalize_tags
+        new_tags = preprocess_tags_for_equality(normalize_tags(tags))
```
inducer, line 687: Here's a use of `preprocess_tags_for_equality` outside the tests.

x-ref: https://github.com/inducer/pytato/pull/284/files#r901804283

pytato/stringifier.py
```diff
@@ -95,7 +95,9 @@ def _map_generic_array(self, expr: Array, depth: int) -> str:
         if depth > self.truncation_depth:
             return self.truncation_string
 
-        fields = tuple(field.name for field in attrs.fields(type(expr)))
+        # type-ignore-reason: https://github.com/python/mypy/issues/16254
```
inducer, line 98: mypy 1.7 is due out in a few days with a fix for this, let's wait for that.

pytato/stringifier.py
```diff
@@ -153,7 +155,8 @@ def _get_field_val(field: str) -> str:
 
         return (f"{type(expr).__name__}("
                 + ", ".join(f"{field.name}={_get_field_val(field.name)}"
-                            for field in attrs.fields(type(expr)))
+                    # type-ignore-reason: https://github.com/python/mypy/issues/16254
```
inducer, line 158: mypy 1.7 is due out in a few days with a fix for this, let's wait for that.

matthiasdiener: Removed in [3ebfcfd](https://github.com/inducer/pytato/pull/284/commits/3ebfcfd30f9a75e3d046e637712e53e835ea2719)

pytato/stringifier.py
matthiasdiener: Removed in [3ebfcfd](https://github.com/inducer/pytato/pull/284/commits/3ebfcfd30f9a75e3d046e637712e53e835ea2719)

pytato/array.py
matthiasdiener: I removed the function in https://github.com/inducer/pytato/pull/284/commits/eb1c052539e90412d5fad022ee38895309e0ea08

pytato/array.py
matthiasdiener: Removed in [3ebfcfd](https://github.com/inducer/pytato/pull/284/commits/3ebfcfd30f9a75e3d046e637712e53e835ea2719)

pytato/array.py
matthiasdiener: I think that is a great suggestion, I implemented this in [eb1c052](https://github.com/inducer/pytato/pull/284/commits/eb1c052539e90412d5fad022ee38895309e0ea08). What do you think?

pytato/array.py
matthiasdiener: This is unnecessary in https://github.com/inducer/pytato/pull/284/commits/eb1c052539e90412d5fad022ee38895309e0ea08

pytato/transform/__init__.py
matthiasdiener: I removed the function in https://github.com/inducer/pytato/pull/284/commits/eb1c052539e90412d5fad022ee38895309e0ea08

pytato/equality.py
matthiasdiener: I removed the function in https://github.com/inducer/pytato/pull/284/commits/eb1c052539e90412d5fad022ee38895309e0ea08

matthiasdiener:
This is ready for another review @inducer.

pytato/array.py
```diff
@@ -1752,6 +1781,112 @@ def _get_default_axes(ndim: int) -> AxesT:
...
 
+@attrs.define(frozen=True, eq=True)
+class _PytatoFrameSummary:
+    """Class to store a single call frame."""
+    filename: str
+    lineno: Optional[int]
+    name: str
+    line: Optional[str]
+
+    def update_persistent_hash(self, key_hash: int, key_builder: Any) -> None:
```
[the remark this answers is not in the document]
matthiasdiener: Removed in 0be08a5 (this code was done before support for `attrs` was added to the `pytools.KeyBuilder`).

pytato/array.py
```diff
@@ -1752,6 +1781,112 @@ def _get_default_axes(ndim: int) -> AxesT:
     return tuple(Axis(frozenset()) for _ in range(ndim))
 
 
+@attrs.define(frozen=True, eq=True)
+class _PytatoFrameSummary:
+    """Class to store a single call frame."""
```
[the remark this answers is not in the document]
matthiasdiener: Added in [b1464e6](https://github.com/inducer/pytato/pull/284/commits/b1464e611b095f5611a78da3158a97facb9187f1)

pytato/array.py
```diff
@@ -1752,6 +1781,112 @@ def _get_default_axes(ndim: int) -> AxesT:
...
+        s2 = s2[:maxlen] + " [...]" if len(s2) > maxlen else s2
+        return s1 + "\n" + s2
+
+    def __repr__(self) -> str:
+        return f"{self.filename}:{self.lineno}, in {self.name}(): {self.line}"
+
+
+@attrs.define(frozen=True, eq=True)
+class _PytatoStackSummary:
+    """Class to store a list of :class:`_PytatoFrameSummary` call frames."""
```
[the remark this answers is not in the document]
matthiasdiener: Added in [b1464e6](https://github.com/inducer/pytato/pull/284/commits/b1464e611b095f5611a78da3158a97facb9187f1)

pytato/tags.py
```diff
@@ -108,6 +110,22 @@ class AssumeNonNegative(Tag):
...
+# https://mypy.readthedocs.io/en/stable/additional_features.html#caveats-known-issues
+# on why this can not be '@tag_dataclass'.
+@dataclass(init=True, eq=True, frozen=True, repr=True)
+class CreatedAt(UniqueTag, IgnoredForEqualityTag):
+    """
+    A tag attached to a :class:`~pytato.Array` to store the traceback
+    of where it was created.
+    """
+
+    traceback: _PytatoStackSummary
```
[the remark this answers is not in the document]
matthiasdiener: Sure, that makes sense. Done in [e2672f7](https://github.com/inducer/pytato/pull/284/commits/e2672f707ff14c4e4d43ab727b6e8963bcb4162e)

inducer:
LGTM, thanks!

PR #289, opened 2022-03-24
Remove np.nan from expression trees

kaushikcfd:
Comparison between `np.NaN`s always results in False. This would inhibit our pickling capabilities.

Draft because:
- [x] Needs https://github.com/inducer/loopy/pull/593
- [x] Containts commits from #293

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1996,7 +1996,14 @@ def full(shape: ConvertibleToShape, fill_value: ScalarType,
         dtype = np.dtype(dtype)
 
     shape = normalize_shape(shape)
-    return IndexLambda(dtype.type(fill_value), shape, dtype, {},
+
+    if np.isnan(fill_value):
+        from pymbolic.primitives import NaN
+        fill_value = NaN(dtype.type)
+    else:
+        fill_value = dtype.type(fill_value)
+
+    return IndexLambda(fill_value, shape, dtype, {},
                        tags=_get_default_tags(),
                        axes=_get_default_axes(len(shape)))
 
@@ -2310,7 +2317,8 @@ def maximum(x1: ArrayOrScalar, x2: ArrayOrScalar) -> ArrayOrScalar:
             or np.issubdtype(common_dtype, np.complexfloating)):
         from pytato.cmath import isnan
         # https://github.com/python/mypy/issues/3186
-        return where(logical_or(isnan(x1), isnan(x2)), np.NaN,  # type: ignore
+        return where(logical_or(isnan(x1), isnan(x2)),  # type: ignore
+                     common_dtype.type(np.NaN),
                      where(greater(x1, x2), x1, x2))
     else:
         return where(greater(x1, x2), x1, x2)
@@ -2328,7 +2336,8 @@ def minimum(x1: ArrayOrScalar, x2: ArrayOrScalar) -> ArrayOrScalar:
             or np.issubdtype(common_dtype, np.complexfloating)):
         from pytato.cmath import isnan
         # https://github.com/python/mypy/issues/3186
-        return where(logical_or(isnan(x1), isnan(x2)), np.NaN,  # type: ignore
+        return where(logical_or(isnan(x1), isnan(x2)),  # type: ignore
+                     common_dtype.type(np.NaN),
                      where(less(x1, x2), x1, x2))
     else:
         return where(less(x1, x2), x1, x2)
--- a/pytato/scalar_expr.py
+++ b/pytato/scalar_expr.py
@@ -263,6 +263,7 @@ class Reduce(ExpressionBase):
     def __getinitargs__(self) -> Tuple[ScalarExpression, ReductionOperation, Any]:
         return (self.inner_expr, self.op, self.bounds)
 
+    init_arg_names = ("inner_expr", "op", "bounds")
     mapper_method = "map_reduce"
 
 # }}}
--- a/pytato/utils.py
+++ b/pytato/utils.py
@@ -158,7 +158,13 @@ def update_bindings_and_get_broadcasted_expr(arr: ArrayOrScalar,
     """
 
     if isinstance(arr, SCALAR_CLASSES):
-        return arr
+        if np.isnan(arr):
+            # allowing NaNs to stay in our expression trees could potentially
+            # lead to spuriously unequal comparisons between expressions
+            from pymbolic.primitives import NaN
+            return NaN(np.array(arr).dtype.type)
+        else:
+            return arr
 
     assert isinstance(arr, Array)
     bindings[bnd_name] = arr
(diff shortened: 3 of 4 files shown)


pytato/scalar_expr.py
```diff
@@ -263,8 +268,25 @@ def __hash__(self) -> int:
     def __getinitargs__(self) -> Tuple[ScalarExpression, ReductionOperation, Any]:
         return (self.inner_expr, self.op, self.bounds)
 
+    init_arg_names = ("inner_expr", "op", "bounds")
     mapper_method = "map_reduce"
 
 # }}}
 
+
+class NaN(ExpressionBase):
```
inducer, line 277: I'd be in favor of having this in pymbolic, because it'll utlimately be a problem all the way down the stack, e.g. in loopy.

inducer, line 277: This isn't a complete representation of a NaN, as NaN can have a "payload": https://en.wikipedia.org/wiki/NaN#Encoding

Should we bother with this?

pytato/scalar_expr.py
```diff
@@ -263,8 +268,25 @@ def __hash__(self) -> int:
...
+    """
+    An expression node representing not-a-number. Unlike, :data:`math.nan` all
+    instances of :class:`NaN` would be equal. This is because we intend to use
+    NaN in a symbolic sense.
```
inducer, line 281:
```suggestion
    An expression node representing not-a-number as a floating point number.
    Unlike, :data:`math.nan`, all instances of :class:`NaN` compare equal,
    as one might reasonably expect for program representation. (If this weren't
    so, programs containing NaNs would effectively be unhashable, because
    they don't compare equal to themselves.)
    
    Note that, in Python, this equality comparison is made *even* more
    complex by `this bug <https://bugs.python.org/issue21873>`__, due
    to which ``np.nan == np.nan`` is *False*, but ``(np.nan,) == (np.nan,)``
    is True.
```

pytato/scalar_expr.py
inducer, line 277: Just like other floating point numbers, NaNs are sized. It might be useful to represent this sizing, maybe via a `cast_func`, which could be `None`, `np.float64`, or `np.float32`. What do you think? Or are we going to hack loopy's type inference to know about (and not consider the size of) NaNs?

inducer (commented):
Thanks! Awesome detective work. A few thoughts below.

pytato/scalar_expr.py
kaushikcfd: Thanks, see https://github.com/inducer/pymbolic/pull/88.

kaushikcfd: Avoided a payload for now. I've never come across a use for this. Probably we _might_ need to revisit this in the future. Resolving this for now.

pytato/scalar_expr.py
kaushikcfd: Thanks a lot! Done as a part of https://github.com/inducer/pymbolic/pull/88.

pytato/scalar_expr.py
kaushikcfd: Thanks, done as a part of https://github.com/inducer/pymbolic/pull/88. :)

test/test_pytato.py
```diff
@@ -790,6 +790,39 @@ def test_einsum_dot_axes_has_correct_dim():
     assert len(einsum.axes) == einsum.ndim
 
 
+def test_pickling_and_unpickling_is_equal():
```
inducer, line 793: Does this test involve a NaN somewhere?

pytato/utils.py
```diff
@@ -158,7 +158,13 @@ def update_bindings_and_get_broadcasted_expr(arr: ArrayOrScalar,
     """
 
     if isinstance(arr, SCALAR_CLASSES):
-        return arr
+        if np.isnan(arr):
+            # allowing NaNs to stay in our expression trees could potentially
+            # lead to incorrect comparisons between expressions
```
inducer, line 163:
```suggestion
            # lead to spuriously unequal between expressions
            # (because NaN != NaN by definition)
```

test/test_pytato.py
inducer, line 793: Maybe add a comment to remind the reader about the concern regarding NaNs.

pytato/array.py
```diff
@@ -2328,7 +2336,8 @@ def minimum(x1: ArrayOrScalar, x2: ArrayOrScalar) -> ArrayOrScalar:
             or np.issubdtype(common_dtype, np.complexfloating)):
         from pytato.cmath import isnan
         # https://github.com/python/mypy/issues/3186
-        return where(logical_or(isnan(x1), isnan(x2)), np.NaN,  # type: ignore
+        return where(logical_or(isnan(x1), isnan(x2)),  # type: ignore
+                     common_dtype.type(np.NaN),
```
inducer, line 2340: (Unrelated to this location)

Would it be useful to error (or at least warn) when an actual NaN is encountered during codegen, to make sure we're aware if one 'slips through'?

inducer (approved):
Thanks! Just a few things, LGTM after that.

test/test_pytato.py
kaushikcfd: They do, `pt.maximum` or `pt.minimum` is one of the possible node. However, I also added some expression graphs that deterministically have a NaN in them.

pytato/utils.py
kaushikcfd: Thanks, done!

test/test_pytato.py
kaushikcfd: Thanks, done!

pytato/array.py
kaushikcfd: Loopy already implements that: https://github.com/inducer/loopy/blob/53bb939913a993f2bb81e229384b273fa1034051/loopy/target/c/codegen/expression.py#L428-L432. So I guess not quite necessary here?

inducer: `arraycontext` has a dict that uses pytato DAGs as keys.

kaushikcfd: 1. I think there is no issue in comparing NaNs in the same interpreter run, so `arraycontext` comparisons should be fine. (cf <https://bugs.python.org/issue21873> and <https://github.com/inducer/pymbolic/blob/21ce5e6116498e2310409504402397c3acbcfe81/pymbolic/primitives.py#L554-L556>)
2. If we plan on emitting a warning (an error would be too harsh since the worst consequence of this is DAGs pickle unequally) then that could be conceived as repetitive as loopy already emits that warning during codegen.

inducer: Fair enough, let's wait and see if it becomes an issue.

inducer:
LGTM, thanks!

PR #291, opened 2022-03-25
Fixes to achieve deterministic codegen

kaushikcfd:
Will implement a regression in a separate PR, as soon as a solution for https://github.com/inducer/pytato/pull/289#discussion_r834778926 lands.


inducer:
LGTM, thanks!

inducer:
For posterity's reference: https://github.com/inducer/pymbolic/pull/88 implements the NaN type in pymbolic.

PR #292, opened 2022-03-25
Fix visualization of index nodes

inducer:
Previously, arrays in `indices` showed up stringified in the node, rather than as an edge.

### DIFF
--- a/pytato/visualization.py
+++ b/pytato/visualization.py
@@ -40,7 +40,8 @@ from pytato.loopy import LoopyCall
 
 from pytato.array import (
         Array, DataWrapper, DictOfNamedArrays, IndexLambda, InputArgumentBase,
-        Stack, ShapeType, Einsum, Placeholder, AbstractResultWithNamedArrays)
+        Stack, ShapeType, Einsum, Placeholder, AbstractResultWithNamedArrays,
+        IndexBase)
 
 from pytato.codegen import normalize_outputs
 from pytato.transform import CachedMapper, ArrayOrNames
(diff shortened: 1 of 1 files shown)


pytato/visualization.py
```diff
@@ -162,6 +164,36 @@ def map_stack(self, expr: Stack) -> None:
 
     map_concatenate = map_stack
 
+    def map_basic_index(self, expr: IndexBase) -> None:
+        info = self.get_common_dot_info(expr)
+
+        indices_parts = []
+        for i, index in enumerate(expr.indices):
+            if isinstance(index, (int, np.integer)):
```
kaushikcfd, line 172: I think this misses a few int-types. Better to use: https://github.com/inducer/pytato/blob/c8189578bc1e165746bd385a41c4b4db5f6b0735/pytato/scalar_expr.py#L74-L75?

kaushikcfd (approved):
LGTM, Thank you!

pytato/visualization.py
kaushikcfd:
```python
>>> for i in (int, np.int8, np.int16, np.int32, np.int64, np.uint8,
...           np.uint16, np.uint32, np.uint64):
...     a = i(2)
...     print(isinstance(a, (int, np.integer)))
... 
True
True
True
True
True
True
True
True
True
```

Oops no, I was wrong.

inducer: Yep, thanks: https://github.com/inducer/pytato/compare/0b21b7a34b195aeb9228d89d0d08a31c982410d2..34d8b5f37c21934f9ea7c28af7edebb946303f3b

PR #293, opened 2022-03-25
Some dtype fixes

kaushikcfd:
- maximum/minimum: should have NaN comparisons only for inexact dtypes
- full: Use the default `dtype=None`


kaushikcfd:
This is pretty simple, setting to auto-merge.

PR #294, opened 2022-03-26
Defines the direction in our Directed Acyclic Graph

kaushikcfd:


doc/design.rst
```diff
@@ -81,6 +81,13 @@ Computation and Results
     Operations that *can* be expressed as :class:`~pytato.array.IndexLambda`
     without loss of information, *should* be expressed that way.
 
+- Every array node denotes a computation graph. In :mod:`pytato` the
+  computation graphs are always Directed Acyclic Graphs (DAG). This is a direct
+  consequence of an :class:`~pytato.array.Array` instance being immutable. We
+  choose the direction of the edges in the DAG to resemble the one typically
+  seen in a data flow graph, i.e. the successors of a node are its *users* and
+  the predecessors of a node are the arrays that it *uses*.
+
```
matthiasdiener, line 90: The items above have three spaces after the dash.

doc/design.rst
```diff
@@ -81,6 +81,13 @@ Computation and Results
     Operations that *can* be expressed as :class:`~pytato.array.IndexLambda`
     without loss of information, *should* be expressed that way.
 
+- Every array node denotes a computation graph. In :mod:`pytato` the
+  computation graphs are always Directed Acyclic Graphs (DAG). This is a direct
```
matthiasdiener, line 85:
```suggestion
  computation graphs are always Directed Acyclic Graphs (DAGs). This is a direct
```

doc/design.rst
matthiasdiener, line 90: Maybe it's worthwhile to add something like this:

```suggestion

  Note that although DAGs are usually visualized with edges from nodes to its users, they are usually traversed
  in the opposite direction, that is, going from a node to the the nodes that it uses.
```

matthiasdiener (commented):
This looks like a great addition to the manual!

doc/design.rst
kaushikcfd: Thanks!

doc/design.rst
kaushikcfd: Thanks, fixed!

kaushikcfd: Thanks, added as a "gotcha" in 116e4b2.

matthiasdiener (approved):
LGTM, thanks!

inducer:
Thanks!

PR #300, opened 2022-04-01
CI: Force Open MPI for pytest

matthiasdiener:
Seems like recent changes in conda-forge prefer mpich over openmpi.

Should fix errors like https://github.com/inducer/pytato/runs/5791092495


inducer:
Thanks!

PR #302, opened 2022-04-02
deduplicate_data_wrappers: Support Empty Buffers

kaushikcfd:
The provided regression fails on `main`.

### DIFF
--- a/pytato/transform.py
+++ b/pytato/transform.py
@@ -1500,9 +1500,12 @@ def _get_data_dedup_cache_key(ary: DataInterface) -> Hashable:
     import sys
     if "pyopencl" in sys.modules:
         from pyopencl.array import Array as CLArray  # type: ignore[import]
+        from pyopencl import Buffer as CLBuffer
         if isinstance(ary, CLArray):
+            # pyopencl represents 0-long arrays' base_data as None
+            assert isinstance(ary.base_data, CLBuffer) or (ary.base_data is None)
             return (
-                    ary.base_data.int_ptr,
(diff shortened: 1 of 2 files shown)


test/test_codegen.py
```diff
@@ -1592,6 +1592,34 @@ def get_np_input_args():
     np.testing.assert_allclose(pt_result, np_result)
 
 
+def test_zero_size_cl_array_dedup(ctx_factory):
+    # before pytato@0d8b909 this regression would fail as
```
inducer, line 1596:
```suggestion
    # At pytato@0d8b909 this regression would fail as
```

pytato/transform.py
```diff
@@ -1500,14 +1500,26 @@ def _get_data_dedup_cache_key(ary: DataInterface) -> Hashable:
...
+                        None,
+                        ary.offset,
+                        ary.shape,
+                        ary.strides,
+                        ary.dtype,
+                        )
+            else:
+                assert isinstance(ary.base_data, CLBuffer)
+                return (
+                        ary.base_data.int_ptr,
```
inducer, line 1517: Why not put the conditional into this tuple entry?

test/test_codegen.py
kaushikcfd: Thanks!

pytato/transform.py
kaushikcfd: Yep, better. Thanks!

inducer:
Thx!

PR #304, opened 2022-04-07
add PooledBuffer to dedup cache key

matthiasdiener:
Followup of #302.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -35,5 +35,5 @@ sys.PYTATO_BUILDING_SPHINX_DOCS = True
 nitpick_ignore_regex = [
     ["py:class", r"numpy.(u?)int[\d]+"],
     ["py:class", r"pyrsistent.typing.(.+)"],
-
+    ["py:class", r"typing_extensions(.+)"],
 ]
--- a/pytato/transform.py
+++ b/pytato/transform.py
@@ -1717,7 +1717,8 @@ def _get_data_dedup_cache_key(ary: DataInterface) -> Hashable:
     if "pyopencl" in sys.modules:
         from pyopencl.array import Array as CLArray  # type: ignore[import]
         from pyopencl import Buffer as CLBuffer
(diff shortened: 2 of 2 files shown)


kaushikcfd:
I wonder if we should check for `pyopencl._cl.MemoryObjectHolder` instead of `(CLBuffer, PooledCLBuffer)`.

doc/conf.py
```diff
@@ -35,5 +35,5 @@
 nitpick_ignore_regex = [
     ["py:class", r"numpy.(u?)int[\d]+"],
     ["py:class", r"pyrsistent.typing.(.+)"],
-
```
matthiasdiener, line 38: Is there a better way to fix this doc build error:

```
/home/runner/work/pytato/pytato/.miniforge3/envs/testing/lib/python3.10/site-packages/pytools/tag.py:docstring of pytools.tag.Taggable.tagged:: WARNING: py:class reference target not found: typing_extensions.Self

/home/runner/work/pytato/pytato/.miniforge3/envs/testing/lib/python3.10/site-packages/pytools/tag.py:docstring of pytools.tag.Taggable.without_tags:: WARNING: py:class reference target not found: typing_extensions.Self
```

matthiasdiener:
> I wonder if we should check for `pyopencl._cl.MemoryObjectHolder` instead of `(CLBuffer, PooledCLBuffer)`. (@inducer might have a better handle at this)

Done in [b0023d2](https://github.com/inducer/pytato/pull/304/commits/b0023d2666b0a0bfd35ff13d8e7d247710867fc4)

doc/conf.py
inducer: Not that I'm aware:

https://github.com/inducer/pytools/blob/9a21c88c241da64407e78141977870df9ea2d64c/doc/conf.py#L36

inducer:
Thanks!

PR #306, opened 2022-04-13
Remove uses of newly-deprecated symbols in `pytools.tag`

inducer:


inducer:
Simple :arrow_right: auto merge.

Issue #307, opened 2022-04-13
Strip `PartIDTag` from outputs of a computation


inducer:
I was debugging the failure in https://github.com/inducer/grudge/runs/5982131517?check_suite_focus=true
```
Traceback (most recent call last):
  File "/home/runner/work/grudge/grudge/examples/wave/wave-op-mpi.py", line 335, in <module>
    main(cl.create_some_context,
  File "/home/runner/work/grudge/grudge/examples/wave/wave-op-mpi.py", line 278, in main
    fields = compiled_lsrk45_step(actx, fields, t, dt, compiled_rhs)
  File "/home/runner/work/grudge/grudge/grudge/shortcuts.py", line 55, in compiled_lsrk45_step
    y, residual  = update(y, a, b, h, residual, rhs_val)
  File "/home/runner/work/grudge/grudge/.miniforge3/envs/testing/lib/python3.10/site-packages/arraycontext/impl/pytato/compile.py", line 363, in __call__
    compiled_func = self._dag_to_compiled_func(
  File "/home/runner/work/grudge/grudge/grudge/array_context.py", line 143, in _dag_to_compiled_func
    distributed_partition = find_distributed_partition(dict_of_named_arrays)
  File "/home/runner/work/grudge/grudge/.miniforge3/envs/testing/lib/python3.10/site-packages/pytato/distributed.py", line 914, in find_distributed_partition
    partitioned_outputs = make_dict_of_named_arrays({
  File "/home/runner/work/grudge/grudge/.miniforge3/envs/testing/lib/python3.10/site-packages/pytato/distributed.py", line 915, in <dictcomp>
    name: part_id_tag_assigner(subexpr, stored_ary_to_part_id[subexpr])
  File "/home/runner/work/grudge/grudge/.miniforge3/envs/testing/lib/python3.10/site-packages/pytato/transform.py", line 169, in __call__
    return self.rec(expr, *args, **kwargs)
  File "/home/runner/work/grudge/grudge/.miniforge3/envs/testing/lib/python3.10/site-packages/pytato/distributed.py", line 792, in rec
    result = super().rec(expr, user_part_id)
[snip]
pytools.tag.NonUniqueTagError: Multiple tags are direct subclasses of the following UniqueTag(s): PartIDTag
```
and it occurred to me that we never strip the `PartIDTag` from outputs of distributed computations. While I haven't verified that this is the source of this bug, it seems plausible:
[...]

PR #308, opened 2022-04-18
Remove `PartIDTag` from nodes in partition 

inducer:
See #307.

- [x] May need a pytools release for https://github.com/inducer/pytools/pull/126: https://pypi.org/project/pytools/2022.1.4/

Best read commit-by-commit.

### DIFF
--- a/pytato/distributed.py
+++ b/pytato/distributed.py
@@ -165,7 +165,7 @@ class DistributedSend(Taggable):
         comm_tag: Optional[CommTagType] = kwargs.get("comm_tag")
         tags = cast(FrozenSet[Tag], kwargs.get("tags"))
         return type(self)(
-                data=data or self.data,
+                data=data if data is not None else self.data,
                 dest_rank=dest_rank if dest_rank is not None else self.dest_rank,
                 comm_tag=comm_tag if comm_tag is not None else self.comm_tag,
                 tags=tags if tags is not None else self.tags)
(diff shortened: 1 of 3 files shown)


inducer:
@kaushikcfd This is now passing and ready for a look.

pytato/distributed.py
```diff
@@ -793,6 +829,15 @@ def rec(self,  # type: ignore[override]
             return result
 
 
+def _remove_part_id_tag(ary: ArrayOrNames) -> Array:
+    assert isinstance(ary, Array)
+
+    # FIXME This spurious assignment helps mypy understand this code.
+    # I wasn't able to reproduce this in isolation.
```
kaushikcfd, line 836: I don't think this particular comment line is too helpful but has the potential to confuse the reader.

kaushikcfd (approved):
LGTM, thanks! I had completely missed nested distributed comm nodes being a thing.

pytato/distributed.py
inducer: Could you explain? What I'm trying to do is explain why this doesn't say `return stuff`, the reason being a (perceived) mypy bug. Would you prefer I just get rid of the comment?

inducer:
>  I had completely missed nested distributed comm nodes being a thing.

To be clear, this is about executing one (distributed) workload that uses results from another, and then being confused by leftover partitioning tags. Maybe that's what you meant by "nesting", but I figured I'd clarify just to be sure.

kaushikcfd:
> Maybe that's what you meant by "nesting", but I figured I'd clarify just to be sure.

Yep, that's what I meant.

pytato/distributed.py
inducer: Turns out I was just being dumb about reproducing it. It reproduces just fine: https://github.com/python/mypy/issues/12626

inducer: (now linked)

inducer:
Thanks for taking a look!

PR #311, opened 2022-04-22
Adds tests that check partitioning, codegen are reproducible

kaushikcfd:
Draft because:
- [x] Includes commits from #289

### DIFF
--- a/test/test_codegen.py
+++ b/test/test_codegen.py
@@ -44,7 +44,7 @@ import pytest  # noqa
 from loopy.version import LOOPY_USE_LANGUAGE_VERSION_2018_2  # noqa
 
 import pytato as pt
-from testlib import assert_allclose_to_numpy
+from testlib import assert_allclose_to_numpy, get_random_pt_dag
 import pymbolic.primitives as p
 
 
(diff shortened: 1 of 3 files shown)


inducer:
Clicked rebase, couldn't help myself. Hope that's OK!

test/test_codegen.py
```diff
@@ -1620,6 +1620,77 @@ def test_zero_size_cl_array_dedup(ctx_factory):
     assert num_nodes_new == (num_nodes_old - 2)
 
 
+# {{{ test_deterministic_codegen
+
+def _get_random_dag_with_comm_nodes(seed):
```
inducer, line 1625: "comm nodes" sounds like "communication nodes", which I'm taking to mean MPI/distributed. But I don't see any of those here? (Looks like maybe this is a copy-paste from the distributed test? Can those be unified and put into testlib? Does that make sense?)

test/test_codegen.py
```diff
@@ -1620,6 +1620,77 @@ def test_zero_size_cl_array_dedup(ctx_factory):
...
+        seed = 120 + i
+        results = mp_ctx.Array("i", (0, ) * nprocs)
+        print(f"Step {i} {seed}")
+
+        ref_dag = _get_random_dag_with_comm_nodes(seed)
+        ref_t_unit = pt.generate_loopy(ref_dag).program
+
+        # {{{ spawn nprocs-processes and verify they all compare equally
+
+        procs = [mp_ctx.Process(target=_check_deterministic_codegen,
```
inducer, line 1673: Would it make sense to explicitly set `PYTHONHASHSEED` to different things?

inducer (commented):
Thanks! Two questions here ahead of a proper review.

test/test_codegen.py
kaushikcfd: Oops, yep. The name was copy-pasted blindly and hence incorrect. (Thanks!)

test/test_codegen.py
kaushikcfd: Yep, now we set PYTHONHASHSEED before spawning the processes.

inducer (approved):
Thanks! LGTM once you've got mypy sorted out.

kaushikcfd:
> Thanks! LGTM once you've got mypy sorted out.

Thanks, fixed! Setting to "Automerge".

PR #312, opened 2022-04-28
index_lambda_to_hlo: Allow NaN as a scalar

kaushikcfd:
/cc @matthiasdiener 

The provided regression fails on `main`.


inducer:
LGTM, thanks!

PR #313, opened 2022-04-29
mypy ignore typeshed-issue-7739

kaushikcfd:
Mypy CI was failing on `main`.


inducer:
Thx! (And also thanks for reporting what appears to be the underlying issue upstream!)

PR #315, opened 2022-05-03
Doc fix

nchristensen:
https://en.wiktionary.org/wiki/purposely
https://en.wiktionary.org/wiki/purposefully


inducer:
Thanks. Based on the articles you linked, I am still not sure when I would use one or the other. They list each other as synonyms... Could you explain?

inducer:
OIC, TIL. Thanks!

PR #320, opened 2022-05-10
Do not emit statements or domains for empty arrays

inducer:
The root of this is @majosm's work on multi-volume DG, which puts empty arrays in the DAG. This used to emit statements executed on empty domains, which loopy didn't like. Also, since there's literally nothing to compute, we might as well drop them, for a potential cost savings.

- [x] Point loopy back to `main`
- [x] Needs https://github.com/inducer/loopy/pull/620


kaushikcfd (approved):
LGTM, thanks!

PR #322, opened 2022-05-19
Check identity in EqualityComparisonMapper

kaushikcfd:
Not doing so would lead to a quadratic equality comparison cost in
if EqualityComparisonMapper is created and destroyed frequently, which
is typically the case in a CachedMapper.

Consider the graphs:

dag1 = 2 * x
dag2 = 2 * x  # note that 'dag1 is not dag2', but dag1 == dag2

where 'x' is a very large graph. Without this patch, comparing dag1 and
dag2 would redundantly traverse all elements of 'x'.

---
Experiment -- deduplicating data wrappers for graphs implemented in https://github.com/illinois-ceesd/drivers_y2-isolator would take ~7 minutes, but after this patch takes 3 seconds.

### DIFF
--- a/pytato/equality.py
+++ b/pytato/equality.py
@@ -82,7 +82,7 @@ class EqualityComparer:
                 else:
                     result = self.map_foreign(expr1, expr2)
             else:
-                result = method(expr1, expr2)
+                result = (expr1 is expr2) or method(expr1, expr2)
 
             self._cache[cache_key] = result
             return result


inducer:
Thanks, nice find!

- You've got a typo in the commit message ("compar**s**ison") (I'm OK with not fixing that, LMK.)
- I wonder whether we should preemptively act on the off-hand comment you make, namely that cached mappers don't keep their equality mappers around. Though that's perhaps not as easy, because we're not the ones calling `==` (the `dict` is)... but maybe that can be hackered around? Maybe that's for a different PR though. What do you think?

kaushikcfd:
Typo: fixed, thanks!
Dict: Agreed, converting to an issue for now as it's not quite a significant cost just yet.

PR #324, opened 2022-05-21
Implement pytato.expand_dims

kaushikcfd:
expand_dims has the potential to hold more metadata than a plain reshape and has cleaner axis-type propagation semantics


inducer:
Thanks, LGTM!

PR #327, opened 2022-05-23
materialize_with_mpms: log number of materialized nodes

matthiasdiener:


pytato/transform/__init__.py
```diff
@@ -1296,13 +1296,19 @@ def materialize_with_mpms(expr: DictOfNamedArrays) -> DictOfNamedArrays:
...
     new_data = {}
     for name, ary in expr.items():
         new_data[name] = materializer(ary.expr).expr
 
-    return DictOfNamedArrays(new_data)
+    res = DictOfNamedArrays(new_data)
+
+    logger.info("materialize_with_mpms: materialized "
+        f"{get_num_tags_of_type(res, ImplStored())} out of "
+        f"{get_num_nodes(res)} nodes")
```
inducer, line 1309: This costs two extra (avoidable) traversals. I'd prefer to not do those by default, perhaps only if asked for "enhanced diagnostics"?

matthiasdiener: Should we wrap this in a `if __debug__` or maybe a `if logger.isEnabledFor(logging.INFO):` ?

inducer: I'd perhaps prefer if this were done where `materialize_with_mpms` is called, with a flag passed to that array context that additional diagnostic precision is desired.

inducer: Sure, not opposed. But I'd also like to be able to control it programmatically.

matthiasdiener: [262787c](https://github.com/inducer/pytato/pull/327/commits/262787c68c32e799ced65c0d3eb1440ce2a50c83) implements the approach based on the env variable (#332)

inducer:
Thanks!

inducer:
I was very surprised just now to find out that this didn't go into main. If you have PRs going anywhere but main, please point this out!

matthiasdiener:
> I was very surprised just now to find out that this didn't go into main. If you have PRs going anywhere but main, please point this out!

Oh, I'm sorry!

PR #329, opened 2022-05-24
More precise `DataWrapper.__eq__`

kaushikcfd:


test/test_codegen.py
```diff
@@ -1601,13 +1601,13 @@ def test_zero_size_cl_array_dedup(ctx_factory):
...
-    x3 = pt.make_data_wrapper(x_cl2)
-    x4 = pt.make_data_wrapper(x_cl2)
+    x1 = pt.make_data_wrapper(x_cl1[:])
+    x2 = pt.make_data_wrapper(x_cl1[:])
+    x3 = pt.make_data_wrapper(x_cl2[:])
+    x4 = pt.make_data_wrapper(x_cl2[:])
 
     out = pt.make_dict_of_named_arrays({"out1": 2*x1,
-                                        "out2": 2*x2,
+                                        "out2": 3*x2,
```
kaushikcfd, line 1610: Had to change this as the test was sensitive to the number of nodes in the graph.

kaushikcfd:
Just timed the operation `pt.transform.dedup_data_wrapper` on a graph with 36k nodes with and without this patch and :drum:  there's no difference. (This was lying around in my tree for a long time and had thought this was somehow helpful in keeping the compile times down. It isn't)

I'll close this. Thanks for the feedback!

kaushikcfd:
> Would this not effectively make the deduplication of DataWrappers just a CopyMapper?

Had missed this before, no that wouldn't made deduplicaion a copy operation as thawing performs `cl_array.with_queue(...)` and that returns a new instance of `cl_array` every time.

inducer:
I'm not sure I understand that. Deduplication operates on the symbolic DAG, but `with_queue` operates on arrays that are actual data...?

PR #331, opened 2022-05-26
Call BranchMorpher after dw deduplication

kaushikcfd:
Some timings on the graphs (~36k nodes) implemented in https://github.com/illinois-ceesd/drivers_y2-isolator:

With this patch:
```
$ time python -O -m mpi4py isolator_injection_run.py -i run_params_combustion.yaml -r restart_data/isolator_init-000000 --log --lazy
real    2m48.997s
user    2m51.500s
sys     0m10.912s
```
Without this patch:
```
$ time python -O -m mpi4py isolator_injection_run.py -i run_params_combustion.yaml -r restart_data/isolator_init-000000 --log --lazy                                                                                                                                                                                                                                             
real    3m20.631s                                                                                                                                                                                                                              
user    3m22.686s                                                                                                      
sys     0m11.08

### CODE REVIEW

File: pytato/transform/__init__.py [hunk 1]
```diff
@@ -1706,6 +1710,33 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
 # }}}
 
 
+# {{{ BranchMorpher
+
+class BranchMorpher(CopyMapper):
```
inducer, line 1715: Isn't this exactly the same as a `CopyMapper` (which already inherits `CachedMapper`?

inducer, line 1715: How come the deduplication doesn't already eliminate these identical parts of the tree? I feel like it should.

inducer, line 1715: I'm not loving the name. It provides me no meaningful context on what this mapper does. Maybe something like `EqualBranchesReuser`?

kaushikcfd: These are different at the time when equality happens. Over here smaller nodes (like placeholders) are first found to be equal and identical objects are returned. This process makes the equality comparison $O(N)$ as opposed to $O(N^2)$ in the CopyMapper.

kaushikcfd: Yes, they do eliminate it but would be expensive.

kaushikcfd: Done!

pytato/transform/__init__.py
```diff
@@ -1706,6 +1711,33 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
 # }}}
 
 
+# {{{ EqualBranchesReuser
```
inducer, line 1714: Sorry to be quibbling about the name again. With better understasnding comes better naming, maybe. How about `PostMapEqualNodeReuser`?

pytato/transform/__init__.py
```diff
@@ -1706,6 +1711,33 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
...
+
+class EqualBranchesReuser(CopyMapper):
+    """
+    A mapper that replaces equal segments of graphs with identical objects.
```
inducer, line 1718: Please add more context to the docstring, contrastng with `CachedMapper` and specifically mentioning that this is about equality after the mapper applies, and describe the resulting complexity reduction.

kaushikcfd: Done!

pytato/transform/__init__.py
kaushikcfd: Done, thanks!

pytato/transform/__init__.py
```diff
@@ -1706,6 +1711,48 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
 # }}}
 
 
+# {{{ PostMapEqualNodeReuser
+
+class PostMapEqualNodeReuser(CopyMapper):
+    """
+    A mapper that reuses the same object instances for equal segments of
+    graphs.
```
inducer, line 1719:
```suggestion
    A mapper that reuses the same object instances for segments of
    graphs that become equal after processing by subclasses of this mapper,
    where the equality test happens immediately after a new node is created.
```

pytato/transform/__init__.py
```diff
@@ -1706,6 +1711,48 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
...
+
+    .. note::
+
+        The operation performed here is equivalent to that of a
+        :class:`CopyMapper`, in that both return a single instance for equal
+        :class:`pytato.Array` nodes. However, they differ at the point where
+        two array expressions are compared. :class:`CopyMapper` compares array
+        expressions before the expressions are mapped i.e. repeatedly comparing
+        equal array expressions but unequal instances, and because of this it
```
inducer, line 1728:
```suggestion
        different instances representing equal array expressions, and, because of this, it
```

pytato/transform/__init__.py
```diff
@@ -1706,6 +1711,48 @@ def map_distributed_recv(self, expr: DistributedRecv, *args: Any) \
...
+        spends super-linear time in comparing array expressions.  On the other
+        hand, :class:`PostMapEqualNodeReuser` has linear complexity in the
+        number of nodes in the number of array expressions as the larger mapped
+        expressions already contain same instances for the predecessors,
+        resulting in a cheaper equality comparison overall.
```
inducer, line 1733: I think what you want to say instead here is that this avoids building potentially large repeated subexpexpressions in the first place, rather than having to (expensively) merge them after the fact.

pytato/transform/__init__.py
```diff
@@ -1782,8 +1829,11 @@ def cached_data_wrapper_if_present(ary: ArrayOrNames) -> ArrayOrNames:
                 len(data_wrapper_cache),
                 data_wrappers_encountered - len(data_wrapper_cache))
 
-    return array_or_names
+    # many paths in the DAG might be semantically equivalent after DWs are
+    # deduplicated => morph them
+    return PostMapEqualNodeReuser()(array_or_names)
```
inducer, line 1834: Hmm, wait. If it's used like this, I would argue that that's equivalent to just using a `CopyMapper`. I figured the data-wrapper-deduplicator should inherit from this to avoid building large identical subtrees in the first place.

pytato/transform/__init__.py
```diff
@@ -1782,8 +1829,11 @@ def cached_data_wrapper_if_present(ary: ArrayOrNames) -> ArrayOrNames:
                 len(data_wrapper_cache),
                 data_wrappers_encountered - len(data_wrapper_cache))
 
-    return array_or_names
+    # many paths in the DAG might be semantically equivalent after DWs are
+    # deduplicated => morph them
```
inducer, line 1833: Still: "morph" isn't good terminology.

inducer (commented):
Thanks! A few more notes/thoughts below.

pytato/transform/__init__.py
inducer: Added `test_post_map_equal_node_reuser_intestine` to investigate whether this is equivalent to a `CopyMapper`, and at least in that example, it appears to be. It remains to figure out where the difference in execution time comes from.

PR #332, opened 2022-05-31
add PYTATO_DEBUG env variable

matthiasdiener:


inducer:
> Not sure if this should be documented.

I'm not sure about the idea, so I'd like to have the option of just removing it if it doesn't pan out.

matthiasdiener:
> > Not sure if this should be documented.
> 
> I'm not sure about the idea, so I'd like to have the option of just removing it if it doesn't pan out.

In other words, don't document it for now?

PR #333, opened 2022-06-01
add materialization counter

matthiasdiener:


pytato/analysis/__init__.py
```diff
@@ -426,3 +429,27 @@ def get_num_tags_of_type(
     return tcm(outputs)
 
 # }}}
+
+
+def get_num_materialized(outputs: Union[Array, DictOfNamedArrays]) \
```
inducer, line 434: My main issue with this approach is that it has at least quadratic complexity. A possible improvement would be a `CombineMapper` that simply counts the materialized predecessors for each node. I.e. `combine` is a sum, and each node returns `1` if it's materialized and uses inherited behavior (reduce ("combine") over predecessors).

PR #336, opened 2022-06-08
Various cleanups

inducer:


pytato/array.py
```diff
@@ -317,9 +317,9 @@ class Axis(Taggable):
     """
     tags: FrozenSet[Tag]
 
-    def copy(self, **kwargs: Any) -> Axis:
+    def _with_new_tags(self, tags: FrozenSet[Tag]) -> Taggable:
```
kaushikcfd, line 320: Should this also return `Self`? (cf https://github.com/inducer/pytools/blob/0a06f30fc746d59cc214b6e2e982fe34915327f3/pytools/tag.py#L268)

pytato/transform/__init__.py
```diff
@@ -1776,7 +1776,7 @@ def cached_data_wrapper_if_present(ary: ArrayOrNames) -> ArrayOrNames:
     array_or_names = map_and_copy(array_or_names, cached_data_wrapper_if_present)
 
     if data_wrappers_encountered:
-        logger.info("data wrapper de-duplication: "
+        logger.debug("data wrapper de-duplication: "
                 "%d encountered, %d kept, %d eliminated",
                 data_wrappers_encountered,
                 len(data_wrapper_cache),
```
kaushikcfd, line 1782: Superficial: Violates [PEP8](https://peps.python.org/pep-0008/#indentation).

kaushikcfd (approved):
LGTM!

pytato/array.py
inducer: - It only makes a difference if this is subclassed. (Unlikely here.)
- `Self` actually isn't fully implemented in mypy yet: https://github.com/python/mypy/pull/11666 (kind of puzzled why it works at all in pytools)

pytato/transform/__init__.py
inducer: It's no worse than before. :)

TBH: I do not understand the rationale for this [PEP8 recommendation](https://pep8.org/#indentation):

> When using a hanging indent the following should be considered; there should be no arguments on the first line and further indentation should be used to clearly distinguish itself as a continuation line.

Using this code as an example, I don't find the arguments after the `(` easy to miss, however I find the waste of vertical space quite noticeable (which I find at a premium even on my terminal, which has 104 rows available, cf. `stty size`). That's why I typically disable this rule in flake8. I also find this suggestion by flake8 to be puzzling:
```py
foo = long_function_name(var_one, var_two,
                         var_three, var_four)
```
because it wastes horizontal *and* vertical space. I can see where the parenthesis is just fine. :)

inducer:
(I'll wait for your responses here before merging.)

pytato/transform/__init__.py
kaushikcfd: > It's no worse than before

That's true :-)

> because it wastes horizontal and vertical space.

That's fair. I personally find it easier to read arguments aligned with the parentheses. If the horizontal space starts becoming a concern one could insert a line break before the first argument.

But, it's subjective and so resolving :).

pytato/array.py
kaushikcfd: Fair, thanks for the links!

kaushikcfd:
LGTM from my end, thanks! We could just do "Enable auto-merge[rebase]" which will avoid us having to update the branch, but if that's not your preferred path please merge as per your preference.

inducer:
> LGTM from my end, thanks! We could just do "Enable auto-merge[rebase]" which will avoid us having to update the branch

If I do that, it just sits there. But there's a handy "update branch with rebase" button, which enables progress... so I clicked that. :)

> but if that's not your preferred path please merge as per your preference.

pytato/transform/__init__.py
inducer: Thanks for chiming in! I propose we stick with the current, "comfortably ambiguous" way, and leave the Flake8 rule disabled.

PR #337, opened 2022-06-12
Deprecate passing CL devices to pytato codegen

inducer:


inducer:
Trivial and largely a no-op, setting to merge without review.

PR #340, opened 2022-06-14
Remove `DataWrapper.name`

inducer:
We have the naming tags (and this starts honoring them). I don't see why we had `DataWrapper.name` attribute in the first place. On top of that, the implementation was AFAICT unsafe.

Predictably, the goal of this is to get better names into the generated code.

This *does* break compatibility in a small way (highlighted), but I think it is not a way that will impact any current users.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -1722,20 +1722,22 @@ class DataWrapper(InputArgumentBase):
         (i.e. the very same instance).
     """
 
-    _fields = InputArgumentBase._fields + ("data", "shape", "name")
+    _fields = InputArgumentBase._fields + ("data", "shape")
     _mapper_method = "map_data_wrapper"
 
     def __init__(self,
-            name: Optional[str],
             data: DataInterface,
             shape: ShapeType,
             axes: AxesT,
             tags: FrozenSet[Tag] = frozenset()):
         super().__init__(axes=axes, tags=tags)
 
-        self.name = name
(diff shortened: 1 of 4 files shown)


pytato/array.py
```diff
@@ -2100,6 +2102,7 @@ def make_size_param(name: str,
 
 
 def make_data_wrapper(data: DataInterface,
+        *,
```
inducer, line 2105: This is the compatibility break, which will allow us to cleanly remove `name` when the time comes.

pytato/codegen.py
```diff
@@ -159,9 +185,7 @@ def map_loopy_call(self, expr: LoopyCall) -> LoopyCall:
                          entrypoint=entrypoint)
 
     def map_data_wrapper(self, expr: DataWrapper) -> Array:
-        name = expr.name
```
inducer, line 162: I think this was unsafe: it did not check whether the name was taken.

kaushikcfd: [This](https://github.com/inducer/pytato/blob/a0fa84a5ca17a32f57431b8d67d0383f4c94839c/pytato/codegen.py#L615-L628) should have caught it.

inducer: Only if the clash is with another input, not an internally-generated name.

inducer:
Thanks for taking a look! Setting to merge.

PR #350, opened 2022-08-04
accept SVMPointer in _get_data_dedup_cache_key

matthiasdiener:


pytato/transform/__init__.py
```diff
@@ -1724,10 +1724,17 @@ def _get_data_dedup_cache_key(ary: DataInterface) -> Hashable:
     if "pyopencl" in sys.modules:
         from pyopencl.array import Array as CLArray  # type: ignore[import]
         from pyopencl import MemoryObjectHolder
+        try:
+            from pyopencl import SVMPointer
+            allowed_base_data_classes = tuple((SVMPointer, MemoryObjectHolder))
+        except ImportError:
+            allowed_base_data_classes = tuple((MemoryObjectHolder,))
```
inducer, line 1731:
```suggestion
            allowed_base_data_classes = (SVMPointer, MemoryObjectHolder)
        except ImportError:
            allowed_base_data_classes = (MemoryObjectHolder,)
```

matthiasdiener: mypy doesn't like that:

```
pytato/transform/__init__.py:1731: error: Incompatible types in assignment (expression has type "Tuple[Any]", variable has type "Tuple[Any, Any]")  [assignment]
```

inducer: You need to annotate one as `: Tuple[Type, ...]`

inducer:
Thanks!

PR #352, opened 2022-08-05
_compare: np.bool8 -> np.dtype(np.bool8)

kaushikcfd:
Reported by @mitkotak


kaushikcfd:
Trivial, setting to auto-merge.

kaushikcfd:
This would've been caught earlier if we didn't have the bare `type: ignore` earlier.

PR #354, opened 2022-08-06
Gets rid of all of bare type-ignores in pytato.array

kaushikcfd:
- uses np.number/np.integer for all numpy types
- Makes pytato.cmath functions take in ArrayOrScalar as input operands

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -153,7 +153,7 @@ Internal stuff that is only here because the documentation tool wants it
 # }}}
 
 from abc import ABC, abstractmethod, abstractproperty
-from functools import partialmethod
+from functools import partialmethod, cached_property
 import operator
 from dataclasses import dataclass
 from typing import (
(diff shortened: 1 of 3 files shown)


pytato/array.py
```diff
@@ -2489,7 +2496,7 @@ def maximum(x1: ArrayOrScalar, x2: ArrayOrScalar) -> ArrayOrScalar:
             or np.issubdtype(common_dtype, np.complexfloating)):
         from pytato.cmath import isnan
         # https://github.com/python/mypy/issues/3186
```
inducer, line 2498: Remove this comment given that the type ignore is gone?

inducer (commented):
This is nice, thanks! I only found one small wrinkle.

inducer:
I'll just take care of it.

inducer:
Thx!

PR #359, opened 2022-09-07
LoopyPyOpenCLTarget: pass through loopy.PyOpenCLTarget

matthiasdiener:
Needs:
- ~~https://github.com/inducer/loopy/pull/642~~

### DIFF
--- a/pytato/target/loopy/__init__.py
+++ b/pytato/target/loopy/__init__.py
@@ -85,17 +85,25 @@ class LoopyPyOpenCLTarget(LoopyTarget):
 
         The :mod:`pyopencl` device used to construct the
         :class:`loopy.PyOpenCLTarget`, or *None*.
+
+    .. attribute:: limit_arg_size_nbytes
+
+        An optional int that indicates the size limit of kernel arguments
+        and gets passed to :class:`loopy.PyOpenCLTarget`.
     """
 
-    def __init__(self, device: Optional["pyopencl.Device"] = None):
+    def __init__(self, device: Optional["pyopencl.Device"] = None,
+                 limit_arg_size_nbytes: Optional[int] = None):
         if device is not None:
             from warnings import warn
             warn("Passing 'device' is deprecated and will stop working in 2023.",
                     DeprecationWarning, stacklevel=2)
+        self.limit_arg_size_nbytes = limit_arg_size_nbytes
 
     def get_loopy_target(self) -> "loopy.LoopyPyOpenCLTarget":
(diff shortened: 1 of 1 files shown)


inducer:
Not sure I love this. In general, we could just set the target on the generated loopy translation unit after it has been generated. This would make pytato's logic entirely independent of the details of loopy target selection, which seems nicer. @kaushikcfd, would you agree?

kaushikcfd:
> This would make pytato's logic entirely independent of the details of loopy target selection, which seems nice

I agree, I think the principled way to do this is not leaking any details about the loopy kernel execution to `generate_loopy`.

pytato/target/loopy/__init__.py
```diff
@@ -85,17 +85,25 @@ class LoopyPyOpenCLTarget(LoopyTarget):
...
         :class:`loopy.PyOpenCLTarget`, or *None*.
+
+    .. attribute:: limit_arg_size_nbytes
+
+        An optional int that indicates the size limit of kernel arguments
+        and gets passed to :class:`loopy.PyOpenCLTarget`.
     """
 
-    def __init__(self, device: Optional["pyopencl.Device"] = None):
+    def __init__(self, device: Optional["pyopencl.Device"] = None,
```
matthiasdiener, line 95: Create a new constructor that accepts an `lp.PyOpenCLTarget()`

matthiasdiener: Is https://github.com/inducer/pytato/pull/359/commits/2942fbfd452bfbdad5b9fbc5e4b637cdb896e9ad what you had in mind @inducer?

matthiasdiener:
Testing this together with https://github.com/inducer/arraycontext/pull/189 appears to work as expected, and avoids the test failures seen in https://github.com/illinois-ceesd/mirgecom/pull/746

pytato/target/loopy/__init__.py
inducer: Sorry to be sending you around in circles, but I just realized we don't even need any pytato changes. We can simply subclass `LoopyPyOpenCLTarget` in the array context and override `get_loopy_target` there.

matthiasdiener: What do you think of https://github.com/inducer/arraycontext/pull/189/commits/2768feea8d7e4846d3c6df731d46238c8f776375 ?

matthiasdiener:
I think this PR can be closed right?

inducer:
Yep.

PR #364, opened 2022-10-14
Functions

inducer:
cc @kaushikcfd 

This is a draft because:
- [x] Contains commits from #387

PR #365, opened 2022-10-16
Use ~~dataclasses~~ attrs for pt.Array

kaushikcfd:
Pros:
- Immutable types (`frozen=True`)
- Derived classes can be dataclasses (noticed while implementing #364)
- In almost all cases uses the generated `__init__`.

Cons:
- All derived classes must do `eq=False` in their call to `dataclass`, otherwise `dataclasses` will implement its own default which has a complexity of $O(2^N)$ in number of nodes in the DAG.
- All derived classes must do `def __hash__(self) -> int: return super().__hash__()`, otherwise `dataclasses` will implement its own default.
- Constructor arguments have to be `_shape`, `_dtype` instead of `shape`, `dtype` as `pt.Array` already has properties of that name.
- :warning: Pylint Failures.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -148,14 +148,19 @@ Internal stuff that is only here because the documentation tool wants it
 .. class:: AxesT
 
     A :class:`tuple` of :class:`Axis` objects.
+
+.. class:: IntegralT
+
+    An integer data type which is a union of integral types of :mod:`numpy` and
+    :class:`int`.
 """
 
 # }}}
 
-from abc import ABC, abstractmethod, abstractproperty
+from abc import ABC, abstractmethod
 from functools import partialmethod, cached_property
 import operator
-from dataclasses import dataclass
+import attrs
 from typing import (
         Optional, Callable, ClassVar, Dict, Any, Mapping, Tuple, Union,
         Protocol, Sequence, cast, TYPE_CHECKING, List, Iterator, TypeVar,
@@ -288,7 +293,7 @@ def _truediv_result_type(arg1: DtypeOrScalar, arg2: DtypeOrScalar) -> np.dtype[A
         return dtype
 
 
-@dataclass(frozen=True, eq=True)
+@attrs.define(frozen=True)
 class NormalizedSlice:
     """
     A normalized version of :class:`slice`. "Normalized" is explained in
@@ -313,7 +318,7 @@ class NormalizedSlice:
     step: IntegralT
 
 
-@dataclass(eq=True, frozen=True)
+@attrs.define(frozen=True)
 class Axis(Taggable):
     """
     A type for recording the information about an :class:`~pytato.Array`'s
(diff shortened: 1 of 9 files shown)


inducer:
This generally seems worthwhile, I think.

>     * All derived classes must do `eq=False` in their call to `dataclass`, otherwise `dataclasses` will implement its own default which has a complexity of O(2N) in number of nodes in the DAG.
>     * All derived classes must do `def __hash__(self) -> int: return super().__hash__()`, otherwise `dataclasses` will implement its own default.

Is this something we could check for in `__post_init__` in `__debug__` mode?

kaushikcfd:
> Ah well. Maybe just disable unexpected-keyword-arg for pylint?

Yep, I will do that for now. Related pylint issue: https://github.com/PyCQA/pylint/issues/7623.

kaushikcfd:
The pylint errors have been ignored. But the current refactoring forces us to depend on `>=Py3.10` as the implementation depends on [kw_only](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) argument of `dataclasses.field` that was introduced in 3.10. Maybe we can merge this once we force the requirement to Py3.10?

kaushikcfd:
> Is this something we could check for in __post_init__ in __debug__ mode?

Yes, I think so. I will try to give it a shot.

inducer:
> The pylint errors have been ignored. But the current refactoring depends on [kw_only](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) argument of `dataclasses.field` that was introduced in 3.10. Maybe we can merge this once we force the requirement to Py3.10?

Is there a way to avoid that? Requiring 3.10 is a healthy ways down the road in my view, given that we're not even completely switched over to 3.8.

kaushikcfd:
> Is there a way to avoid that?

We could use [`attrs`](https://www.attrs.org/en/stable/examples.html) instead of `dataclasses`?

inducer:
> We could use [`attrs`](https://www.attrs.org/en/stable/examples.html) instead of `dataclasses`?

As long as it's easy enough to switch back to dataclasses, I would not be oppposed.

kaushikcfd:
> As long as it's easy enough to switch back to dataclasses, I would not be opposed.

Sure! It should take us an afternoon to switch to dataclasses from `attrs`.

setup.py
```diff
@@ -38,6 +38,7 @@
         "pytools>=2021.1",
         "pyrsistent",
         "immutables",
+        "attrs",
```
inducer, line 41: We should add a design decision in the docs explaining why we use `attrs` and not `dataclasses`.

pytato/array.py
```diff
@@ -428,16 +434,17 @@ class Array(Taggable):
...
     # hashable. Dicts of hashable keys and values are also permitted.
     _fields: ClassVar[Tuple[str, ...]] = ("axes", "tags",)
 
-    __array_priority__ = 1  # disallow numpy arithmetic to take precedence
-
-    def __init__(self, axes: AxesT, tags: FrozenSet[Tag]) -> None:
-        self.axes = axes
-        self.tags = tags
+    # disallow numpy arithmetic from taking precedence
+    __array_priority__: ClassVar[int] = 1
```
inducer, line 447: During review, we flagged a risk that subclasses might use `eq=True` or use the default hash by accident, and that we wanted to catch this during `__post_init__`. Somehow, this must have fallen through the cracks, but I think we still ought to do that.

setup.py
kaushikcfd: Moving this discussion to https://github.com/inducer/pytato/pull/371.

pytato/array.py
kaushikcfd: Moving this discussion to https://github.com/inducer/pytato/pull/371.

PR #366, opened 2022-10-18
Check that distributed DAG is deadlock-free

matthiasdiener:
After merge:
- [ ] https://github.com/inducer/grudge/pull/291

### DIFF
--- a/pytato/distributed.py
+++ b/pytato/distributed.py
@@ -349,7 +349,7 @@ class DistributedGraphPartition(GraphPartition):
     parts: Dict[PartId, DistributedGraphPart]
 
 
-def _map_distributed_graph_partion_nodes(
+def _map_distributed_graph_partition_nodes(
         map_array: Callable[[Array], Array],
         map_send: Callable[[DistributedSend], DistributedSend],
         gp: DistributedGraphPartition) -> DistributedGraphPartition:
@@ -1021,7 +1021,7 @@ def find_distributed_partition(outputs: DictOfNamedArrays
     def map_send(send: DistributedSend) -> DistributedSend:
         return send.copy(data=cmac(send.data))
 
-    return _map_distributed_graph_partion_nodes(map_array, map_send, gp)
+    return _map_distributed_graph_partition_nodes(map_array, map_send, gp)
 
 # }}}
 
--- a/pytato/distributed.py
+++ b/pytato/distributed.py
@@ -51,6 +51,10 @@ from pymbolic.mapper.optimize import optimize_mapper
 
 import numpy as np
 
+import logging
+logger = logging.getLogger(__name__)
+
+
 if TYPE_CHECKING:
     import mpi4py.MPI
 
@@ -83,6 +87,7 @@ nodes that are/are not a dependency of a receive or that feed/do not feed a send
 
 .. autoclass:: DistributedGraphPart
 .. autoclass:: DistributedGraphPartition
+.. autofunction:: verify_distributed_partition
 
 .. currentmodule:: pytato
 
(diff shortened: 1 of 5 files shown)


inducer:
> Is this going in the direction you had in mind @inducer?

Yes, but unless I'm mistaken, you're currently still missing the edges relating to inter-rank communication.

kaushikcfd:
I'm guessing the DAG being constructed here has nodes of the form `(part_id, rank_nr)` with edges of 3 types:
1. Inter-rank dependencies from the comm nodes matched by their tags
2. Intra-rank dependencies of the form P1->P2 iff P2 uses P1's output as an input.

pytato/distributed.py
```diff
@@ -1069,10 +1069,40 @@ def set_union(
...
     else:
         sym_tag_to_int_tag, next_tag = mpi_communicator.bcast(None, root=root_rank)
 
+    if __debug__:
+        all_parts_inputs = [(k, v.src_rank)
+                            for p in partition.parts.values()
+                            for k, v in p.input_name_to_recv_node.items()]
+        all_parts_outputs = [(k, v.dest_rank)
+                             for p in partition.parts.values()
+                             for k, v in p.output_name_to_send_node.items()]
```
inducer, line 1085: IIUC, this doesn't look at the tags at all. It should, in terms of uniqueness, and that sends and receives are matched exactly 1-to-1.

pytato/distributed.py
```diff
@@ -844,6 +849,191 @@ def _remove_part_id_tag(ary: ArrayOrNames) -> Array:
...
+                defining_pid = output_to_defining_pid[input_name]
+                assert defining_pid.rank == sumpart.pid.rank
+                add_needed_pid(sumpart.pid, defining_pid)
+
+        # Loop through all sends again, making sure there's exactly one recv.
+        for s in all_sends:
+            assert s in all_recvs, f"Missing recv: {s=} --- {all_recvs=}"
+
+        from pytools.graph import reverse_graph
+        pid_to_needing__pids = reverse_graph(pid_to_needed_pids)
```
kaushikcfd, line 1029: Why do we need to reverse here?
IIUC the whole point of assembling this graph is to check if it has any cycles and a directed graph has cycles if and only if its reverse graph has cycles.

inducer: That's a good point. Fixed. (Caution: force push.)

pytato/distributed.py
```diff
@@ -844,6 +849,188 @@ def _remove_part_id_tag(ary: ArrayOrNames) -> Array:
...
+        def add_needed_pid(pid: _DistributedPartId,
+                           needed_pid: _DistributedPartId) -> None:
+            pid_to_needed_pids.setdefault(pid, set()).add(needed_pid)
+
+        all_recvs: Set[_CommIdentifier] = set()
+        all_sends: Set[_CommIdentifier] = set()
+
+        print(all_summarized_parts)
+        output_to_defining_pid: Dict[_DistributedName, _DistributedPartId] = {}
+        for sumpart in all_summarized_parts.values():
```
matthiasdiener, line 966: For the keys in `part.input_name_to_recv_node`, there are no corresponding `output_names`, which is why this fails later on.

inducer: Ah, good catch. I believe that that's what we were missing. Do you want to push a fix?

matthiasdiener: I'm not sure what the right fix is - should we just ignore those input names in the later loop (around line 1023)?

pytato/distributed.py
matthiasdiener: Fixed in [446daf5](https://github.com/inducer/pytato/pull/366/commits/446daf5fb6ecf05bbfba6524233d7a1af12a3894)

kaushikcfd:
Is the global parts DAG verifier here noticing cycles in any of the distributed tests? I ask because it seems some of the CI runs are getting timed out by running into a deadlock -- https://github.com/inducer/pytato/actions/runs/3365558563/jobs/5583742094.

kaushikcfd:
Looks like [the CI here](https://github.com/inducer/pytato/actions/runs/3373139212/jobs/5597378204) is also experiencing the deadlock? I don't see any other reason why the tests would take this long.

matthiasdiener:
> Looks like [the CI in this PR](https://github.com/inducer/pytato/actions/runs/3373139212/jobs/5597378204) is also experiencing the deadlock? I don't see any other reason why the tests would take this long.

The tests run fine on my local machine. My guess is that the MPI runtime inside the CI trips over the fact that one rank exited with an exception and just hangs.

kaushikcfd:
> The tests run fine on my local machine. My guess is that the MPI runtime inside the CI trips over the fact that one rank exited with an exception and just hangs.

It seems to me moving to Py3.11 is triggering the error.

kaushikcfd:
Thanks! I ran [deadlock_checker](https://gist.github.com/kaushikcfd/2ed70af0502165abeb555f1e2b326705) with this branch and the user experience was not great. The driver reported the error -- `ValueError: duplicate send for comm id: '_CommIdentifier(source_rank=1, dest_rank=0, comm_tag=43)'` and deadlocked. The non-ideal bits were:
- I don't think there is a duplicate send on rank 1.
- The deadlocking bit is not great, it should gracefully terminate.

kaushikcfd:
(I'll remove myself as a reviewer since you guys are more knowledgeable about this than me.)

matthiasdiener:
> Thanks! I ran [deadlock_checker](https://gist.github.com/kaushikcfd/2ed70af0502165abeb555f1e2b326705) 

Thanks, this is really helpful! There was a bug in `_SummarizedDistributedSend` which is hopefully fixed now.

> with this branch and the user experience was _almost_ great. The driver reported the error -- `ValueError: duplicate send for comm id: '_CommIdentifier(source_rank=1, dest_rank=0, comm_tag=43)'` and deadlocked. The error was very helpful, but there were some non-ideal bits:
> 
> * I don't think there is a duplicate send from rank 1.

Right, with the fix there is a duplicate send from rank 0-> 1 now, which seems correct (the visualized DAG shows the same duplicate).

> * The deadlocking bit is not great, it should gracefully terminate.

Not sure what the best way to handle this is. It seems like its MPI runtime specific.

inducer:
I'm working on cleaning this up. In 

https://github.com/inducer/pytato/compare/794e1c79d1995cdee842368188cc4afd4143cb25..a8a72dcbfa68be5edf58515945e8ba11cb066703

I've dropped all the unrelated stuff.

test/test_distributed.py
```diff
@@ -312,6 +313,53 @@ def test_deterministic_partitioning():
...
+                dest_rank=(rank-1) % size, comm_tag=42, stapled_to=x)
+
+    outputs = pt.make_dict_of_named_arrays({"out": halo})
+    distributed_parts = find_distributed_partition(outputs)
+
+    if rank == 0:
+        with pytest.raises(ValueError):
+            verify_distributed_partition(comm, distributed_parts)
+    else:
+        verify_distributed_partition(comm, distributed_parts)
```
inducer, line 358: These tests don't cover many of the things the verifier guards against. While I'm tempted to merge this so that we can start using it, I feel like it might be wise to wait for coverage of at least the following scenarios:

- Duplicate send
- Duplicate recv
- Cycle in the graph

matthiasdiener: Thanks - I added these additional tests in [8d112fe](https://github.com/inducer/pytato/pull/366/commits/8d112fe0e626d8cafe084d761b006f1515fad63c)

inducer:
Thanks!

And thanks for making these extra commits, that made the changes easy to review. I've folded them into the verifier and test commits, fixed some typos, and added a base class and eliminated what I believe was a redundant test. (The two spots that raised `DuplicateSend` made that easy to spot.)

PR #368, opened 2022-10-19
Make AbstractResultWithNamedArrays taggable

kaushikcfd:
Metadata on AbstractResultsWithNamedArrays could come in handy. For eg. in #364, one might want to tag the `Call` with implementation tags (like InlineTag) as directives for the call-site.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -736,7 +736,8 @@ class NamedArray(Array):
         return self.expr.dtype
 
 
-class AbstractResultWithNamedArrays(Mapping[str, NamedArray], ABC):
+@attrs.define(frozen=True, eq=False)
+class AbstractResultWithNamedArrays(Mapping[str, NamedArray], Taggable, ABC):
     r"""An abstract array computation that results in multiple :class:`Array`\ s,
     each named. The way in which the values of these arrays are computed
     is determined by concrete subclasses of this class, e.g.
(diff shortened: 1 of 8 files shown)


pytato/array.py
```diff
@@ -751,7 +752,7 @@ class AbstractResultWithNamedArrays(Mapping[str, NamedArray], ABC):
 
         This container deliberately does not implement arithmetic.
     """
-
+    tags: FrozenSet[Tag] = attrs.field(kw_only=True, default=frozenset())
```
inducer, line 755: I wonder if it might be better to allow concrete implementations of the concept `AbstractResultWithNamedArrays` to be tagged, but to not force this on the overall concept. Interfaces that bring data with them kind of smell weird to me.

inducer, line 755: Sort of inconsistent that this defaults to `frozenset`, when elsewhere that is `get_default_tags`. Maybe this should not default?

kaushikcfd: Emits a Deprecation Warning in [469c339](https://github.com/inducer/pytato/pull/368/commits/469c339ffba38a61dca322f75cadfc3d0aa3fbfd). And `make_dict_of_named_arrays` uses `_get_default_tags`. Wdyt?

inducer: Works for me.

PR #369, opened 2022-10-19
Remove unnecessary function def. in PreprocessMapper.map_stack

kaushikcfd:


inducer:
Thx!

inducer:
To confirm my read of this: The previous codegen for `stack` was just broken, correct?

kaushikcfd:
No it wasn't broken. It just defined a function for no reason. The arguments were never used. So, in this patch we converted the function with no arguments it's return value.

inducer:
Ah nvm. I misread it. I confused the argument `array_index` with `i`. Thanks!

Issue #370, opened 2022-10-19
Post-review of transition to `attrs`


inducer:
After merging #365, I noticed a few things I wish I'd flagged, see https://github.com/inducer/pytato/pull/365#pullrequestreview-1147957251. They're not a huge deal, but I think we should address them to avoid setting traps for our future selves.

kaushikcfd:
From https://github.com/inducer/pytato/pull/371#discussion_r999778468, automaticall populate `_fields`.

matthiasdiener:
Out of curiosity, should we now consider moving back to `dataclasses` (which would imply Python 3.10+)?

inducer:
In general, I'm in favor of that. Good reasons to go back:

- uniformity? Everywhere else in the stack uses dataclasses.
- https://docs.python.org/3/library/typing.html#typing.dataclass_transform would allow us to define a single class decorator for array-likes, vs. the inconsistent mess that we have currently.

Obstacles:
- We're using `cache_hash=True` in attrs, for which dataclasses has no equivalent. We would need to generate our own code for `__hash__`. Could be worse, I think.

PR #371, opened 2022-10-19
Implement checks on new array types to prevent developer mishaps

kaushikcfd:
See the discussion in #365.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -446,6 +446,18 @@ class Array(Taggable):
     # disallow numpy arithmetic from taking precedence
     __array_priority__: ClassVar[int] = 1
 
+    def _is_eq_valid(self) -> bool:
+        return (self.__class__.__eq__ is Array.__eq__
+                and self.__class__.__hash__ is Array.__hash__)
+
+    def __post_init__(self) -> None:
+        # ensure that a developer does not uses dataclass' "__eq__"
+        # or "__hash__" implementation as they have exponential complexity.
+        assert self._is_eq_valid()
+
+    def __attrs_post_init__(self) -> None:
+        return self.__post_init__()
+
     def copy(self: ArrayT, **kwargs: Any) -> ArrayT:
         for field in self._fields:
             if field not in kwargs:
@@ -1471,6 +1483,7 @@ class Reshape(IndexRemappingBase):
     def __post_init__(self) -> None:
         # FIXME: Get rid of this restriction
         assert self.order == "C"
+        super().__post_init__()
 
     __attrs_post_init__ = __post_init__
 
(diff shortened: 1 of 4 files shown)


pytato/array.py
```diff
@@ -446,6 +446,15 @@ class Array(Taggable):
     # disallow numpy arithmetic from taking precedence
     __array_priority__: ClassVar[int] = 1
 
+    def __post_init__(self) -> None:
```
inducer, line 449: Maybe restrict these method definitions to `if __debug__`?

pytato/array.py
```diff
@@ -1674,6 +1684,9 @@ class DataWrapper(InputArgumentBase):
                                                           "shape")
```
inducer, line 1684: While I'm here: Might it be possible to get rid of `_fields` altogether, instead using something like [`attrs.fields`](https://www.attrs.org/en/stable/api.html#attrs.fields)?

pytato/array.py
```diff
@@ -1674,6 +1684,9 @@ class DataWrapper(InputArgumentBase):
...
     _mapper_method: ClassVar[str] = "map_data_wrapper"
 
+    def __post_init__(self) -> None:
+        pass
```
inducer, line 1688: - Add a comment to explain?
- This somehow feels quite broad, in that it assumes that the checks are the only thing happening in `__post_init__`? This might make more sense if the checks were in a sub-method which could then be selectively neutered? I'm not insisting on this, FWIW, merely raising the question.

test/test_pytato.py
```diff
@@ -941,6 +942,22 @@ def test_with_tagged_reduction():
             .tags_of_type(FooRednTag))
 
 
+def test_derived_class_uses_correct_array_eq():
+    @attrs.define(frozen=True)
+    class MyNewArrayT(pt.Array):
+        pass
+
+    if __debug__:
```
inducer, line 950: Pytest tests always run with `__debug__`.

inducer (commented):
Thanks! A few follow-ups below, otherwise LGTM!

pytato/array.py
```diff
@@ -446,6 +446,15 @@ class Array(Taggable):
     # disallow numpy arithmetic from taking precedence
     __array_priority__: ClassVar[int] = 1
 
+    def __post_init__(self) -> None:
+        # ensure that a developer does not uses dataclass' "__eq__"
+        # or "__hash__" implementation as they have exponential complexity.
+        assert self.__class__.__eq__ == Array.__eq__
+        assert self.__class__.__hash__ == Array.__hash__
```
inducer, line 453:
```suggestion
        assert self.__class__.__eq__ is Array.__eq__
        assert self.__class__.__hash__ is Array.__hash__
```
?

kaushikcfd: Thanks, done!

pytato/array.py
kaushikcfd: Noted this in #370.

pytato/array.py
kaushikcfd: Not loving it. At this point all our statements in these are asserts, so should be pretty lean.

pytato/array.py
kaushikcfd: Added the comment, thanks!

pytato/array.py
inducer: Fair enough.

inducer:
Thanks!

PR #376, opened 2022-11-02
Remove unused reverse_graph

matthiasdiener:
~~Needs https://github.com/inducer/pytools/pull/152~~


kaushikcfd:
There are no "real" users of `reverse_graph`:
```python
[line@line pytato]$ git grep 'reverse_graph('
pytato/transform/__init__.py:def reverse_graph(graph: Mapping[ArrayOrNames, FrozenSet[ArrayOrNames]]
test/test_pytato.py:    rev_graph = reverse_graph(uc.node_to_users)
test/test_pytato.py:    rev_graph2 = reverse_graph(reverse_graph(rev_graph))
test/test_pytato.py:    assert reverse_graph(rev_graph2) == uc.node_to_users
test/test_pytato.py:        rev_graph = reverse_graph(uc.node_to_users)
test/test_pytato.py:        rev_graph2 = reverse_graph(reverse_graph(rev_graph))
```
We can just get rid of it and get rid of referring to it in the regressions.

matthiasdiener:
> There are no "real" users of `reverse_graph`:
> 
> ```python
> [line@line pytato]$ git grep 'reverse_graph('
> pytato/transform/__init__.py:def reverse_graph(graph: Mapping[ArrayOrNames, FrozenSet[ArrayOrNames]]
> test/test_pytato.py:    rev_graph = reverse_graph(uc.node_to_users)
> test/test_pytato.py:    rev_graph2 = reverse_graph(reverse_graph(rev_graph))
> test/test_pytato.py:    assert reverse_graph(rev_graph2) == uc.node_to_users
> test/test_pytato.py:        rev_graph = reverse_graph(uc.node_to_users)
> test/test_pytato.py:        rev_graph2 = reverse_graph(reverse_graph(rev_graph))
> ```
> 
> We can just get rid of it and remove the references to it in the regressions.

I'm not opposed to removing, but we did use it in the previous partitioning algorithm IIRC. Not sure if this would make restoring that more difficult

kaushikcfd:
> Not sure if this would make restoring that more difficult

I don't think so. If we bring it back then we can simply call the `pytools` one.

test/test_pytato.py
```diff
@@ -324,18 +324,14 @@ def test_toposortmapper():
...
     array = pt.make_placeholder(name="array", shape=1, dtype=np.int64)
     y = array+1
 
     uc = UsersCollector()
     uc(y)
-    rev_graph = reverse_graph(uc.node_to_users)
-    rev_graph2 = reverse_graph(reverse_graph(rev_graph))
-
-    assert reverse_graph(rev_graph2) == uc.node_to_users
+    assert len(uc.node_to_users) == 2
```
kaushikcfd, line 334: I think this is testing for a very relaxed condition. Maybe make this condition tighter?

test/test_pytato.py
```diff
@@ -349,10 +345,7 @@ def test_userscollector():
         uc = UsersCollector()
         uc(dag)
 
-        rev_graph = reverse_graph(uc.node_to_users)
-        rev_graph2 = reverse_graph(reverse_graph(rev_graph))
-
-        assert rev_graph2 == rev_graph
+        assert len(uc.node_to_users) > 0
```
kaushikcfd, line 348: I think this is testing for a very relaxed condition. Maybe make this condition tighter?

test/test_pytato.py
matthiasdiener: What do you think of 3f5fe9e?

test/test_pytato.py
matthiasdiener: What do you think of https://github.com/inducer/pytato/commit/3f5fe9e3d66f75989759e3b5e244254687e95cea? - Strangely, that test fails currently...

test/test_pytato.py
inducer: Why not use `reverse_graph` from pytools for these?

matthiasdiener: We could, but then we would depend on a not-yet-released version of pytools for this test. If that is acceptable, I can re-add `reverse_graph` here (but imho, `reverse_graph` isn't super relevant for this test here).

test/test_pytato.py
matthiasdiener: Resolved in https://github.com/inducer/pytato/pull/376/commits/1d2c5a35d55060937870c1aa8830dfa849f6fc6a

test/test_pytato.py
```diff
@@ -324,23 +324,25 @@ def test_toposortmapper():
...
+    assert len(uc.node_to_users) == 2
+    assert uc.node_to_users[y] == set()
+    assert uc.node_to_users[array].pop() == y
+    assert len(uc.node_to_users[array]) == 0
 
     # Test random DAGs
     axis_len = 5
 
     for i in range(100):
+        print(i)
```
inducer, line 345: Stray debug print.

test/test_pytato.py
inducer: https://pypi.org/project/pytools/2022.1.13/

matthiasdiener: I re-added `reverse_graph` to the test in [f314a7d](https://github.com/inducer/pytato/pull/376/commits/f314a7daf78a2b81c4c84f4b27e5ee358842da9d).

test/test_pytato.py
matthiasdiener: We have this in a few locations in the tests as a progress indicator. I added a comment regarding this in [a0ece5f](https://github.com/inducer/pytato/pull/376/commits/a0ece5fc27f6cdd5b1d6d81f7befcfd25dbf8464). Would that be ok?

inducer:
Thanks!

PR #382, opened 2022-11-06
Fix return type annotation of `Axis._with_new_tag`

inducer:
Example CI failure: https://gitlab.tiker.net/inducer/pytato/-/jobs/482487


inducer:
Trivial, setting to merge.

PR #383, opened 2022-11-06
pyrsistent -> immutables

kaushikcfd:
Made this change to discourage different code patterns like `pmap(dict) | Map(dict)` and force `Map(dict)`. This should lead to more homogeneous code patterns. After this patch, we should see --

```
[line@line pytato]$ git grep pyrsistent | wc
      0       0       0
```


inducer:
LGTM, thanks!

PR #384, opened 2022-11-06
Lower to index lambda refactor

kaushikcfd:
Trie to implement the suggestions in https://github.com/inducer/pytato/pull/375#issuecomment-1304895412

### DIFF
--- a/pytato/codegen.py
+++ b/pytato/codegen.py
@@ -25,16 +25,11 @@ THE SOFTWARE.
 import dataclasses
 from typing import Union, Dict, Tuple, List, Any
 
-from pytato.array import (Array, DictOfNamedArrays,
-                          DataWrapper, Roll, AxisPermutation,
-                          Stack, Placeholder, Reshape,
-                          Concatenate, DataInterface, SizeParam,
-                          InputArgumentBase, Einsum,
-                          AdvancedIndexInContiguousAxes,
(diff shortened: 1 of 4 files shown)


kaushikcfd:
@inducer: I'm unable to make the MRO to pick the correct mapper methods. Other than that everything works here. Suggestions?

pytato/codegen.py
```diff
@@ -85,7 +80,7 @@ def _generate_name_for_temp(
 
 # {{{ preprocessing for codegen
 
-class CodeGenPreprocessor(CopyMapper):
+class CodeGenPreprocessor(CopyMapper, ToIndexLambdaMapper):
```
inducer, line 83: Shouldn't `ToIndexLambdaMapper` come first?

kaushikcfd: That would make `CodeGenPreprocessor.rec == ToIndexLambdaMapper.rec`.

kaushikcfd: Aah I guess another abstract class. :facepalm: My bad.

kaushikcfd: This has been fixed!

kaushikcfd:
This is a bit annoying to review as the <kbd>Tab</kbd> makes the `git-diff` very verbose. No updates were made to the mapper methods just moved pieces.

inducer:
LGTM, thanks!

PR #385, opened 2022-11-07
Minor distributed visualization improvements

matthiasdiener:
From #366.

**Before:**

image

---

**After:**

image


pytato/loopy.py
```diff
@@ -149,7 +149,8 @@ def copy(self, *,  # type: ignore[override]
                                axes=axes,
                                tags=tags)
 
-    def expr(self) -> Array:
+    # type-ignore-reason: incompatible with superclass
+    def expr(self) -> Array:  # type: ignore[override]
```
inducer, line 153: https://github.com/inducer/pytato/pull/387/commits/f79d2313b42b023831dc616df521415cde0cea30

pytato/visualization.py
```diff
@@ -482,7 +486,11 @@ def get_dot_graph_from_partition(partition: GraphPartition) -> str:
                     for name, send in (
                             part.output_name_to_send_node.items()):
                         node_id = id_gen("send")
-                        _emit_array(emit, "Send", {
+                        _emit_array(emit, "DistributedSend", {
+                            "addr": str(hex(id(send))),
+                            "shape": str(send.data.shape),
+                            "dtype": str(send.data.dtype),
+                            "tags": str(send.tags),
```
inducer, line 493: I think our vis nodes are way too big as it is. I could live with those things in the tooltip, but not in the node itself.

matthiasdiener: Ok, I have changed the amount of information to the prior one for `DistributedSend`/`DistributedRecv` in [0e86660](https://github.com/inducer/pytato/pull/385/commits/0e86660cd4d920aee8cab1f73ac31b1664bd4000), just updating their displayed names and adding the `comm_tag` to `DistributedSendRefHolder` for symmetry with `DistributedSend`. What do you think?

PR #277, opened 2022-03-20
Implements a distributed partitioner that is lazier than `find_distributed_partitions`

kaushikcfd:
- Only materialized nodes can be inputs/outputs to a partition
- Non-materialized nodes that are used in multiple partitions are duplicated to uphold the lazy behavior of the expressions
- Implementation includes a working `examples/distributed_partition_sends_asap.py` which demonstrates the output for the partitioner.

Draft because:
- [x] Cleanup
- [x] Add tests
- [x] Pass CI
- [x] Nomenclature incoherence: Users/Successors/Predecessors

PR #386, opened 2022-11-07
Restore older distributed partitioning algorithm as find_distributed_partition_v0

matthiasdiener:
The new one was introduced in https://github.com/inducer/pytato/pull/277.


inducer:
Superseded by #434 .

PR #387, opened 2022-11-07
Implement WalkMapperWithExtraArgs

kaushikcfd:
Best read commit-by-commit.

Draft because:
- [x] Includes commits from #384 as it performs some renames that avoid collision to allow the `optimize_mapper` to work here.

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -33,6 +33,7 @@ from pytato.array import (Array, IndexLambda, Stack, Concatenate, Einsum,
                           ShapeType)
 from pytato.transform import Mapper, ArrayOrNames, CachedWalkMapper
 from pytato.loopy import LoopyCall
+from pymbolic.mapper.optimize import optimize_mapper
 
 if TYPE_CHECKING:
     from pytato.distributed import DistributedRecv, DistributedSendRefHolder
@@ -352,6 +353,7 @@ class DirectPredecessorsGetter(Mapper):
 
 # {{{ NodeCountMapper
 
+@optimize_mapper(drop_args=True, drop_kwargs=True, inline_get_cache_key=True)
 class NodeCountMapper(CachedWalkMapper):
     """
     Counts the number of nodes in a DAG.
@@ -365,7 +367,12 @@ class NodeCountMapper(CachedWalkMapper):
         super().__init__()
         self.count = 0
 
-    def post_visit(self, expr: Any) -> None:
+    # type-ignore-reason: dropped the extra `*args, **kwargs`.
+    def get_cache_key(self, expr: ArrayOrNames) -> int:  # type: ignore[override]
+        return id(expr)
+
+    # type-ignore-reason: dropped the extra `*args, **kwargs`.
+    def post_visit(self, expr: Any) -> None:  # type: ignore[override]
         self.count += 1
 
 
--- a/pytato/partition.py
+++ b/pytato/partition.py
@@ -38,6 +38,7 @@ from pytato.array import (
         DictOfNamedArrays, make_placeholder, make_dict_of_named_arrays)
 
 from pytato.target import BoundProgram
+from pymbolic.mapper.optimize import optimize_mapper
 
 
 __doc__ = """
@@ -372,12 +373,18 @@ def find_partition(outputs: DictOfNamedArrays,
     return result
 
 
+@optimize_mapper(drop_args=True, drop_kwargs=True, inline_get_cache_key=True)
 class _SeenNodesWalkMapper(CachedWalkMapper):
     def __init__(self) -> None:
         super().__init__()
         self.seen_nodes: Set[ArrayOrNames] = set()
 
-    def visit(self, expr: ArrayOrNames) -> bool:
+    # type-ignore-reason: dropped the extra `*args, **kwargs`.
+    def get_cache_key(self, expr: ArrayOrNames) -> int:  # type: ignore[override]
+        return id(expr)
+
+    # type-ignore-reason: dropped the extra `*args, **kwargs`.
+    def visit(self, expr: ArrayOrNames) -> bool:  # type: ignore[override]
         super().visit(expr)
         self.seen_nodes.add(expr)
         return True
--- a/pytato/transform/__init__.py
+++ b/pytato/transform/__init__.py
@@ -50,7 +50,7 @@ if TYPE_CHECKING:
     from pytato.distributed import DistributedSendRefHolder, DistributedRecv
 
 ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
-T = TypeVar("T", bound=ArrayOrNames)
+MappedT = TypeVar("MappedT", bound=ArrayOrNames)
 CombineT = TypeVar("CombineT")  # used in CombineMapper
 CachedMapperT = TypeVar("CachedMapperT")  # used in CachedMapper
 IndexOrShapeExpr = TypeVar("IndexOrShapeExpr")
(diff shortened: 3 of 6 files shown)


pytato/transform/__init__.py
```diff
@@ -951,6 +953,168 @@ def map_loopy_call(self, expr: LoopyCall) -> None:
 # }}}
 
 
+# {{{ WalkMapperWithExtraArgs
+
+class WalkMapperWithExtraArgs(Mapper):
```
inducer, line 958: Instead of having with-args and without-args variants of each mapper, what I'd prefer is to have args on all mappers by default and use [`optimize_mapper`](https://github.com/inducer/pymbolic/blob/36af97d97ab944b7a049ef83843152111db0ae22/pymbolic/mapper/optimize.py#L233-L241) to specialize them away when needed. I wrote that specifically so it would/should work unchanged on pytato mappers.

kaushikcfd: Thanks, that was very convenient. Done!

pytato/transform/__init__.py
```diff
@@ -333,7 +336,7 @@ def map_reshape(self, expr: Reshape) -> Array:
                        tags=expr.tags)
 
     def map_distributed_send_ref_holder(
-            self, expr: DistributedSendRefHolder) -> Array:
+            self, expr: "DistributedSendRefHolder") -> Array:
```
inducer, line 339: Are the quotes necessary/helpful given that we have
```
from __future__ import annotations
```
?

pytato/transform/__init__.py
```diff
@@ -788,165 +791,196 @@ def map_size_param(self, expr: SizeParam) -> FrozenSet[SizeParam]:
...
-        self.post_visit(expr)
+        self.post_visit(expr, *args, **kwargs)
+
+# }}}
+
+
+# {{{ WalkMapper
+
+@optimize_mapper(drop_args=True, drop_kwargs=True)
+class WalkMapper(WalkMapperWithExtraArgs):
```
inducer, line 960: I figured we would not have classes like this at all, and instead apply `@optimize_mapper` to each "leaf" subclass as needed. This would leave the original `WalkMapper` having arguments, which I think is OK from the point of view of backward-compatibility.

inducer (commented):
Thanks! Some minor questions, otherwise this is headed in the right direction.

pytato/transform/__init__.py
kaushikcfd: In the earlier version where we had an explicit `WalkMapperWithExtraArgs` it was needed. In the newer version, it is no longer needed.

pytato/transform/__init__.py
kaushikcfd: Yep, that's more elegant. Fixed!

pytato/transform/__init__.py
kaushikcfd: Oh no, I was wrong it is needed in the newer version as well:
```python
../pytato/transform/__init__.py:989: in <module>
    @optimize_mapper(drop_args=True, drop_kwargs=True)
../../pymbolic/pymbolic/mapper/optimize.py:352: in wrapper
    exec(compile(
<'/home/line/projects/pytato_env/src/pytato/pytato/transform/__init__.py' modified by optimize_mapper>:1: in <module>
    ???
<'/home/line/projects/pytato_env/src/pytato/pytato/transform/__init__.py' modified by optimize_mapper>:66: in TopoSortMapper
    ???
E   NameError: name 'DistributedRecv' is not defined
```

Reverting back to quotes.

inducer: :facepalm: That's a bug in `optimize_mapper`! Hold your horses. :)

inducer: https://github.com/inducer/pymbolic/pull/117

inducer: Expedited the merge on that so that you can start using it. :)

kaushikcfd: The namespace collision checker is now unhappy for some common code patterns, like:
```
test_pytato.py:34: in <module>
    import pytato as pt
../pytato/__init__.py:90: in <module>
    from pytato.visualization import (get_dot_graph, show_dot_graph,
../pytato/visualization.py:49: in <module>
    from pytato.partition import GraphPartition
../pytato/partition.py:376: in <module>
    @optimize_mapper(drop_args=True, drop_kwargs=True)
../../pymbolic/pymbolic/mapper/optimize.py:350: in wrapper
    raise ValueError(
E   ValueError: symbol disagreement in environment: 'logger', most recently from 'pytato.partition'
```

inducer: Hmm. My inclination would be to go with renaming those to avoid the conflicts. But we *could* also fix this on the level of the AST transform. What do you think?

kaushikcfd: Sure, that's not a big inconvenience.

inducer: :+1: My thought with `optimize_mapper` was that it's OK for it to have *some* sharp edges (like this), as long as failures would be obvious. It *is* a pretty invasive thing that's being done after all.

kaushikcfd: After https://github.com/inducer/pymbolic/pull/118 this works.

kaushikcfd:
I'll handle the conflicts here.

pytato/transform/__init__.py
```diff
@@ -788,7 +790,7 @@ def map_size_param(self, expr: SizeParam) -> FrozenSet[SizeParam]:
 # }}}
 
 
-# {{{ WalkMapper
+# {{{ WalkMapperWithExtraArgs
```
inducer, line 793:
```suggestion
# {{{ WalkMapper
```

pytato/transform/__init__.py
```diff
@@ -45,12 +45,13 @@
...
 from pytato.tags import ImplStored
 from immutables import Map
+from pymbolic.mapper.optimize import optimize_mapper
 
 if TYPE_CHECKING:
     from pytato.distributed import DistributedSendRefHolder, DistributedRecv
 
 ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
-T = TypeVar("T", bound=ArrayOrNames)
+MapperT = TypeVar("MapperT", bound=ArrayOrNames)
```
inducer, line 54:
```suggestion
MappedT = TypeVar("MappedT", bound=ArrayOrNames)
```

inducer (commented):
Thanks! A few (hopefully small) wrinkles, then this is good to go.

pytato/analysis/__init__.py
```diff
@@ -365,7 +365,12 @@ def __init__(self) -> None:
         super().__init__()
         self.count = 0
 
-    def post_visit(self, expr: Any) -> None:
+    # type-ignore-reason: dropped the extra `*args, **kwargs`.
```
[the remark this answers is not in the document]
kaushikcfd: Yep, that slipped through. Fixed now. Thanks!

pytato/transform/__init__.py
kaushikcfd: Thanks, done!

pytato/transform/__init__.py
kaushikcfd: Thanks, done

pytato/partition.py
```diff
@@ -372,12 +373,18 @@ def find_partition(outputs: DictOfNamedArrays,
     return result
 
 
+@optimize_mapper(drop_args=True, drop_kwargs=True)
```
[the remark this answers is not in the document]
inducer: Ah, ok.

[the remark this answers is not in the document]
inducer: But wait: There's `inline_cache` and `inline_get_cache_key`. I'm talking about the latter. It seems you're talking about the former.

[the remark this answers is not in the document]
kaushikcfd: > I'm talking about the latter. It seems you're talking about the former.

Aah yep.

Fixed now.

inducer (approved):
Thanks!

inducer:
I wish there was a cleaner way to tell the type checker about the dropped `*args` and `**kwargs`, but :shrug: I can't think of anything.

PR #389, opened 2022-11-14
Use `-m mpi4py` in MPI test invocations

inducer:


kaushikcfd:
I don't think we advertise `python -m mpi4py ...` to be the only way to run pytato DAGs with comm nodes in them. I was thinking maybe the manager rank broadcasts to every other rank in the communicator whether the DAG was sound or not.

kaushikcfd:
FWIW, this has been my user experience -- https://asciinema.org/a/IM3GNhDKuTBVcK8jz1avZ77L9.

inducer:
> I don't think we advertise `python -m mpi4py ...` to be the only way to run pytato DAGs with comm nodes in them. I was thinking maybe the manager rank broadcasts to every other rank in the communicator whether the DAG was sound or not.

I agree that, as user experience goes, what you experienced isn't fantastic. That said, this hanging behavior is not unique to this particular exception. Any exception that occurs on any rank in an MPI+Python program will lead to a hang of this nature, and we can't possibly equip every exception with broadcasting logic of the type you suggest. As such, I would say that it's on the user to know that `-m mpi4py` is needed in order to avoid hangs in case of error.

Issue #378, opened 2022-11-02
DistributedRecv reachable from multiple parts results in multiple receives for the same tag/rank pair


kaushikcfd:
Consider the MWE:
```python
import pytato as pt


size = 4
rank = 2
send_rank = 3
recv_rank = 1


x = pt.make_placeholder("x", 10, "float64")
recv = pt.make_distributed_recv(
            src_rank=recv_rank, comm_tag=42,
            shape=x.shape, dtype=x.dtype)
y = x + recv

send1 = pt.staple_distributed_send(
       x, dest_rank=send_rank, comm_tag=43,
       stapled_to=y)

send2 = pt.staple_distributed_send(
    send1 + recv, dest_rank=send_rank, comm_tag=44,
    stapled_to=send1)

out = pt.make_dict_of_named_arrays({"out": send1 + send2})

parts = pt.find_distributed_partition(out)
pt.show_dot_graph(parts)
```

Notice how there is only one receive-node, but the partition is emitted as --
![code](https://user-images.githubusercontent.com/15399010/199566512-366973be-5524-4ac0-a413-6b2aca724fc1.svg)

This is definitely a bug in `pt.find_distributed_partition`'s `_PartIdTagAssigner` which introduces another receive-node.

I'm not too sure about the implementation in `execute_partition`, but I guess this could lead to deadlocks?

kaushikcfd:
Stumbled onto this while debugging #374. I'm not currently working on a fix for this, at least until mid-December.

matthiasdiener:
There is also a duplicated Send for (dest_rank=3, comm_tag=43).

kaushikcfd:
> There is also a duplicated Send for (dest_rank=3, comm_tag=43).

Yep, it's botched all the way through.

matthiasdiener:
Here is the graph of `partitioned_outputs` just after running the `_PartIDTagAssigner`:

<img width="803" alt="image" src="https://user-images.githubusercontent.com/1772435/199880351-ac5a779a-642d-4073-9208-5b1c7ba02fcc.png">

matthiasdiener:
Interestingly, the "older" distributed partitioning algorithm (https://github.com/inducer/pytato/pull/386) seems to have the same issue for this MWE.

PR #391, opened 2022-11-18
Add distributed prepartitioning check

matthiasdiener:

Issue #272, opened 2022-03-17
[distributed] InputArgumentBase should not become "extra placeholders"


kaushikcfd:
The following program:
```python
import pytato as pt
import numpy as np
from pytato.distributed import (staple_distributed_send, make_distributed_recv,
                                find_distributed_partition)

from pytato.partition import generate_code_for_partition

rank = 0
size = 2

img1 = pt.make_placeholder("img1", shape=(100, 3), dtype="float32")
scl_np = np.random.rand(3)
scl = pt.make_data_wrapper(scl_np, name="scl")

y1 = img1 * scl

img2 = staple_distributed_send(
    y1, dest_rank=(rank-1) % size,
    comm_tag=42,
    stapled_to=make_distributed_recv(shape=img1.shape, dtype=img1.dtype,
                                     src_rank=(rank+1) % size, comm_tag=42,
                                     axes=img1.axes))
out = img1 + img2*scl


distributed_parts = find_distributed_partition(
    pt.make_dict_of_named_arrays({"out": out}))
prg_per_partition = generate_code_for_partition(distributed_parts)

for _, prg in prg_per_partition.items():
    print(prg.program)
```

performs unnecessary copies  (redundant copies have been annotated in the kernel)

```python
---------------------------------------------------------------------------
KERNEL: _pt_kernel
---------------------------------------------------------------------------
ARGUMENTS:
_pt_dist_id_0: type: np:dtype('float64'), shape: (100, 3), dim_tags: (N1:stride:3, N0:stride:1) aspace: global
_pt_part_ph_id: type: np:dtype('float32'), shape: (100, 3), dim_tags: (N1:stride:3, N0:stride:1) aspace: global
_pt_part_ph_id_0: type: np:dtype('float64'), shape: (3), dim_tags: (N0:stride:1) aspace: global
img1: type: np:dtype('float32'), shape: (100, 3), dim_tags: (N1:stride:3, N0:stride:1), offset: <class 'loopy.kernel.data.auto'> aspace: global
scl: type: np:dtype('float64'), shape: (3), dim_tags: (N0:stride:1), offset: <class 'loopy.kernel.data.auto'> aspace: global
---------------------------------------------------------------------------
DOMAINS:
{ [_pt_part_ph_id_dim0, _pt_part_ph_id_dim1] : 0 <= _pt_part_ph_id_dim0 <= 99 and 0 <= _pt_part_ph_id_dim1 <= 2 }
[...]

PR #393, opened 2022-11-22
New distributed-memory DAG partitioner

inducer:
This has the beginnings of a new algorithm for the DAG partitioner that provably (I think!) will not artificially introduce cycles into the partitioned DAG.

Draft because:
- [ ]  It is not possible to implement a guaranteed cycle-free `find_distributed_partition` without communication. As a result, the interface to `find_distributed_partition` has to change in a backwards-incompatible way, so that we can get an MPI communicator. So this needs to merge at the same time as corresponding downstream PRs. `find_distributed_partition` is now also MPI-collective.
    - https://github.com/inducer/grudge/pull/298/files
- [x] The implementation is unfinished and largely untested. It implements the four steps in the docstring, so that it knows the graph parts at the level of the communication delimiting them. Reinserting the computation into the batches still remains to be done.
- [x] Make sure the sends and receives are exactly what's given in the communication batches.
- [x] ~~Maybe split off the initial, communication-needing part of `find_distributed_partition` from the remainder of the routine.~~
- [x] Add a test made from @kaushikcfd's MWE: https://gist.github.com/kaushikcfd/efc4661814f46924b75194dfec177246
- [x] Rework `test_deterministic_partitioning`: it no longer makes sense to run it in the absence of MPI.
- [x] ~~Add a test based on #378.~~
    - It's structurally not possible for sends and receives to get duplicated, so I think we don't need this.
- [ ] @majosm observed that running `examples/distributed.py` with three ranks caused cycles at a point. Add tests based on it.
- [x] Catch when the breadth-first batch-maker fails to make progress, indicating a cycle.
- [ ] Update the description of the algorithm in the docstring.

Notes:
- There's a description of the proposed algorithm in the docstring of the new `find_distributed_partition`. That may be nicer to look at rendered in sphinx.
- This generously pilfers ideas and some code from @kaushikcfd's approach in #277.

cc @majosm @matthiasdiener @kaushikcfd 

Closes: #378.


kaushikcfd:
>  It is not possible to implement a guaranteed cycle-free find_distributed_partition without communication. As a result, the interface to find_distributed_partition has to change in a backwards-incompatible way, so that we can get an MPI communicator. So this needs to merge at the same time as corresponding downstream PRs. find_distributed_partition is now also MPI-collective.

Can you demonstrate with an example where the global communication DAG is sound (i.e. has no dependency cycles) but our generated partition is not? I understand there is #378, but that seems like a minor implementation bug.

inducer:
I mean... the burden is really the other way around. Algorithms are suspect until proven correct, not the other way around. :slightly_smiling_face: I did not see a way to prove the existing one correct.




- #378 is not the background to this; I agree that that's independently fixable and not a big deal.
- There is also https://github.com/illinois-ceesd/mirgecom/issues/783#issuecomment-1319386960, but we don't have firm evidence on whether the input graph is OK.

inducer:
OK, did not mean to either close this *or* send this comment just yet. But anyway. Constructing a concrete counterexample (other than the potential one in mirgecom) is a task in itself, that's why tools like TLA+ exist. :)

I have a hunch that getting the "clocks" on two ranks misaligned would yield something, using a graph like the following:
grafik

You need three or more ranks in order to be able to do that, which is somewhat consistent with when the problems in mirgecom started appearing.

kaushikcfd:
> Algorithms are suspect until proven correct, not the other way around. slightly_smiling_face I did not see a way to prove the existing one correct.

That's completely fair. But, before we propose an improvement we need an MWE for which the old implementation leads to undesired partitions. The previous rewrite effort did provide MWEs to justify itself via https://github.com/inducer/grudge/issues/237, https://github.com/inducer/pytato/issues/272.

> I have a hunch that getting the "clocks" on two ranks misaligned would yield something, using a graph like the following:

Getting an MWE for that should be fairly easy. FWIW, it's not clear to me that it should result in cyclic partition. I just don't want us to spend effort looking for a solution for an unknown problem.

inducer:
> Getting an MWE for that should be fairly easy. FWIW, it's not clear to me that it should result in cyclic partition. I just don't want us to spend effort looking for a solution for an unknown problem.

I see your point, and in the absence of this cycle thing being a show-stopper for much of CEESD, this would not have happened! Potentially @matthiasdiener (via #391) will supply evidence that the mirgecom-generated graph is OK. In the meantime, I'll also think about a counterexample for the existing verifier. But, perhaps understandably, my first priority is to get CEESD un-stuck.

inducer:
I think I've got an example that shows why information on just the local rank won't suffice. Consider this distributed DAG:

grafik

The point is that the "left-going" and the "right-going" communication will look identical to rank 1. I do not think it has a way of knowing that those two can't be batched together without creating a cycle, or that the right-going must come before the left-going.

kaushikcfd:
Here's a reproducer for the hypothesized comm pattern: https://gist.github.com/kaushikcfd/efc4661814f46924b75194dfec177246. It does emit a `PartitionInducedCycleError`.

pytato/distributed/partition.py
```diff
@@ -654,53 +855,158 @@ def find_distributed_partition(outputs: DictOfNamedArrays
...
+    assert all(
+            (
+                materialized_ary_to_part_id_start[ary]
+                < materialized_ary_to_part_id_end[ary])
+            for ary in materialized_arrays), \
+        "unable to find suitable part for materialized array"
+
+    materialized_ary_to_part_id: Dict[Array, int] = {
+            ary: materialized_ary_to_part_id_end[ary] - 1
+            for ary in materialized_arrays}
```
inducer, line 985: @majosm IIUC, this represents a policy of computing things "as late as possible", whereas I'd personally maybe go with computing things "as early as possible". What made you go with late over early? (We should probably add a comment that this is a choice either way.)

pytato/distributed/partition.py
```diff
@@ -654,53 +855,158 @@ def find_distributed_partition(outputs: DictOfNamedArrays
...
+                comm_id_to_part_id[
+                    _send_to_comm_id(local_rank,
+                                     sent_array_to_send_node[sent_ary])] + 1)
+
+    assert all(
+            (
+                materialized_ary_to_part_id_start[ary]
+                < materialized_ary_to_part_id_end[ary])
+            for ary in materialized_arrays), \
+        "unable to find suitable part for materialized array"
```
inducer, line 981: On the assumption of non-empty ranges, we could conceivably stick one of "sides" behind an `if __debug__`, i.e. I don't think we strictly really need both.

pytato/distributed/partition.py
majosm: I didn't have any particular reason for choosing that way over the other, just had to pick one. Maybe one way to retroactively justify that choice is that it would keep parts from holding up the show by doing unrelated work before firing off their sends? 🤷‍♂️

pytato/distributed/partition.py
```diff
@@ -654,118 +572,221 @@ def find_distributed_partition(outputs: DictOfNamedArrays
...
+    dict_union_mpi_op = MPI.Op.Create(
+            # type ignore reason: mpi4py misdeclares op functions as returning
+            # None.
+            _dict_union_mpi,  # type: ignore[arg-type]
+            commute=True)
+    try:
+        comm_to_needed_comms = mpi_communicator.allreduce(
+                local_send_id_to_needed_local_recv_ids, dict_union_mpi_op)
+    finally:
+        dict_union_mpi_op.Free()
```
inducer, line 596: Talking with @matthiasdiener, it occurred to me that this will not include receives that directly feed an output. Those will not get assigned to a batch, which I think is an issue.

majosm: I might be misunderstanding, but I think as written `comm_to_needed_comms` will pick up receives that are also outputs (Kaushik's hang reproducer has one of these). There will be a send on some rank matching up with that receive, so the comm ID ends up as a key in `local_send_id_to_needed_local_recv_ids` on that rank. And then it gets `allreduce`d onto every rank.

inducer: You're right, they will show up. Nvm!

inducer:
The deterministic partitioning test is back, but unfortunately it fails at the moment. See https://github.com/inducer/pytato/pull/393/commits/c8db611f0d926787d10332d446a303d54835f08f and https://github.com/inducer/pymbolic/pull/125 for my current, somewhat radical approach to this.

inducer:
Superseded by #434.

PR #397, opened 2022-11-28
Misc. fixes for new distributed partitioning

majosm:
Targets #393.

### DIFF
--- a/pytato/distributed/partition.py
+++ b/pytato/distributed/partition.py
@@ -923,6 +923,8 @@ def find_distributed_partition(
     comm_batches, comm_id_to_part_id = _make_local_comm_batches(
             local_rank, local_comm_to_needed_local_comms)
 
+    # FIXME?: comm_batches isn't being used for anything
+
     # }}}
 
     # {{{ assign each materialized array to a batch/part
--- a/pytato/distributed/partition.py
+++ b/pytato/distributed/partition.py
@@ -925,6 +925,8 @@ def find_distributed_partition(
 
     # FIXME?: comm_batches isn't being used for anything
 
+    nparts = max(part_id for part_id in comm_id_to_part_id.values()) + 1
+
     # }}}
 
     # {{{ assign each materialized array to a batch/part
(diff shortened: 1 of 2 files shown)


pytato/distributed/partition.py
```diff
@@ -950,15 +954,17 @@ def get_needed_local_sends(
     materialized_ary_to_part_id: Dict[Array, int] = {
             ary: max(
                 max(
-                    comm_id_to_part_id[
+                    (comm_id_to_part_id[
                         _send_to_comm_id(local_rank,
                                          sent_array_to_send_node[sent_array])]
                     for sent_array in sent_array_dep_mapper(ary)),
+                    default=nparts-1),
```
inducer, line 961: Add comment to explain the default.

pytato/distributed/partition.py
```diff
@@ -950,15 +954,17 @@ def get_needed_local_sends(
     materialized_ary_to_part_id: Dict[Array, int] = {
```
inducer, line 954: This is totally wrong. The right thing to do use here is "two overlapping ranges":
- A receive says: the array must be in part number $\\\{\text{receving part nr}, \dots, \text{nparts}\\\}$.
- A send says: the array must be in part number $\\\{0, \dots, \text{send part nr}\\\}$.

majosm: Something like https://github.com/inducer/pytato/pull/397/commits/ee18e54c5cfa9e7bcc5cac4a5fff7dbfece8f3b0?

majosm: Wait, I think it's still wrong. Seems like we need to reverse the dependency lookup for the sends? (Since we want to find the sends that depend on the materialized array, not the other way around...)

majosm: https://github.com/inducer/pytato/pull/397/commits/b80875f6f491bd6f973f97423fe24e6162793e1e?

inducer: I buy that.

inducer:
> (@inducer Didn't you add multiple comments during the meeting? I only see one of them here.)

I did, but I deleted them when they became moot when we discovered that things were even wrong-er.

inducer:
Thanks! I've read through these changes, and they make sense to me.  Would you be opposed to me squashing those into the existing commits in #393?

majosm:
> Thanks! I've read through these changes, and they make sense to me. Would you be opposed to me squashing those into the existing commits in #393?

Sure, no objections.

PR #399, opened 2022-12-01
[trivial] add comment regarding _DistributedCommReplacer.map_distributed_send

matthiasdiener:


pytato/distributed/partition.py
```diff
@@ -150,6 +150,8 @@ def map_distributed_send_ref_holder(
         raise ValueError("DistributedSendRefHolder should not occur in partitioned "
                 "graphs")
 
+    # Note: map_distributed_send() is not called recursively, since DistributedSend
+    # is not an Array. It must be called explicitly.
```
inducer, line 154: I agree with the conclusion, but I do not totally agree with the reasoning. Rather, at the point where this mapper is used, the `SendRefHolder`s have been removed from the DAG, and therefore, there are no more references to the `Send`s from within the DAG. Therefore, `map_distributed_send` needs to be called separately from the DAG traversal.

matthiasdiener: I think I understand what you mean, and I added your explanation to the comment. My initial motivation for the comment was that `map_distributed_send` looks like a normal mapper method, but since DistributedSend has no `_mapper_method`, `map_distributed_send` should not be automatically called in any DAG traversal, as far as I understand.

pytato/distributed/partition.py
```diff
@@ -150,6 +147,12 @@ def map_distributed_send_ref_holder(
         raise ValueError("DistributedSendRefHolder should not occur in partitioned "
                 "graphs")
 
+    # Note: map_distributed_send() is not called like other mapped methods in a
+    # DAG traversal, since a DistributedSend is not an Array and has no
+    # '_mapper_method' field. Furthermore, at the point where this mapper is used,
```
inducer, line 152: I was wondering whether maybe it *should* have that method.

pytato/distributed/partition.py
```diff
@@ -124,9 +124,8 @@ class _DistributedCommReplacer(CopyMapper):
         so that received data can be externally supplied, making a note
         in :attr:`input_name_to_recv_node`.
 
-    -   Eliminates :class:`DistributedSendRefHolder` and
-        :class:`DistributedSend` from the DAG, making a note of data
-        to be send in :attr:`output_name_to_send_node`.
+    -   Makes note of data to be sent from :class:`DistributedSend` nodes
```
inducer, line 127: Why drop the bit about `DistributedSendRefHolder`?

matthiasdiener: Doesn't this mapper expect to be called with DistributedSendRefHolders already eliminated?

pytato/distributed/partition.py
matthiasdiener: I think having such a method might make code like https://github.com/inducer/pytato/blob/bc28de872183bc63f6e638fb775964d21e83f424/pytato/transform/__init__.py#L1446-L1451 a bit clearer...

pytato/distributed/partition.py
inducer: Ah, you're right. That docstring was stale.

PR #400, opened 2022-12-02
DistributedSend: add shape, dtype

matthiasdiener:


inducer:
Those are kind of deliberately not there, to avoid confusion on whether `Send` has an (array) value.

PR #402, opened 2022-12-10
distributed example: make it run with arbitrary rank numbers

matthiasdiener:
cc @majosm

### DIFF
--- a/examples/distributed.py
+++ b/examples/distributed.py
@@ -52,7 +52,10 @@ def main():
 
     final_res = context["out"].get(queue)
 
-    ref_res = comm.bcast(final_res)
+    comm.isend(x_in, dest=(rank-1) % size, tag=42)
+    ref_halo = comm.recv(source=(rank+1) % size, tag=42)
+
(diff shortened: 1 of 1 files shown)


examples/distributed.py
```diff
@@ -52,7 +52,10 @@ def main():
 
     final_res = context["out"].get(queue)
 
-    ref_res = comm.bcast(final_res)
+    comm.isend(x_in, dest=(rank-1) % size, tag=42)
+    ref_halo = comm.recv(source=(rank+1) % size, tag=42)
+
+    ref_res = x_in + ref_halo
```
inducer, line 58: Not sure I get it. Why does this work on arbitrary rank numbers, but the broadcast doesn't?

majosm: The broadcast only works if all the ranks operate on the same two sets of random numbers as rank 0 (which is true by coincidence for `n == 2`, but not for `n > 2`).

inducer: OIC, I get it now. Thanks!

PR #404, opened 2022-12-13
Add distributed prepartitioning check based on _LocalSendRecvDepGatherer

matthiasdiener:
Alternative to #391.
Based on #393.

### DIFF
--- a/examples/distributed.py
+++ b/examples/distributed.py
@@ -30,6 +30,9 @@ def main():
 
     # Find the partition
     outputs = pt.DictOfNamedArrays({"out": y})
+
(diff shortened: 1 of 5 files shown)


inducer:
I don't think this is helpful, sorry. It dilutes commit authorship, moves code to places where it shouldn't be, imports private symbols across module boundaries, and I think re-merging this into #393 will be more trouble than me redoing this from scratch.

inducer:
As it stands, these checks are now done as part of partitioning, see #404.

inducer:
Superseded by #434 .

PR #406, opened 2022-12-25
Swap bool8 -> bool_ to placate numpy deprecation

inducer:


inducer:
Force-merging because the downstream failure is unrelated.

PR #417, opened 2023-02-01
Make `get_dot_graph_from_partition` work with duplicated computation

majosm:
Currently each array is given a single name in the emitted dot code, even if it is computed by multiple parts. This can lead to strangeness like the duplicated edges entering the `_in0[_0, _1] + 1` node in the graph below:

before_fix

This PR tweaks the code to make sure that each node to be emitted gets assigned a unique name. With the changes, the above graph becomes:

after_fix


inducer:
Thanks!

PR #423, opened 2023-03-28
Enable isort

matthiasdiener:


inducer:
We should probably wait with this until we have most pending PRs merged, as this will likely cause conflicts in all of them.

inducer:
Superseded by #527.

PR #427, opened 2023-04-19
Make the parametric type of CachedMapper to be the return type

kaushikcfd:
Before this patch, the parametric type in CachedMapper denoted the input type of the mapper which is unnecessary as the input type is always `ArrayOrNames`.

I think https://github.com/inducer/pytato/issues/409 is also talking about this same thing.

### DIFF
--- a/pytato/transform/__init__.py
+++ b/pytato/transform/__init__.py
@@ -49,8 +49,11 @@ from pytato.tags import ImplStored
 from pymbolic.mapper.optimize import optimize_mapper
 
 ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
-MappedT = TypeVar("MappedT", bound=ArrayOrNames)
+MappedT = TypeVar("MappedT",
+                  Array, AbstractResultWithNamedArrays, ArrayOrNames)
 CombineT = TypeVar("CombineT")  # used in CombineMapper
+CopyMapperResultT = TypeVar("CopyMapperResultT",  # used in CopyMapper
+                            Array, AbstractResultWithNamedArrays, ArrayOrNames)
 CachedMapperT = TypeVar("CachedMapperT")  # used in CachedMapper
 IndexOrShapeExpr = TypeVar("IndexOrShapeExpr")
 R = FrozenSet[Array]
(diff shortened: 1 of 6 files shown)


pytato/transform/__init__.py
```diff
@@ -49,8 +49,11 @@
 from pymbolic.mapper.optimize import optimize_mapper
 
 ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
-MappedT = TypeVar("MappedT", bound=ArrayOrNames)
+MappedT = TypeVar("MappedT",
+                  Array, AbstractResultWithNamedArrays, ArrayOrNames)
 CombineT = TypeVar("CombineT")  # used in CombineMapper
+CopyMapperT = TypeVar("CopyMapperT",  # used in CopyMapper
+                      Array, AbstractResultWithNamedArrays, ArrayOrNames)
```
inducer, line 56: Isn't the inclusion of `ArrayOrNames` redundant here?

pytato/transform/__init__.py
```diff
@@ -49,8 +49,11 @@
 from pymbolic.mapper.optimize import optimize_mapper
 
 ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
-MappedT = TypeVar("MappedT", bound=ArrayOrNames)
+MappedT = TypeVar("MappedT",
+                  Array, AbstractResultWithNamedArrays, ArrayOrNames)
 CombineT = TypeVar("CombineT")  # used in CombineMapper
+CopyMapperT = TypeVar("CopyMapperT",  # used in CopyMapper
```
inducer, line 55:
```suggestion
MapperResultT = TypeVar("CopyMapperT",  # used in CopyMapper
```
?
Or `CopyMapperResultT`?

inducer (commented):
Thanks! Looks good generally, just some nits below.

pytato/transform/__init__.py
inducer: :shrug: Oh well. Not a big issue. Thanks for checking!

inducer:
Thx! Pushed a minor fix, LGTM.

PR #432, opened 2023-04-21
Type cache key as `Hashable`

inducer:
This was the fix I thought I had pushed to #430  but didn't. :facepalm:


kaushikcfd (approved):
LGTM, thanks! I noticed the diff was zero, but then I thought maybe you noticed a bug in the commit message:).

PR #434, opened 2023-05-04
New distributed-memory DAG partitioner, rebased

inducer:
This supersedes #393 and #401 and is a rebase of those onto current main.

This has the beginnings of a new algorithm for the DAG partitioner that provably (I think!) will not artificially introduce cycles into the partitioned DAG.

Draft because:
- [ ]  It is not possible to implement a guaranteed cycle-free `find_distributed_partition` without communication. As a result, the interface to `find_distributed_partition` has to change in a backwards-incompatible way, so that we can get an MPI communicator. So this needs to merge at the same time as corresponding downstream PRs. `find_distributed_partition` is now also MPI-collective.
    - [ ] Remove the fall-back code path introduced by https://github.com/inducer/grudge/pull/298.
- [x] @majosm observed that running `examples/distributed.py` with three ranks caused cycles at a point. Add tests based on it.
- [x] Update the description of the algorithm in the docstring.

Notes:
- There's a description of the proposed algorithm in the docstring of the new `find_distributed_partition`. That may be nicer to look at rendered in sphinx.
- This generously pilfers ideas and some code from @kaushikcfd's approach in #277.

cc @majosm @matthiasdiener @kaushikcfd @nkoskelo

Closes: #378.


inducer:
@majosm It'd be nice if you could take a look to see if you spot any issues with the rebase.

My goal is to review the for-now separate commits and squash them into the "New DAG partitioner one" as I go, over the next few days, hopefully readying this for merge pretty soon.

inducer:
OK, this is all reviewed. I made a few notes and will push a few fixes, hopefully tomorrow. After that, this should be good to merge.

inducer:
This is ready now. I've updated the docs and cleaned up a few minor things. Thanks for your work on this @majosm! I think this turned out rather nicely.

PR #437, opened 2023-05-18
"Concatenate calls" transformation

inducer:
This is derived from a rebased version #364 (which it supersedes) and includes only the "concatenate calls" transformation.

Draft because:
- [ ] Review comments from #364 need to be worked on.
- [x] Sits on top of #436.

### DIFF
--- a/pytato/transform/calls.py
+++ b/pytato/transform/calls.py
@@ -1,8 +1,13 @@
-"""
+from __future__ import annotations
+
+__doc__ = """
 .. currentmodule:: pytato.transform.calls
 
 .. autofunction:: inline_calls
+.. autofunction:: concatenate_calls
 .. autofunction:: tag_all_calls_to_be_inlined
+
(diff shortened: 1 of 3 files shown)


pytato/transform/calls.py
```diff
@@ -111,4 +141,1398 @@ def tag_all_calls_to_be_inlined(expr: ArrayOrNames) -> ArrayOrNames:
...
+    """
+    ndim = next(iter(arrays)).ndim
+    return (all(ary.ndim == ndim for ary in arrays)
+            and all(_have_same_axis_length(arrays, idim)
+                    for idim in range(ndim)
+                    if idim != iaxis))
+
+
+@attrs.define(frozen=True)
+class _InputConcatabilityGetterAcc:
```
inducer, line 322: Maybe spell out "Accumulator" if that's what's meant?

inducer:
Superseded by #477.

PR #438, opened 2023-05-22
Split visualization.py

kaushikcfd:
Use separate file for emitting dot and emitting asciidag.

(A hidden agenda is that I'm working on another mode of visualization and would prefer keeping its logic in a separate file)


inducer:
Thx!

PR #449, opened 2023-07-25
Disable bounds checking when no loopy call present

matthiasdiener:
Alternative to https://github.com/inducer/arraycontext/pull/239.

### DIFF
--- a/pytato/target/loopy/codegen.py
+++ b/pytato/target/loopy/codegen.py
@@ -1087,6 +1087,7 @@ def generate_loopy(result: Union[Array, DictOfNamedArrays, Dict[str, Array]],
     # avoid such reduction iname collisions.
     t_unit = lp.make_reduction_inames_unique(state.t_unit)
 
+    # Disable bounds checking if there is no hand-written LoopyCall in the DAG.
     if not cg_mapper.has_loopy_call:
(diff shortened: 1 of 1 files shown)


matthiasdiener:
Is this what you had in mind @inducer?

inducer:
Yep, that's spot on. I think this makes sense. I'll wait for you to un-draft.

kaushikcfd:
Why are we disabling the bounds check only when there's a loopy call. Shouldn't we always disable it?

kaushikcfd:
> 
Why are we disabling the bounds check only when there's a loopy call. Shouldn't we always disable it?

Aah I think I see it. loopy_call corresponds to user code which needs bounds checking. Sorry for the noise.

matthiasdiener:
> Yep, that's spot on. I think this makes sense. I'll wait for you to un-draft.

Thanks! I've set it to ready for review.

inducer:
Given @kaushikcfd's question, I think it'd make sense to include a comment to explain the reasoning.

matthiasdiener:
> Given @kaushikcfd's question, I think it'd make sense to include a comment to explain the reasoning.

What do you think of ab38959?

inducer:
Works for me. Thanks!

PR #451, opened 2023-08-01
Pass-through is the more appropriate spelling for the adjective.

nkoskelo:
You are supposed to use pass-through for the adjective when it is directly next to the noun.


inducer:
Thanks!

PR #452, opened 2023-08-03
Remove `Array._fields`, deterministic-ify some iteration orders

inducer:
Best viewed commit-by-commit.

cc @majosm


pytato/array.py
```diff
@@ -1674,6 +1674,8 @@ def _is_eq_valid(self) -> bool:
         return True
 
     def __hash__(self) -> int:
+        # It would be better to hash the data, but we have no way of getting to
+        # it.
         return id(self)
```
kaushikcfd, line 1679: Is `(id(self.data), self.shape, self.dtype, self.name)` more appropriate?

pytato/stringifier.py
```diff
@@ -80,7 +80,7 @@ def map_foreign(self, expr: Any, depth: int) -> str:  # type: ignore[override]
         elif isinstance(expr, (dict, Map)):
             return ("{"
                     + ", ".join(f"{key!r}: {self.rec(val, depth)}"
-                                for key, val in expr.items())
+                                for key, val in sorted(expr.items()))
```
kaushikcfd, line 83: Nit: `sorted(expr.items(), key=lambda k_x_v: k_x_v[0])`.

inducer: Applied.

inducer:
Thanks for taking a look!

matthiasdiener:
As a side note, while playing around with https://github.com/matthiasdiener/orderedsets , I noticed that `immutables.Map` is also not deterministic, in contrast to `dict`/`frozendict`, e.g.:

```python
from immutables import Map
from frozendict import frozendict

m = Map({"a": None, "b": None, "c": None})
d = {"a": None, "b": None, "c": None}
f = frozendict({"a": None, "b": None, "c": None})

print("map", list(m.keys()))
print("dict", list(d.keys()))
print("frozendict", list(f.keys()))
```

```console
$ python map.py
map ['b', 'a', 'c']
dict ['a', 'b', 'c']
frozendict ['a', 'b', 'c']

$ python map.py
map ['a', 'c', 'b']
dict ['a', 'b', 'c']
frozendict ['a', 'b', 'c']
```

inducer:
Yep, I know. And that's OK. I also discussed this with @majosm the other day. Rather than rely on insertion orders throughout, I would prefer to have code generation be deterministic by construction, e.g. by sorting during traversals. That's also the direction that this PR takes.

PR #453, opened 2023-08-04
Define `__attrs_post_init__` only if `__debug__`, for all `Array` classes

inducer:
Follow-on to #452, noticed by @mtcam after attempting to pytato code with `-O`.


kaushikcfd:
Was this a noticeable performance hit? The body of such routines is typically just a bunch of asserts.

inducer:
No, I don't think it was measurable. Maybe only defining the routines `if __debug__` was a bad idea. Not sure.

PR #455, opened 2023-09-11
Allow reshapes for 'F' ordered arrays

a-alveyblanc:
- Add case for 'F' ordered arrays in `_get_reshaped_indices`
- Remove requirements that arrays *must* be 'C' ordered


inducer:
There seem to be some merge conflicts.

inducer:
> Remove requirements that arrays must be 'C' ordered

Note that the `order` argument in `reshape` has nothing to do with in-memory ordering of the array. Pytato arrays *do not have* a defined in-memory ordering, FWIW.

pytato/transform/lower_to_index_lambda.py
```diff
@@ -51,27 +51,50 @@ def _get_reshaped_indices(expr: Reshape) -> Tuple[ScalarExpression, ...]:
...
+            oldstrides.insert(0, oldstrides[0]*axis_len)
+
+        assert isinstance(expr.array.shape[-1], INT_CLASSES)
+        oldsizetills = [expr.array.shape[-1]]  # input array size
+                                               # till for axes idx
+        for old_axis_len in reversed(expr.array.shape[:-1]):
+            assert isinstance(old_axis_len, INT_CLASSES)
+            oldsizetills.insert(0, oldsizetills[0]*old_axis_len)
+
+    else:
```
inducer, line 79: For improved robsutness: `elif-else-error` is a good pattern.

pytato/transform/lower_to_index_lambda.py
```diff
@@ -51,27 +51,50 @@ def _get_reshaped_indices(expr: Reshape) -> Tuple[ScalarExpression, ...]:
...
+        oldstrides: List[IntegralT] = [1]  # input array strides
+        for axis_len in expr.array.shape[:-1]:
+            assert isinstance(axis_len, INT_CLASSES)
+            oldstrides.append(oldstrides[-1]*axis_len)
+
+        assert isinstance(expr.array.shape[0], INT_CLASSES)
+        oldsizetills = [expr.array.shape[0]]  # input array size till for axes idx
+        for old_axis_len in expr.array.shape[1:]:
+            assert isinstance(old_axis_len, INT_CLASSES)
+            oldsizetills.append(oldsizetills[-1]*old_axis_len)
```
inducer, line 97: Could the two branches be unified?

inducer, line 97: Check to make sure that this leaves axes alone that are "along for the ride". (I think that's already a bug in the existing code, too.)

inducer:

pytato/transform/lower_to_index_lambda.py
a-alveyblanc: I think this is only possible if we have information about the index order (C or F) before the reshape. As a small example, I'm assuming you mean that the resulting strides from reshaping (128, 64) -> (128, 4, 4) with the inner axes being F-ordered after the reshape should result in strides (64, 1, 4) instead of (1, 4, 64). Though, this assumes the index order is C ordered to begin with.

a-alveyblanc:
@inducer This is ready for another look

pytato/transform/lower_to_index_lambda.py
inducer: As discussed, likely possible by not indiscriminately linearizing all axes, but instead establishing an old $\to$ new axis mapping.)

inducer:

pytato/transform/lower_to_index_lambda.py
kaushikcfd: Is there a big need to not lineare the index expressions? With the exception of parametric shapes, ISL should be able to simplify them away.

I don't see anything wrong with pytato getting the old to new axis mapping, but the special case would make the implementation clunky.

a-alveyblanc: I've pushed a rough draft of the old -> new axis mapping to motivate the conversation a bit more

a-alveyblanc:
@inducer This is ready for another look. 

If it seems like there's still a fair amount of work to be done on the updated index computation (i.e. leaving unchanged axes alone), maybe we should consider moving that work to another branch in favor of getting F-ordered reshapes merged.

PR #461, opened 2023-10-05
replace immutables.Map with immutabledict

matthiasdiener:
Based on https://github.com/inducer/pytato/pull/457#issuecomment-1725926293, `immutabledict` has a consistent iteration order (like `dict`, i.e. insertion order), and is likely faster than `immutables.Map`, especially for actually immutable dicts (see also the graph [here](https://github.com/MagicStack/immutables#immutables)).


inducer:
Thanks!

PR #462, opened 2023-10-10
make MPI tags deterministic

matthiasdiener:
This requires tags to be sortable. The only other option I can see is to use some kind of ordered set type.

### DIFF
--- a/pytato/distributed/tags.py
+++ b/pytato/distributed/tags.py
@@ -31,7 +31,7 @@ THE SOFTWARE.
 """
 
 
-from typing import TYPE_CHECKING, Tuple, FrozenSet, Any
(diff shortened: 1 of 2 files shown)


pytato/distributed/tags.py
```diff
@@ -99,7 +99,7 @@ def set_union(
         next_tag = base_tag
         assert isinstance(all_tags, frozenset)
 
-        for sym_tag in all_tags:
+        for sym_tag in sorted(all_tags):
```
inducer, line 102: Are they necessarily ordered? What are we sorting by?

inducer:
Pushed some minor stuff, LGTM. Thanks!

PR #448, opened 2023-07-25
Separate `InlinedExpressionGenMapper` recursion from DAG recursion

majosm:
Calls `CodeGenMapper.rec()` for `IndexLambda`'s bound arrays up front before calling `InlinedExpressionGenMapper` in order to reduce the overall recursion depth.

PR #446, opened 2023-07-24
Eliminate recursive forwarding in mappers

majosm:
@inducer Turns out I didn't revert these changes after all, I just wasn't looking far enough down the commit list. 🙂

(I'll submit the `InlinedExpressionGenMapper` changes as a separate PR since they're a little less trivial.)

PR #466, opened 2023-10-31
Partially restore recursive forwarding in mappers

majosm:
Partially undoes #446 by restoring `__call__`s that forward to `rec`. Using `__call__ = rec` is wrong, because it doesn't pick up versions of `rec` defined by subclasses.

Doesn't appear to have much impact on recursion depth post-#448.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,10 @@ jobs:
         -   name: "Main Script"
             run: |
                 curl -L -O https://gitlab.tiker.net/inducer/ci-support/raw/main/build-and-test-py-project-within-miniconda.sh
+                # pytest gets angry when 'import pytato.array' may come
+                # from both the project file and the venv. This matters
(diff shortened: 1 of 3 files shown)


inducer:
The failures are a consequence of https://github.com/inducer/ci-support/commit/1475971303c26c0cbe995e4383777aebc97ed33e, which in turn appeared because nanobind does not allow `--editable` installs on Python 3.12. :scream: 

I don't totally understand why pytest is unhappy, TBH.

inducer:
Nvm, I guess I do. It has two ways to import, e.g. `pytato.utils`, and it is grouchy about the ambiguity.

inducer:
https://github.com/inducer/ci-support/commit/499226cc4d1388c6e35b16e2323fc8df18330503 might help, rerunning

inducer:
Nvm, I was confused. I think adding back the `--editable` locally is probably easiest.

PR #467, opened 2023-11-09
number_distributed_tags: sort tags by repr

matthiasdiener:
cc @MTcam

Should resolve issues like
```
TypeError: '<' not supported between instances of 'type' and 'type'
```

Followup to #462.

### DIFF
--- a/pytato/distributed/tags.py
+++ b/pytato/distributed/tags.py
@@ -106,7 +106,7 @@ def number_distributed_tags(
         next_tag = base_tag
         assert isinstance(all_tags, frozenset)
 
-        for sym_tag in sorted(all_tags):
+        for sym_tag in sorted(all_tags, key=lambda tag: repr(tag)):
             sym_tag_to_int_tag[sym_tag] = next_tag
             next_tag += 1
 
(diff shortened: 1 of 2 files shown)


pytato/distributed/tags.py
```diff
@@ -106,7 +106,7 @@ def set_union(
         next_tag = base_tag
         assert isinstance(all_tags, frozenset)
 
-        for sym_tag in sorted(all_tags):
+        for sym_tag in sorted(all_tags, key=lambda tag: repr(tag)):
```
inducer, line 109: This now requires that tags have a textual representation... so please update the docs.

kaushikcfd:
The repr should be consistent too. I don't think `Record.__repr__` is same
across interpreter runs.

matthiasdiener:
> The repr should be consistent too. I don't think `Record.__repr__` is same across interpreter runs.

Yes, that's true, and also applies to frozenset etc. Between this approach and #469, I'm not sure which one is better.

pytato/distributed/tags.py
matthiasdiener: Done in [2c9ad50](https://github.com/inducer/pytato/pull/467/commits/2c9ad5086b009fb2c0facf849c273a338c67d021).

matthiasdiener:
Closing in favor of #469.

PR #465, opened 2023-10-18
make find_distributed_partition deterministic

matthiasdiener:
Needs https://github.com/matthiasdiener/orderedsets

Needs:
- [x] https://github.com/inducer/loopy/pull/825
- [x] a new pytools release

Partially addresses https://github.com/inducer/pytato/issues/498

PR #469, opened 2023-11-09
number_distributed_tags: non-set, non-sorted numbering

matthiasdiener:
Alternative to #462/#467

cc @MTCam

*Please squash*

### DIFF
--- a/pytato/distributed/tags.py
+++ b/pytato/distributed/tags.py
@@ -65,6 +65,11 @@ def number_distributed_tags(
     """
     from pytools import flatten
 
+    # A potential optimization here could be to use a 'set' to collect the tags,
+    # but this would introduce non-determinism in the tag numbering. Another
+    # option would be use something like pytools.unique() to reduce the amount
+    # of data communicated, but since all sends and receives should each
+    # have unique tags, this would at most buy us a factor of 2.
     tags = tuple([
             recv.comm_tag
             for part in partition.parts.values()
@@ -77,6 +82,8 @@ def number_distributed_tags(
 
     root_rank = 0
 
+    # We can't let MPI do a set union here, since the result would be
+    # non-deterministic.
     all_tags = mpi_communicator.gather(tags, root=root_rank)
 
     if mpi_communicator.rank == root_rank:
--- a/pytato/distributed/tags.py
+++ b/pytato/distributed/tags.py
@@ -31,7 +31,7 @@ THE SOFTWARE.
 """
 
 
-from typing import TYPE_CHECKING, Tuple, FrozenSet, Optional, TypeVar
+from typing import TYPE_CHECKING, Tuple, TypeVar
 
 from pytato.distributed.partition import DistributedGraphPartition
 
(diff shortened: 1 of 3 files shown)


pytato/distributed/tags.py
```diff
@@ -62,53 +62,33 @@ def number_distributed_tags(
...
         *mpi_communicator*.
-
-    .. note::
-
-        This function requires that symbolic tags are comparable.
     """
-    tags = frozenset({
+    from pytools import flatten
+
+    tags = tuple([
```
matthiasdiener, line 68: Could potentially run `unique` here to shorten what needs to be sent.

test/test_distributed.py
```diff
@@ -881,7 +889,13 @@ def test_number_symbolic_tags_bare_classes(ctx_factory):
...
     partition = pt.find_distributed_partition(comm, outputs)
 
-    pt.number_distributed_tags(comm, partition, base_tag=4242)
+    (distp, next_tag) = pt.number_distributed_tags(comm, partition, base_tag=4242)
+
+    assert next_tag == 4244
+
+    # For the next assertion, find_distributed_partition needs to be
+    # deterministic too (https://github.com/inducer/pytato/pull/465).
+    # assert next(iter(distp.parts[0].name_to_send_nodes.values()))[0].comm_tag == 4242  # noqa: E501
```
matthiasdiener, line 898: When combining this PR with #465, this assertion passes.

matthiasdiener:
This is ready for review @inducer. As mentioned in the comment above, when combining this PR with #465, the (simple) determinism test passes.

pytato/distributed/tags.py
```diff
@@ -63,48 +63,32 @@ def number_distributed_tags(
...
-            # type ignore reason: mpi4py misdeclares op functions as returning
-            # None.
-            set_union,  # type: ignore[arg-type]
-            commute=True)
-    try:
-        all_tags = mpi_communicator.reduce(
-                tags, set_union_mpi_op, root=root_rank)
-    finally:
-        set_union_mpi_op.Free()
+    all_tags = mpi_communicator.gather(tags, root=root_rank)
```
inducer, line 80: Add a comment here on why we should not let MPI do the set union.

pytato/distributed/tags.py
inducer: It's at most a factor of two (since all receives and all sends should each have unique tags). Not worth the headache IMO.

test/test_distributed.py
inducer: File an issue to remind us to reenable it?

inducer (approved):
Thanks, LGTM once one comment is added.

pytato/distributed/tags.py
matthiasdiener: I added a comment regarding this in fab72fb5282a53e300e2ae375b22857bc7cc8e1b.

pytato/distributed/tags.py
matthiasdiener: Added in fab72fb5282a53e300e2ae375b22857bc7cc8e1b

test/test_distributed.py
matthiasdiener: Done in #482

inducer:
Thx!

Issue #470, opened 2023-11-10
Functions and axis tags don't get along


majosm:
The current implementation of `unify_axes_tags` doesn't play well with expressions containing function calls. Due to the possibility that the function could be something generic that accepts inputs of varying shapes (e.g. `a + b`), it's not allowed to propagate axis tags into/out of function definitions. This means that calls must be inlined before axis tag propagation can occur (which is OK for now, but we'll need to revisit this later).

inducer:
Upon thinking about this more, I think varying shapes are not an issue---pytato DAGs are always shape-specific (though parameters are *technically* allowed as part of shapes). So propagating tags *out of* a function should always be fine, but propagating *into* is not, because a function may be called from several call sites, which may disagree about how to tag things.

At any rate, thanks for getting this issue filed, @majosm!

PR #471, opened 2023-11-10
Fixes for functions

majosm:
Adds fixes needed to make outlining of inviscid/viscous fluxes work in mirgecom.

Note: Only works if functions are inlined at the beginning of `_dag_to_compiled_func` in grudge's lazy function caller, due to interactions with axis tagging (see #470).

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -31,7 +31,7 @@ from pytato.array import (Array, IndexLambda, Stack, Concatenate, Einsum,
                           DictOfNamedArrays, NamedArray,
                           IndexBase, IndexRemappingBase, InputArgumentBase,
                           ShapeType)
-from pytato.function import FunctionDefinition, Call
+from pytato.function import FunctionDefinition, Call, NamedCallResult
 from pytato.transform import Mapper, ArrayOrNames, CachedWalkMapper
 from pytato.loopy import LoopyCall
 from pymbolic.mapper.optimize import optimize_mapper
@@ -173,6 +173,15 @@ class NUserCollector(Mapper):
             if isinstance(dim, Array):
                 self.nusers[dim] += 1
                 self.rec(dim)
+
+    def map_call(self, expr: Call) -> None:
+        for ary in expr.bindings.values():
+            if isinstance(ary, Array):
+                self.nusers[ary] += 1
+                self.rec(ary)
+
+    def map_named_call_result(self, expr: NamedCallResult) -> None:
+        self.rec(expr._container)
 
 # }}}
 
@@ -358,6 +367,11 @@ class DirectPredecessorsGetter(Mapper):
                                         ) -> FrozenSet[Array]:
         return frozenset([expr.passthrough_data])
 
+    def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
+        # Not sure if this should descend into the function definition or not?
+        return frozenset(expr.call.bindings.values())
+
+
 # }}}
 
 
@@ -423,7 +437,7 @@ class CallSiteCountMapper(CachedWalkMapper):
         if not self.visit(expr):
             return
 
-        new_mapper = self.clone_for_callee()
+        new_mapper = self.clone_for_callee(expr)
         for subexpr in expr.returns.values():
             new_mapper(subexpr, *args, **kwargs)
 
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -431,6 +431,13 @@ class CallSiteCountMapper(CachedWalkMapper):
     def get_cache_key(self, expr: ArrayOrNames) -> int:
         return id(expr)
 
+    def update_from_callee_clone(
+            self,
+            function: FunctionDefinition,
+            callee_clone: CallSiteCountMapper) -> None:
+        super().update_from_callee_clone(function, callee_clone)
+        self.count += callee_clone.count
+
     @memoize_method
     def map_function_definition(self, /, expr: FunctionDefinition,
                                 *args: Any, **kwargs: Any) -> None:
@@ -440,8 +447,7 @@ class CallSiteCountMapper(CachedWalkMapper):
         new_mapper = self.clone_for_callee(expr)
         for subexpr in expr.returns.values():
             new_mapper(subexpr, *args, **kwargs)
-
-        self.count += new_mapper.count
+        self.update_from_callee_clone(expr, new_mapper)
 
         self.post_visit(expr, *args, **kwargs)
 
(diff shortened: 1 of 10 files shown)


pytato/analysis/__init__.py
```diff
@@ -358,6 +367,11 @@ def map_distributed_send_ref_holder(self,
                                         ) -> FrozenSet[Array]:
         return frozenset([expr.passthrough_data])
 
+    def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
+        # Not sure if this should descend into the function definition or not?
+        return frozenset(expr.call.bindings.values())
```
inducer, line 372: Refuse to guess, for now.

.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
                 curl -L -O https://tiker.net/ci-support-v0
                 . ./ci-support-v0
                 build_py_project_in_conda_env
-                python -m pip install mypy
+                python -m pip install mypy==1.6.1
```
inducer, line 56: Revert before merge

pytato/transform/__init__.py
```diff
@@ -251,13 +254,23 @@ def __call__(self,  # type: ignore[override]
...
-    def clone_for_callee(self: _SelfMapper) -> _SelfMapper:
+    def clone_for_callee(
+            self: _SelfMapper, function: FunctionDefinition) -> _SelfMapper:
         """
         Called to clone *self* before starting traversal of a
         :class:`pytato.function.FunctionDefinition`.
         """
         return type(self)()
 
+    def update_from_callee_clone(
```
inducer, line 265: Maybe shuffle the `update_from_callee_clone` changes into a separate PR?

pytato/transform/__init__.py
```diff
@@ -907,13 +921,9 @@ def map_function_definition(self, expr: FunctionDefinition
                                                  for ret in expr.returns.values()])
         result: Set[InputArgumentBase] = set()
         for inp in all_callee_inputs:
-            if isinstance(inp, Placeholder):
-                if inp.name in expr.parameters:
-                    # drop, reference to argument
-                    pass
-                else:
-                    raise ValueError("function definition refers to non-argument "
```
inducer, line 915: This error needs to stay I think. But we could add an additional interface to `outline` that "silently" introduces arguments for symbols used from the surrounding code. Probably not willy-nilly, but from a given list.

pytato/transform/__init__.py
inducer, line 1: Just don't attempt to handle functions in MPMS. (Make sure to raise an error.)

pytato/transform/calls.py
```diff
@@ -50,7 +50,10 @@ def __init__(self, substitutions: Mapping[str, Array]) -> None:
         self.substitutions = substitutions
 
     def map_placeholder(self, expr: Placeholder) -> Array:
-        return self.substitutions[expr.name]
+        if expr.name in self.substitutions:
```
inducer, line 53: Back out.

pytato/transform/calls.py
```diff
@@ -63,20 +66,20 @@ def map_call(self, expr: Call) -> AbstractResultWithNamedArrays:
         assert isinstance(new_expr, Call)
 
         if expr.tags_of_type(InlineCallTag):
-            substitutor = PlaceholderSubstitutor(expr.bindings)
+            substitutor = PlaceholderSubstitutor(new_expr.bindings)
```
inducer, line 69: I agree with the change, but I'd like to make sure I trust my understanding.

I'm surprised you ran into an issue with it and would like to know about the failure you saw.

pytato/transform/calls.py
```diff
@@ -63,20 +66,20 @@ def map_call(self, expr: Call) -> AbstractResultWithNamedArrays:
...
+                tags=new_expr.tags
             )
         else:
             return new_expr
 
     def map_named_call_result(self, expr: NamedCallResult) -> Array:
         new_call = self.rec(expr._container)
         assert isinstance(new_call, AbstractResultWithNamedArrays)
-        return new_call[expr.name]
+        return new_call[expr.name].expr
```
inducer, line 82: This code probably needs to check whether `new_call` (fix naming!) actually is a call or a `DictOfNamedArrays`.

pytato/codegen.py
```diff
@@ -112,8 +113,16 @@ def __init__(self, target: Target,
         self.target = target
         self.kernels_seen: Dict[str, lp.LoopKernel] = kernels_seen or {}
 
-    def clone_for_callee(self) -> CodeGenPreprocessor:
-        return CodeGenPreprocessor(self.target, self.kernels_seen)
+    def clone_for_callee(
+            self: _SelfMapper, function: FunctionDefinition) -> CodeGenPreprocessor:
+        # Would need to sync bound_arguments/var_name_gen from clones
```
inducer, line 118: Disagree with the comment.

pytato/codegen.py
```diff
@@ -112,8 +113,16 @@ def __init__(self, target: Target,
...
+        raise NotImplementedError("CodeGenPreprocessor does not support functions.")
+
+    def update_from_callee_clone(
+            self: _SelfMapper, function: FunctionDefinition,
+            callee_clone: _SelfMapper) -> None:
+        # Would need to sync bound_arguments/var_name_gen from clones
```
inducer, line 124: Disagree with the comment.

pytato/function.py
```diff
@@ -132,19 +131,19 @@ class FunctionDefinition(Taggable):
...
-        from functools import reduce
 
         mapper = InputGatherer()
 
-        all_input_args: FrozenSet[InputArgumentBase] = reduce(
-            frozenset.union,
-            (mapper(ary) for ary in self.returns.values()),
-            frozenset()
-        )
+        all_placeholder_args: FrozenSet[Placeholder] = frozenset()
```
inducer, line 137: Back out.

pytato/distributed/partition.py
```diff
@@ -288,8 +299,33 @@ def __init__(self,
...
+            # duplicating placeholders
+            self._placeholder_generator)
+
+    def update_from_callee_clone(
+            self,
+            function: FunctionDefinition,
+            callee_clone: _DistributedInputReplacer) -> None:
+        super().update_from_callee_clone(function, callee_clone)
+
+        self.user_input_names |= (
```
inducer, line 326: This also seems rooted in wanting to allow names other than parameters in functions, which I think is not what we want.

pytato/transform/calls.py
majosm: I don't recall if I actually ran into an issue with this. But if I understand correctly: if any of the bindings contain calls themselves, `new_expr.bindings` will have them inlined (where tagged), whereas `expr.bindings` won't.

inducer (approved):
:tada: Thanks!

inducer:
@majosm The mypy fixes are in, as is #474. Could you update this so we can merge it?

majosm:
@inducer Updated and removed mypy version change. Should be good to go.

inducer:
Thx!

PR #472, opened 2023-11-11
Fixes for mypy 1.7

inducer:


inducer:
Thanks for the patch! The reason I got stuck here is that I suspect some of these are honest bugs in mypy which should be reported rather than "papered over" with `cast` (because they're somewhat likely to bite us again, in a different form).

Here's one example that I filed as I was working on this: https://github.com/python/mypy/issues/16468. (This implies that bugs aren't out of the question.) Fortunately, this one was straightforward to work around on our end.

inducer:
Here's the next one that I think is wrong: https://github.com/python/mypy/issues/16499

inducer:
OK, looks like this will perhaps all take a bit longer on the mypy end. I've worked through the failures here, should be all good to go. @matthiasdiener Thanks for the patch, though I've decided to go with type-ignores over casts so that mypy gets a chance to tell us when (hopefullly/finally) they won't be needed any longer.

PR #477, opened 2023-12-12
"Concatenate calls" transformation (updated)

majosm:
Updates #437. (I don't have write access so I can't do it directly.)


majosm:
@inducer I can reproduce the zip length mismatch error in `wave-op-mpi` with the above pytato branch, inducer/arraycontext#221, and [this grudge branch](https://github.com/majosm/grudge/tree/concatenate-calls-error-reproducer).

PR #479, opened 2024-01-26
Add `make_distributed_send_ref_holder`

majosm:
Adds a function to make a `DistributedSendRefHolder`, so we don't need to call its `__init__` method directly in inducer/grudge#285. (Calling `__init__` doesn't play well with the `CreatedAt` tags added in #284.)

### DIFF
--- a/pytato/distributed/nodes.py
+++ b/pytato/distributed/nodes.py
@@ -61,7 +61,7 @@ from pytools.tag import Taggable, Tag
 
 from pytato.array import (
         Array, _SuppliedShapeAndDtypeMixin, ShapeType, AxesT,
-        _get_default_axes, ConvertibleToShape, normalize_shape)
+        _get_default_axes, _get_default_tags, ConvertibleToShape, normalize_shape)
 
 CommTagType = Hashable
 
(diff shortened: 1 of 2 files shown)


majosm:
@inducer @matthiasdiener How's this look?

matthiasdiener:
LGTM, thanks!

pytato/distributed/nodes.py
```diff
@@ -226,17 +223,28 @@ def make_distributed_send(sent_data: Array, dest_rank: int, comm_tag: CommTagTyp
...
 
 
+def make_distributed_send_ref_holder(
+        send: DistributedSend,
+        passthrough_data: Array,
+        tags: FrozenSet[Tag] = frozenset()
+        ) -> DistributedSendRefHolder:
+    """Make a :class:`DistributedSendRefHolder` object."""
+    return DistributedSendRefHolder(
+        send=send, passthrough_data=passthrough_data, tags=tags)
```
inducer, line 233: Call `_get_default_tags` here. (Probably also for `make_distributed_send` above and `make_distributed_recv` below.)

majosm: https://github.com/inducer/pytato/pull/479/commits/d7bebfff8c0c2df0657caa22267d6adcd4963748?

inducer:
LGTM, thanks!

PR #483, opened 2024-02-06
fix some warnings in tests

matthiasdiener:
This reduces the number of warnings from ~7800 to ~380.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -2230,7 +2230,10 @@ def arange(*args: Any, **kwargs: Any) -> Array:
     step = dtype.type(inf.step)
     stop = dtype.type(inf.stop)
 
-    size = max(0, int(np.ceil((np.real(stop)-np.real(start))/np.real(step))))
+    from math import ceil
+    # np.real() suppresses "ComplexWarning: Casting complex values to real
+    # discards the imaginary part":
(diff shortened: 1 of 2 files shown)


pytato/array.py
```diff
@@ -2230,8 +2230,7 @@ def arange(*args: Any, **kwargs: Any) -> Array:
     step = dtype.type(inf.step)
     stop = dtype.type(inf.stop)
 
-    from math import ceil
-    size = max(0, int(ceil((stop-start)/step)))
+    size = max(0, int(np.ceil((np.real(stop)-np.real(start))/np.real(step))))
```
inducer, line 2233: The `np.real` kind of worries me. Can you add (in a comment) a link documenting that this matches numpy's behavior? Numpy's behavior isn't great FWIW:
```
>>> import numpy as np
>>> np.arange(2j, 5j, 1j)
array([], dtype=complex128)
```
but I think at least we should match numpy as best we can.

matthiasdiener: I added a comment and some new tests in [42f8f7a](https://github.com/inducer/pytato/pull/483/commits/42f8f7a4156d9c9d56870b10064f0c8f31d0e5f8). As far as I can see, adding `np.real` doesn't change the behavior of `pt.arange`, it just suppresses the `ComplexWarning`.

inducer: OK, fair enough.

PR #487, opened 2024-03-05
Update to pass *args and **kwargs through

nkoskelo:
When copying an IndexLambda with the CopyMapperAndExtraArgs we do not pass the extra arguments to the dispatcher looking at the subexpressions.


inducer:
Thanks!

inducer:
As a comment, when writing commit messages (and PR titles), assume the reader has no context. (i.e. they're looking at the first line of the commit message like an email subject line).

So, instead of "Update to pass *args and **kwargs through", a better message might be "CopyMapperAndExtraArgs: actually pass args for IndexLambda" or so. Similar for the other commit.

Also, strive to include as much information in the commit message, as opposed to the PR description. The latter will be lost to time. (Plus, Github automatically grabs the commit message and uses it as the PR description.)

PR #489, opened 2024-03-15
Add node type counter

MTCam:
Adds an analysis util to count nodes of each type.


pytato/analysis/__init__.py
```diff
@@ -413,6 +413,49 @@ def get_num_nodes(outputs: Union[Array, DictOfNamedArrays]) -> int:
...
+        from collections import defaultdict
+        super().__init__()
+        self.counts = defaultdict(int)
+
+    def get_cache_key(self, expr: ArrayOrNames) -> int:
+        return id(expr)
+
+    def post_visit(self, expr: Any) -> None:
+        if type(expr) not in self.counts:
+            self.counts[type(expr)] = 0
```
majosm, line 438: I think `defaultdict` takes care of this, right?

pytato/analysis/__init__.py
```diff
@@ -413,6 +413,49 @@ def get_num_nodes(outputs: Union[Array, DictOfNamedArrays]) -> int:
...
+        self.counts[type(expr)] += 1
+
+
+def get_num_node_types(outputs: Union[Array, DictOfNamedArrays]) -> Dict[Type, int]:
```
majosm, line 442: `get_num_node_types` is a little confusing (sounds like it would return the number of distinct types in the DAG). I used `get_node_counts` in mine, which isn't much better... `get_node_type_counts`? 🤷‍♂️

majosm (commented):
This looks familiar. 🙂 I added something almost exactly the same in my concatenation changes.

Added a couple of suggestions here.

pytato/analysis/__init__.py
MTCam: TIL.

pytato/analysis/__init__.py
MTCam: Agreed.  `get_node_type_counts` seems like the clear choice.

PR #490, opened 2024-03-16
Placeholder default dtype to float64

kaushikcfd:
Mildly controversial, but borrows design decisions from `np.(empty|ones|zeros)`.


inducer:
Could you speak regarding the motivation? I'm not totally against it, but still curious.

kaushikcfd:
Less characters, makes it easier to use, shorter demos? :)

inducer:
Fair enough. :)

PR #494, opened 2024-04-22
Update tag propagation logic

a-alveyblanc:
- Move tag propagation graph generation to  [pytools](https://github.com/inducer/pytools/pull/220)
- Respect `IgnoredForPropagationTag` in tag propagation


requirements.txt
a-alveyblanc, line 1: Revert this

pytato/transform/metadata.py
a-alveyblanc, line 1: Add example using tensor product axes

a-alveyblanc:
Looks like pylint and mypy are failing even though the changes in https://github.com/inducer/pytools/pull/220 have been merged. They don't fail locally :man_shrugging:

inducer:
https://pypi.org/project/pytools/2024.1.14/

inducer:
We should add a dependency on that version in `pyproject.toml`.

a-alveyblanc:
Added dependencies to `setup.py`. It seems like that's where other project information lives, and adding dependencies to `pyproject.toml` caused some errors.

pytato/transform/metadata.py
```diff
@@ -687,6 +656,12 @@ def __call__(self, expr: ArrayOrNames) -> ArrayOrNames:  # type: ignore[override
         assert isinstance(result, (Array, AbstractResultWithNamedArrays))
         return result
 
+# }}}
+
+
+class IgnoredForPropagationTag(Tag):
```
inducer, line 662: - Include in docs.
- Explain the semantics.
- Add a use case/justification for why one might possibly want this.

inducer (approved):

PR #497, opened 2024-05-08
Require `Mapping` attributes of arrays to be `immutabledict`s

### LINKED ISSUE #496: Equality of objects doesn't imply equality of hashes for certain array types
Array types that contain `Mapping`s can, under some circumstances, compare equally but hash unequally. This is due to a combination of two factors:

1. Equality comparison and hashing are implemented slightly differently for arrays. `EqualityComparer` compares the keys and values between the two mappings. Hashing uses whatever is generated by `attrs.frozen` (presumably calling `hash()` on the whole mapping object?).
2. Different mapping types (e.g., `immutables.Map` and `immutabledict`) implement hashing differently, so two mappings with different types can have the same contents but not hash the same.

This can be avoided by always using the same mapping class for these attributes, but that raises the question of whether the type annotation should continue to be `Mapping`, or if it should be made more restrictive (and possibly asserted to be a certain type in `__init__`?). Another option might be to modify the hashing so that it hashes the keys/values instead of the whole mapping object. I'm not sure which way is preferable.

@inducer This is what was causing my `CopyMapper` duplication issues. The code in main had been updated to use `immutabledict` in #461, but the concatenation branches of pytato/arraycontext still had some `immutables.Map`s left over.

cc @matthiasdiener

majosm:
Fixes #496.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -807,7 +807,8 @@ class DictOfNamedArrays(AbstractResultWithNamedArrays):
 
     .. automethod:: __init__
     """
-    _data: Mapping[str, Array]
+    _data: Mapping[str, Array] = attrs.field(
+        validator=attrs.validators.instance_of(immutabledict))
 
     _mapper_method: ClassVar[str] = "map_dict_of_named_arrays"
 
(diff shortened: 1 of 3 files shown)


inducer:
Perfect, thanks!

PR #501, opened 2024-06-11
Require arrays to have a `CreatedAt` tag if traceback tagging is enabled

majosm:
Also, make sure `non_equality_tags` gets propagated through mappers.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -449,9 +449,9 @@ class Array(Taggable):
     tags: FrozenSet[Tag] = attrs.field(kw_only=True)
 
     # These are automatically excluded from equality in EqualityComparer
-    non_equality_tags: FrozenSet[Optional[Tag]] = attrs.field(kw_only=True,
-                                                              hash=False,
-                                                              default=frozenset())
+    non_equality_tags: FrozenSet[Tag] = attrs.field(kw_only=True,
+                                                    hash=False,
+                                                    default=frozenset())
 
     _mapper_method: ClassVar[str]
 
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -463,6 +463,16 @@ class Array(Taggable):
 
     if __debug__:
         def __attrs_post_init__(self) -> None:
+            if _ENABLE_TRACEBACK_TAG:
+                from pytato.tags import CreatedAt
+                ntags = sum(
+                    1 if isinstance(tag, CreatedAt) else 0
+                    for tag in self.non_equality_tags)
+                if ntags < 1:
+                    raise AssertionError("Array has no CreatedAt tag.")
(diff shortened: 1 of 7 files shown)


pytato/array.py
```diff
@@ -463,6 +463,16 @@ def _is_eq_valid(self) -> bool:
 
     if __debug__:
         def __attrs_post_init__(self) -> None:
+            if _ENABLE_TRACEBACK_TAG:
+                from pytato.tags import CreatedAt
+                ntags = sum(
+                    1 if isinstance(tag, CreatedAt) else 0
+                    for tag in self.non_equality_tags)
```
inducer, line 470:
```suggestion
                ntags = len(
                    tag for tag in self.non_equality_tags
                    if isinstance(tag, CreatedAt))
```

pytato/array.py
```diff
@@ -723,16 +733,22 @@ def copy(self, *,  # type: ignore[override]
...
+             tags: Optional[FrozenSet[Tag]] = None,
+             non_equality_tags: Optional[FrozenSet[Tag]] = None) -> NamedArray:
         container = self._container if container is None else container
         name = self.name if name is None else name
         tags = self.tags if tags is None else tags
         axes = self.axes if axes is None else axes
+        non_equality_tags = (
+            self.non_equality_tags
+            if non_equality_tags is None
+            else non_equality_tags)
```
inducer, line 745: This is a generator expression, not a tuple. It can only be iterated over once. Make this explicitly a tuple.

pytato/array.py
```diff
@@ -723,16 +733,22 @@ def copy(self, *,  # type: ignore[override]
...
 
         return type(self)(container=container,
                           name=name,
                           tags=tags,
-                          axes=axes)
+                          axes=axes,
+                          non_equality_tags=non_equality_tags)
```
inducer, line 751: Could you look into why mypy didn't complain here? I don't believe the generator expression should be a `Sequence[...]`.

inducer (commented):
LGTM. Just a few small wrinkles.

pytato/array.py
majosm: There's no loop here, so the parens are just for line breakage. This is probably why mypy isn't complaining. 🙂

pytato/array.py
majosm: Oh, hm. Looks like I can't do that, since the argument is a generator expression. Should I make it a list instead? Or revert?

inducer: In that case, totally fine the way it is.

pytato/array.py
inducer: Ah, nvm. I had misread this code.

PR #502, opened 2024-06-14
Enable Parameter Study or Uncertainty Quantification into Inner Loop

nkoskelo:
A typical strategy for conducting parameter studies or uncertainty quantification is to rerun the same code multiple times with slightly different data.

The goal of this PR is to make it so that you can write your application code for the single application instance. Then, when you want to conduct a parameter study instead of rerunning the code multiple times you compile the code differently and then run it with larger inputs.

You still need to specify which variables can change between instances and how many instances you want to run.


inducer:
Doesn't seem finished, needs tesst at least, not passing linter. Converting to draft.

PR #503, opened 2024-06-15
More fixes for functions

majosm:
(Probably easier to read commit by commit.)

### DIFF
--- a/pytato/function.py
+++ b/pytato/function.py
@@ -242,7 +242,6 @@ class NamedCallResult(NamedArray):
         The name by which the returned array is referred to in
         :attr:`FunctionDefinition.returns`.
     """
-    name: str
     _mapper_method: ClassVar[str] = "map_named_call_result"
 
     def with_tagged_axis(self, iaxis: int,
--- a/pytato/function.py
+++ b/pytato/function.py
@@ -215,6 +215,7 @@ class FunctionDefinition(Taggable):
             return tuple(call_site[f"_{iarg}"]
                          for iarg in range(len(self.returns)))
         elif self.return_type == ReturnType.DICT_OF_ARRAYS:
+            # FIXME: Should this be immutabledict?
             return {kw: call_site[kw] for kw in self.returns}
         else:
             raise NotImplementedError(self.return_type)
--- a/pytato/function.py
+++ b/pytato/function.py
@@ -226,6 +226,7 @@ class FunctionDefinition(Taggable):
             raise NotImplementedError(self.return_type)
 
 
+@attrs.frozen(eq=False, repr=False, hash=True, cache_hash=True)
 class NamedCallResult(NamedArray):
     """
     One of the arrays that are returned from a call to :class:`FunctionDefinition`.
@@ -242,15 +243,6 @@ class NamedCallResult(NamedArray):
     call: Call
     name: str
     _mapper_method: ClassVar[str] = "map_named_call_result"
-
-    def __init__(self,
-                 call: Call,
-                 name: str) -> None:
-        super().__init__(call, name,
-                         axes=call.function.returns[name].axes,
-                         tags=call.function.returns[name].tags,
-                         non_equality_tags=(
-                             call.function.returns[name].non_equality_tags))
 
     def with_tagged_axis(self, iaxis: int,
                          tags: Sequence[Tag] | Tag) -> Array:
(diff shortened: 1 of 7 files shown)


majosm:
@inducer I think this is ready for a look. (I'm getting some lingering doc errors about `np.bool_`, but I see them with main too, so I think they're unrelated?)

pytato/function.py
```diff
@@ -163,7 +164,8 @@ def _with_new_tags(
             self: FunctionDefinition, tags: FrozenSet[Tag]) -> FunctionDefinition:
         return attrs.evolve(self, tags=tags)
 
-    def __call__(self, **kwargs: Array
+    @memoize_method
```
inducer, line 170: I'm not sure I like this, as this will conceivably store giant DAGs (for the arguments) in the function. If this is to avoid redundant processing in mappers, I think the caches for that would be better off living in those mappers.

pytato/function.py
```diff
@@ -143,14 +143,6 @@ def _placeholders(self) -> Mapping[str, Placeholder]:
...
             all_placeholders |= new_placeholders
 
-        # FIXME: Need a way to check for *any* captured arrays, not just placeholders
-        if __debug__:
-            pl_names = frozenset(arg.name for arg in all_placeholders)
-            extra_pl_names = pl_names - self.parameters
-            assert not extra_pl_names, \
-                f"Found non-argument placeholder '{next(iter(extra_pl_names))}' " \
-                "in function definition."
-
```
inducer, line 152: I know the commit message says this will be done in arraycontext, but unless it's a huge expense, it would let me rest easier if we left this check in place.

pytato/function.py
```diff
@@ -197,6 +197,7 @@ def __call__(self, /, **kwargs: Array
             return tuple(call_site[f"_{iarg}"]
                          for iarg in range(len(self.returns)))
         elif self.return_type == ReturnType.DICT_OF_ARRAYS:
+            # FIXME: Should this be immutabledict?
```
inducer, line 211: Yep!

pytato/transform/__init__.py
```diff
@@ -938,6 +938,8 @@ def map_function_definition(self, expr: FunctionDefinition
                 if inp.name in expr.parameters:
                     # drop, reference to argument
                     pass
+                # FIXME: Checked upon function definition creation in arraycontext
+                # now, can probably drop this
```
inducer, line 942: As in the other comment: nah, unless it's very expensive.

inducer (commented):
Thanks for working on this. Just a few hopefully relatively simple things, then this should be good to go.

pytato/transform/__init__.py
```diff
@@ -1204,6 +1205,20 @@ def rec(self, expr: ArrayOrNames, *args: Any, **kwargs: Any
     def clone_for_callee(
             self: _SelfMapper, function: FunctionDefinition) -> _SelfMapper:
         return type(self)()
+
+    def map_function_definition(self, expr: FunctionDefinition,
+                                *args: Any, **kwargs: Any) -> None:
+        if not self.visit(expr) or expr in self._visited_functions:
```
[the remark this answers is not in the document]
majosm: Added `get_func_def_cache_key`.

pytato/function.py
majosm: Reverted; doesn't appear to be needed anymore.

pytato/function.py
majosm: Reverted.

pytato/transform/__init__.py
majosm: Removed FIXME.

pytato/transform/__init__.py
```diff
@@ -1199,6 +1200,10 @@ def rec(self, expr: ArrayOrNames, *args: Any, **kwargs: Any
         super().rec(expr, *args, **kwargs)
         self._visited_nodes.add(cache_key)
 
+    @memoize_method
```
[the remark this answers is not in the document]
majosm: Isn't that the way it should be? AFAIK all other node types would be traversed twice if they're in both `f2` and `f3`. Should functions be the same or different?

majosm:
Should be ready for another look. (Previous review was up through the second "add FIXME".)

pytato/transform/__init__.py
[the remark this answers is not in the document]
majosm: I think I'm going to add this in a new PR, the change set is getting kind of big (it's also overlapping with some of the changes in #515, so I might split a few of those also into a separate PR and try to get them merged first). I've removed the `@memoize_method` annotations here in the meantime.

majosm:
> Should be ready for another look. (Previous review was up through the second "add FIXME".)

Ready for real this time. @inducer

pytato/transform/__init__.py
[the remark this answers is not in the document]
majosm: #531

pytato/transform/__init__.py
```diff
@@ -1215,19 +1227,44 @@ class CachedWalkMapper(WalkMapper):
 
     def __init__(self) -> None:
         super().__init__()
-        self._visited_nodes: set[Any] = set()
+        self._visited_arrays_or_names: set[Any] = set()
```
inducer, line 1230: Leave these changes for #531?

majosm: Reverted.

PR #506, opened 2024-06-17
doc: Work around np.bool_ change

matthiasdiener:
Doc builds fail due to https://github.com/numpy/numpy/pull/25080


inducer:
Thx!

PR #507, opened 2024-06-18
Add node type counter and tests

kajalpatelinfo:
Replacing #489 .

This PR builds off the work in #489 by adding tests to check the count of nodes and types of nodes in a DAG. `NodeCountMapper` counts duplicate nodes once.


pytato/analysis/__init__.py
```diff
@@ -381,22 +383,39 @@ def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
...
+
+
+def get_node_type_counts(outputs: Union[Array, DictOfNamedArrays]) -> Dict[Type[Any], int]:
+    """
+    Returns a dictionary mapping node types to node count for that type
+    in DAG *outputs*.
+    """
+
+    from pytato.codegen import normalize_outputs
+    outputs = normalize_outputs(outputs)
```
majosm, line 413: Question for @inducer, @matthiasdiener: do we actually need the `normalize_outputs` call here (and in `get_num_nodes`)? It looks like `outputs` is either an `Array` or a `DictOfNamedArrays`, either of which should already be a valid input to `NodeCountMapper`, right?

I ask because it leads to slightly unintuitive behavior when `outputs` is an `Array` (`normalize_outputs` adds a `DictOfNamedArrays`, so the node count ends up being 1 more than expected).

test/test_pytato.py
```diff
@@ -585,7 +584,7 @@ def test_repr_array_is_deterministic():
         assert repr(dag) == repr(dag)
 
 
-def test_nodecountmapper():
+def test_node_count_mapper():
```
majosm, line 587: Does this original test do anything that the added tests below don't? If not, we could probably remove it. If it does, maybe change the name to something that describes specifically what it tests?

majosm (commented):
Looks great! Just added a couple of minor thoughts.

pytato/analysis/__init__.py
majosm: @kajalpatelinfo The consensus seems to be that we should just drop the `DictOfNamedArrays` from the node counts (for both `get_num_nodes` and `get_node_type_counts`). I'm told that `DictOfNamedArrays` can potentially occur in the middle of DAGs too, not just at the end, so the best way to handle this might be to check inside `NodeCountMapper.post_visit` and only increment if it's not a `DictOfNamedArrays`.

We should probably also document this behavior. Could you add a line to the docstrings of `get_num_nodes` and `get_node_type_counts` explaining that `DictOfNamedArrays`s are ignored?

kajalpatelinfo: Got it, I'll add those changes tonight!

pytato/analysis/__init__.py
```diff
@@ -381,34 +383,58 @@ def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
...
+        if type(expr) is not DictOfNamedArrays:
+            self.counts[type(expr)] += 1
+
+
+def get_node_type_counts(outputs: Union[Array, DictOfNamedArrays]) -> Dict[Type[Any], int]:
+    """
+    Returns a dictionary mapping node types to node count for that type
+    in DAG *outputs*.
+
+    `DictOfNamedArrays` are added when *outputs* is normalized and ignored.
```
majosm, line 412: I think the normalization is an implementation detail that the caller doesn't necessarily know about. I would maybe instead say something along the lines of "Instances of `DictOfNamedArrays` are excluded from counting.".

pytato/analysis/__init__.py
```diff
@@ -381,34 +383,58 @@ def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
...
+
+    return ncm.counts
 
 
 def get_num_nodes(outputs: Union[Array, DictOfNamedArrays]) -> int:
-    """Returns the number of nodes in DAG *outputs*."""
+    """
+    Returns the number of nodes in DAG *outputs*.
+
+    `DictOfNamedArrays` are added when *outputs* is normalized and ignored.
```
majosm, line 428: Same for this one.

test/test_pytato.py
```diff
@@ -585,19 +584,91 @@ def test_repr_array_is_deterministic():
...
+    data = np.random.rand(4, 4)
+    single_node_dag = pt.make_dict_of_named_arrays(
+        {"result": pt.make_data_wrapper(data)})
+
+    # Get counts per node type
+    node_counts = get_node_type_counts(single_node_dag)
+
+    # Assert that there is only one node of type DataWrapper
+    assert node_counts == {pt.DataWrapper: 1}
+    assert sum(node_counts.values()) == 1  # Total node count is 1
```
majosm, line 611:
```suggestion
```
(The first assertion seems sufficient.)

majosm (commented):
A couple more suggestions, otherwise looks good to go on my end.

majosm:
Looks good to me! @inducer?

majosm:
Hey @kajalpatelinfo , it looks like there's a test in `test_codegen.py` that depends on the previous behavior of `get_num_nodes`. I talked to @inducer and we were thinking maybe we should just go straight to #508 instead of merging this one separately. Then we can just add `count_duplicates=True` to the test instead of figuring out what to do with it here in the meantime. Sound OK?

kajalpatelinfo:
Yep, that makes sense!! I'll go over the changes to #508 tonight. Thank you!

inducer:
Closing, superseded by #508.

PR #508, opened 2024-06-23
Add duplicate node counter functionality and tests

kajalpatelinfo:
Building off #507.

This PR builds off #507 by adding functionality and tests to count duplicates however many times they appear in the DAG. The default functionality is to _not_ count duplicates.


kajalpatelinfo:
@majosm Please review when convenient!

pytato/analysis/__init__.py
```diff
@@ -381,34 +383,86 @@ def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
...
+    return sum(ncm.expr_type_counts.values())
+
+
+def get_expr_calls(
+        outputs: Union[Array, DictOfNamedArrays],
+        count_duplicates: bool = False
+        ) -> Dict[Type[Any], int]:
+    """
+    Returns the count of calls per `expr`.
+    """
```
majosm, line 457: Couple of suggestions for this:

1. The name is a little ambiguous, since [Call](https://github.com/inducer/pytato/blob/13d19a44526e4d61c60b2f2bbe5b47e99ae7258d/pytato/function.py#L267) is a thing. 🙂 Maybe `get_node_multiplicities`?
2. Think about making a separate mapper for this (`NodeMultiplicityMapper`?), so `NodeCountMapper` doesn't end up doing too much extra work.
3. `count_duplicates=True` is probably implied for this to be useful, so I think it could be left out of the function signature.

majosm (commented):
Took a quick glance through this; looks good so far! I'll give it a closer look once #507 gets merged.

pytato/analysis/__init__.py
kajalpatelinfo: I agree, this makes the flow cleaner and prevents `NodeCountMapper` from initializing/keeping track of attributes it may potentially never use! Moved into a new Mapper :D

pytato/analysis/__init__.py
```diff
@@ -381,34 +385,104 @@ def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
...
+class NodeMultiplicityMapper(CachedWalkMapper):
+    """
+    Counts the number of unique nodes by ID in a DAG.
+
+    .. attribute:: expr_multiplicity_counts
+    """
+    def __init__(self) -> None:
+        from collections import defaultdict
+        super().__init__()
+        self.expr_multiplicity_counts = defaultdict(int)  # type: Dict[Any, int]
```
inducer, line 464:
```suggestion
        self.expr_multiplicity_counts: Dict[Array, int] = defaultdict(int)  # type: Dict[Any, int]
```
or so?

kajalpatelinfo: Yep, that's accurate! Thank you!

inducer:
@majosm and I discussed this, and we agreed that we'd prefer to merge this and #507 in one go, so I'll go ahead and close #507.

Could you look at the test and linter failures?

kajalpatelinfo:
Sounds good, I'll fix those errors today. Thank you!

test/test_pytato.py
```diff
@@ -586,19 +585,166 @@ def test_repr_array_is_deterministic():
...
+    node_multiplicity = get_node_multiplicities(dag)
+
+    num_duplicates = sum(
+        count - 1 for count in node_multiplicity.values() if count > 1)
+
+    counts = get_node_type_counts(dag, count_duplicates=True)
+    assert len(counts) >= 1
+    assert counts[pt.array.Placeholder] == 1
+    assert counts[pt.array.IndexLambda] == 100  # 100 operations
+    assert sum(counts.values()) == iterations + 1
```
majosm, line 743: Maybe add an assertion to check that there are actually duplicates (something like `assert any(m > 1 for m in node_multiplicities.values())`)?

I think with the way `make_large_dag` is currently written, the probability of getting duplicates is pretty low, so this test might not be checking anything most of the time. Same likely goes for the other duplicate tests, now that I think of it. It may be better to construct a canned DAG that is guaranteed to have duplicates (e.g., `y = 2*x + 2*x`).

majosm (commented):
Mostly looks good, just some suggestions for doc stuff and a couple of the tests.

FYI there will likely be some type annotations to fix up now that https://github.com/inducer/pytato/pull/527 has gone in.

pytato/analysis/__init__.py
```diff
@@ -381,34 +385,105 @@ def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
...
+        from collections import defaultdict
         super().__init__()
-        self.count = 0
+        self.expr_type_counts = defaultdict(int)  # type: Dict[Type[Any], int]
+        self.count_duplicates = count_duplicates
 
-    def get_cache_key(self, expr: ArrayOrNames) -> int:
-        return id(expr)
+    def get_cache_key(self, expr: ArrayOrNames) -> Union[int, ArrayOrNames]:
+        # Returns unique nodes only if count_duplicates is True
```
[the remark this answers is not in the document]
kajalpatelinfo: I think we defined 'unique' different but your comment below about np.unique makes sense, fixed!

pytato/analysis/__init__.py
```diff
@@ -381,34 +385,105 @@ def map_named_call_result(self, expr: NamedCallResult) -> FrozenSet[Array]:
...
+
+    .. attribute:: expr_multiplicity_counts
+    """
+    def __init__(self) -> None:
+        from collections import defaultdict
+        super().__init__()
+        self.expr_multiplicity_counts: Dict[Array, int] = defaultdict(int)
+
+    def get_cache_key(self, expr: ArrayOrNames) -> Union[int, ArrayOrNames]:
+        # Returns unique nodes
```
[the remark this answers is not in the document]
kajalpatelinfo: Got it!

test/test_pytato.py
```diff
@@ -586,19 +585,166 @@ def test_repr_array_is_deterministic():
...
+
+        # Get the number of types of expressions
+        node_count = get_num_nodes(dag, count_duplicates=True)
+
+        # Get the number of expressions and the amount they're called
+        node_multiplicity = get_node_multiplicities(dag)
+
+        # Get difference in duplicates
+        num_duplicates = sum(
+            count - 1 for count in node_multiplicity.values() if count > 1)
```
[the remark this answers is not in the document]
kajalpatelinfo: Yep, it does!

test/test_pytato.py
kajalpatelinfo: You're completely right, I had tested with seeds that had duplicates a portion of the time but definitely not all the time. I replaced the DAG generator previously used with two that intentionally cause duplicates. Thank you!

test/testlib.py
```diff
@@ -327,6 +328,74 @@ def gen_comm(rdagc: RandomDAGContext) -> pt.Array:
...
+        if rng.uniform() > 0.2:
+            dup1 = operation(a, value)
+            dup2 = operation(a, value)
+            duplicates.append(dup1)
+            duplicates.append(dup2)
+            current = operation(current, dup1)
+
+    all_exprs = [current, *duplicates]
+    print(type(duplicates))
+    print(len(duplicates))
```
majosm, line 393:
```suggestion
```

majosm (approved):
👍

majosm:
How's this look @inducer?

pytato/analysis/__init__.py
```diff
@@ -396,34 +400,108 @@ def map_named_call_result(self, expr: NamedCallResult) -> frozenset[Array]:
...
 
-    .. attribute:: count
+    .. attribute:: expr_type_counts
+    .. attribute:: count_duplicates
 
-       The number of nodes.
+       Dictionary mapping node types to number of nodes of that type.
+    """
+
+    def __init__(self, count_duplicates: bool = False) -> None:
```
inducer, line 411: This breaks compatibility (`= True` would retain compatibility).

Maybe deprecate not specifying the argument. (@majosm can explain)

pytato/analysis/__init__.py
```diff
@@ -396,34 +400,108 @@ def map_named_call_result(self, expr: NamedCallResult) -> frozenset[Array]:
 @optimize_mapper(drop_args=True, drop_kwargs=True, inline_get_cache_key=True)
 class NodeCountMapper(CachedWalkMapper):
     """
-    Counts the number of nodes in a DAG.
+    Counts the number of nodes of a given type in a DAG.
 
-    .. attribute:: count
+    .. attribute:: expr_type_counts
+    .. attribute:: count_duplicates
```
inducer, line 406:
```suggestion
    .. autoattribute:: expr_type_counts
    .. autoattribute:: count_duplicates
```

pytato/analysis/__init__.py
```diff
@@ -396,34 +400,108 @@ def map_named_call_result(self, expr: NamedCallResult) -> frozenset[Array]:
...
+def get_num_nodes(
+        outputs: Array | DictOfNamedArrays,
+        count_duplicates: bool = False
+        ) -> int:
     """
+    Returns the number of nodes in DAG *outputs*.
 
+    Instances of `DictOfNamedArrays` are excluded from counting.
+    """
+
```
majosm, line 455:
```suggestion
def get_num_nodes(
        outputs: Array | DictOfNamedArrays,
        count_duplicates: bool | None = None
        ) -> int:
    """
    Returns the number of nodes in DAG *outputs*.

    Instances of `DictOfNamedArrays` are excluded from counting.
    """
    if count_duplicates is None:
        from warnings import warn
        warn(
            "The default value of 'count_duplicates' will change from True to False "
            "in 2025. For now, pass the desired value explicitly.",
            DeprecationWarning, stacklevel=2)
        count_duplicates = True
            
```
Something like this?

kajalpatelinfo: Added the changes! Rerequesting reviews.

pytato/analysis/__init__.py
majosm: I think `NodeCountMapper` is only used by `get_num_nodes`/`get_node_type_counts` (it's not publicly documented, at least). So maybe just the one deprecation warning in `get_num_nodes` is good enough?

majosm (approved):

inducer (commented):
Thanks! One and a half more wrinkles, and this should be ready to go.

test/testlib.py
```diff
@@ -327,6 +328,72 @@ def gen_comm(rdagc: RandomDAGContext) -> pt.Array:
...
+    y = expr1 + expr2
+
+    # Should have duplicates of the 2*x operation
+    return pt.make_dict_of_named_arrays({"result": y})
+
+
+def make_large_dag_with_duplicates(iterations: int,
+                                   seed: int = 0) -> pt.DictOfNamedArrays:
+
+    rng = np.random.default_rng(seed)
```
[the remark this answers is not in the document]
kajalpatelinfo: Done!

inducer:
Thanks!

PR #512, opened 2024-07-02
Misc. fixes from concatenation

majosm:
A few small fixes from the WIP concatenation branch that don't fit anywhere else.

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -614,7 +614,7 @@ class Array(Taggable):
     def __eq__(self, other: Any) -> bool:
         if self is other:
             return True
-        elif not isinstance(other, Array):
+        if not isinstance(other, Array):
             return False
 
         from pytato.equality import EqualityComparer
@@ -913,7 +913,7 @@ class AbstractResultWithNamedArrays(Mapping[str, NamedArray], Taggable, ABC):
     def __eq__(self, other: Any) -> bool:
         if self is other:
             return True
-        elif not isinstance(other, AbstractResultWithNamedArrays):
+        if not isinstance(other, AbstractResultWithNamedArrays):
             return False
 
         from pytato.equality import EqualityComparer
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -614,6 +614,8 @@ class Array(Taggable):
     def __eq__(self, other: Any) -> bool:
         if self is other:
             return True
+        elif not isinstance(other, Array):
+            return False
 
         from pytato.equality import EqualityComparer
         return EqualityComparer()(self, other)
@@ -911,6 +913,8 @@ class AbstractResultWithNamedArrays(Mapping[str, NamedArray], Taggable, ABC):
     def __eq__(self, other: Any) -> bool:
         if self is other:
             return True
+        elif not isinstance(other, AbstractResultWithNamedArrays):
+            return False
 
         from pytato.equality import EqualityComparer
         return EqualityComparer()(self, other)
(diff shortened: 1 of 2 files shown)


pytato/transform/lower_to_index_lambda.py
```diff
@@ -326,7 +326,7 @@ def map_contiguous_advanced_index(self,
                             idx.shape,
                             (1,)*i_adv_indices[0]+adv_idx_shape))
 
-                    if not idx.tags_of_type(AssumeNonNegative):
+                    if axis_len > 1 and not idx.tags_of_type(AssumeNonNegative):
```
inducer, line 329: Could you add some context here? (Maybe in a comment?) Under what circumstances was this relevant?

inducer (commented):
LGTM. Just one question, out of curiosity, mostly.

pytato/array.py
```diff
@@ -613,6 +613,8 @@ def T(self) -> Array:
     def __eq__(self, other: Any) -> bool:
         if self is other:
             return True
+        elif not isinstance(other, Array):
+            return False
```
kaushikcfd, line 617: Minor: this can just be "if". Since there's no else-block in this branching structure.

pytato/transform/lower_to_index_lambda.py
kaushikcfd: I hope this need for simplification arose while we were indexing into a 1-long axes and not because we were allowing building advanced indexing expressions into a 0-long axis. :)

kaushikcfd (approved):
LGTM, thanks!

pytato/transform/lower_to_index_lambda.py
majosm: Hmm. Well, at one point IIRC I was seeing index lambdas with a lot of unnecessary `% 1`s in their `expr`s. But now I'm no longer able to reproduce that. Maybe it was just being triggered by something else that was wrong in an intermediate version of the concatenation code, that has since been fixed? 🤷‍♂️ I've reverted the change for now, I can make a new PR if I run into it again.

pytato/distributed/nodes.py
inducer, line 1: There's a question here that we should answer: Is `DistributedSendRefHolder` allowed to have tags that are independent of the `passthrough_data` array? If yes, then these changes are OK, though we should additionally document this decision. If not, we should avoid inheriting stored `tags` and `non_equality_tags`, and we should ensure that `with_tags` (or so) does the right thing, because every array needs to be able to have its tags changed.

cc @kaushikcfd Any thoughts?

kaushikcfd: > Is DistributedSendRefHolder allowed to have tags that are independent of the passthrough_data array?

Currently, I don't see a need for it.

inducer: To be clear, that makes this a far more intrusive change, since we would have to rip the `tags` and `non_equality_tags` fields out of `Array`. I'm OK with that, and it's probably the better thing to do. But I do want to confirm that that's what you mean before we head down that path.

kaushikcfd: > To be clear, that makes this a far more intrusive change

In that case, I would vote to just take the path of least resistance. We have this issue of NamedArray potentially having different tags than the expression it represents. Maybe we can move all those issues into a single different PR?

inducer: #520 solves this by removing the inherited writable `tags`, `axes`, ... from `Array`. There just have to be attributes.

@majosm, could you remove that change from this PR?

inducer:
Thx!

PR #515, opened 2024-07-12
Avoid duplicating arrays

### LINKED ISSUE #590: Factor out if-same-return-old code into functions
There are lots of repetitions of this type of code all over:

```
        if (
                      tuple(new_args),	            all(
                      expr.redn_axis_to_redn_descr,	                new_arg is arg
                      tags=expr.tags,	                for arg, new_arg in zip(expr.args, new_args, strict=True))
                      axes=expr.axes,)	            and all(
                new_acc_descr is acc_descr
                for acc_descr, new_acc_descr in zip(
                    expr.access_descriptors,
                    new_access_descriptors,
                    strict=True))):
            return expr
        else:
            return Einsum(tuple(new_access_descriptors),
                          tuple(new_args),
                          axes=expr.axes,
                          redn_axis_to_redn_descr=expr.redn_axis_to_redn_descr,
                          tags=expr.tags,
                          non_equality_tags=expr.non_equality_tags)

```
This could be factored into a function per node type, or maybe even done in a dataclass-generic manner.

majosm:
Depends on ~~#503~~ (merged), ~~#512~~ (merged), ~~#531~~ (merged), ~~#549~~ (merged), ~~#585~~ (merged), ~~#583~~ (merged), and ~~#550~~ (merged).

~~@inducer note to self: Reviewed up to (but not including) "tweak Inliner/PlaceholderSubstitutor implementations".~~

Draft because:
- [ ] :warning: Squash "corresponding" commits before merging.

Closes #590.


majosm:
Ready for a look after #585, #583, and #550. I've left the duplication checks turned off by default for now, because some things in arraycontext, grudge, etc. still duplicate.

pytato/array.py
```diff
@@ -467,8 +467,11 @@ class Axis(Taggable):
     tags: frozenset[Tag]
 
     def _with_new_tags(self, tags: frozenset[Tag]) -> Axis:
```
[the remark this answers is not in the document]
majosm: Done in https://github.com/inducer/pytools/pull/293.

pytato/transform/__init__.py
```diff
@@ -561,44 +561,53 @@ def __init__(
...
-                        result_pred is pred
-                        for pred, result_pred in zip(
-                            # type-ignore-reason: mypy doesn't seem to recognize
-                            # overloaded Mapper.__call__ here
-                            pred_getter(inputs.expr),  # type: ignore[arg-type]
-                            pred_getter(result),  # type: ignore[arg-type]
-                            strict=True))):
-                raise MapperCreatedDuplicateError from None
+        try:
+            result = self._result_to_cached_result[result]
```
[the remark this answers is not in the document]
majosm: https://github.com/inducer/pytato/pull/515/commits/e00662b2fd17cdc743fdfb76490c2ddbe4a6256a?

pytato/transform/calls.py
```diff
@@ -75,21 +97,36 @@ class Inliner(CopyMapper):
...
-        new_expr = super().map_call(expr)
-        assert isinstance(new_expr, Call)
+    def __init__(
+            self,
+            _cache: TransformMapperCache[ArrayOrNames, []] | None = None,
+            _function_cache: TransformMapperCache[FunctionDefinition, []] | None = None
+            ) -> None:
+        # Must disable collision/duplication checking because we're combining
+        # expressions that were previously in two different call stack frames
+        # (and were thus cached separately)
```
inducer, line 107: Add to doc: `Inliner` may produce DAGs with duplicates.

pytato/transform/remove_broadcasts_einsum.py
```diff
@@ -97,6 +169,6 @@ def rewrite_einsums_with_no_broadcasts(expr: MappedT) -> MappedT:
         alter its value.
     """
     mapper = EinsumWithNoBroadcastsRewriter()
-    return cast("MappedT", mapper(expr))
+    return cast("MappedT", mapper(expr, None))
```
inducer, line 172:
```
return cast("MappedT", mapper(expr, ())
```
and save the `None` checking above?

pytato/transform/remove_broadcasts_einsum.py
```diff
@@ -28,46 +28,118 @@
...
 
         assert len(new_args) == len(expr.args)
         assert len(new_access_descriptors) == len(expr.access_descriptors)
 
-        return Einsum(tuple(new_access_descriptors),
-                      tuple(new_args),
-                      expr.redn_axis_to_redn_descr,
-                      tags=expr.tags,
-                      axes=expr.axes,)
+        if (
```
inducer, line 125: Maybe tackle #590? (OK if it doesn't pan out.)

inducer (approved):
LGTM!

pytato/transform/calls.py
majosm: Added.

pytato/transform/remove_broadcasts_einsum.py
majosm: Done.

pytato/transform/remove_broadcasts_einsum.py
majosm: Took a stab at it. Seemed to work out OK.

pytato/transform/__init__.py
```diff
@@ -538,6 +542,335 @@ def clone_for_callee(
...
+            a.keys() == b.keys()
+            and all(
+                b_k is a_k
+                for a_k, b_k in zip(
+                    a.values(), b.values(), strict=True)))
+    else:
+        return all(b_i is a_i for a_i, b_i in zip(a, b, strict=True))
+
+
+class IdentityMappingHelperMixin:
```
inducer, line 600: It struck me that it would likely be better to bake these into the node type. (I.e. Placeholder._if_modified(self, other).) We can probably generate them automatically, too. As it stands, dispatch is name-based, even though it's only "single", i.e. based on the type of the node.

majosm: How does 69c4e3c look?

I'm not sure how to make it work for `DictOfNamedArrays`, though. See here:
https://github.com/inducer/pytato/blob/69c4e3cbac40289aaa8cf35960a475fa32793cd2/pytato/transform/__init__.py#L881-L883

inducer: Special-casing `DictOfNamedArrays` is OK, for the reason you mention.

I'm OK with the approach, though I think I'd prefer generated code for this, since it's bound to be a fairly hot code path. There's already some code generation in `@array_dataclass`, it should be straightforward to add there.

inducer: > Edit: I guess this is similar to the issues encountered in #598?

I don't think the errors you're seeing here are all that related. Some just outright don't make any sense, IMO. For example:
```
pytato/transform/__init__.py:1115:25: error: Invalid index type "str" for "Call"; expected type "Call"  [index]
```
I can't make heads or tails of that.

The Pylint issues are similarly weird, maybe it's time to drop Pylint, too?

I think the best proposal I've got is to switch to pyright (the "based" version, to avoid drowning in noise). I've pushed changes that do that. At least most of what pyright has to say seems to make some sense.

inducer: Unfortunately, the "baseline" feature of basedpyright seems a bit imperfect, too. It seems to complain about errors that are already contained in the baseline. I don't see this locally, but it does seem to happen in CI. :thinking:

inducer:
Pre-rebase state is 0302c181db66088609544df39682a008bfaa33d9.

PR #516, opened 2024-07-12
Improve mapper type annotations

inducer:
cc @majosm


majosm:
I thought about this a little more, and it got me wondering: does it actually make sense to require `CopyMapper`'s `rec`/`__call__` to return the same type of thing as it takes in? In other words, couldn't a mapper in theory replace an `Array` with a `DictOfNamedArrays`, and vice versa?

PR #520, opened 2024-07-16
No inherited attrs

inducer:


pytato/array.py
```diff
@@ -800,7 +818,7 @@ class _SuppliedShapeAndDtypeMixin:
 # {{{ dict of named arrays
 
 @attrs.frozen(eq=False, repr=False, hash=True, cache_hash=True)
-class NamedArray(Array):
+class NamedArray(_SuppliedAxesAndTagsMixin, Array):
```
kaushikcfd, line 821: Should this be "_SuppliedAxesAndTagsMixin"?

kaushikcfd (approved):
LGTM!

pytato/array.py
inducer: What do you mean? Are you saying this should not inherit from `Array`?

kaushikcfd: I meant NamedArray should not be taggable, instead it should propagate the tags and axes from the array it points to.

inducer: The difficulty with that is that tags on `NamedArray`s can then no longer be changed (because we would have to change the `DictOfNamedArrays`, but that gets pointed to by potentially a bunch more `NamedArray`s.

Are we expecting that we can usefully call `_with_new_tags` on all `Array`s?

inducer: `DistributedSendRefHolder` also currently can't have its tags changed, but that at least *could* be implemented.

pytato/distributed/nodes.py
```diff
@@ -106,7 +107,7 @@ def copy(self, **kwargs: Any) -> DistributedSend:
 
 # {{{ send ref holder
 
-@attrs.frozen(eq=False, repr=False, init=False, hash=True)
+@attrs.frozen(eq=False, repr=False, hash=True)
```
majosm, line 110:
```suggestion
@attrs.frozen(eq=False, repr=False, hash=True, cache_hash=True)
```
(This is the only `DistributedSendRefHolder` change that would be left in #512 after these changes, might as well do it here.)

inducer: Done, thanks!

PR #521, opened 2024-07-17
change attrs to dataclasses

### LINKED ISSUE #370: Post-review of transition to `attrs`

matthiasdiener:
Closes #370.

TODOs:
- [x] restore validators
- [x] (cached) hashing

_Please squash_


matthiasdiener:
This is ready for a first look.

pytato/array.py
```diff
@@ -427,11 +427,11 @@ class ReductionDescriptor(Taggable):
     tags: FrozenSet[Tag]
 
     def _with_new_tags(self, tags: FrozenSet[Tag]) -> ReductionDescriptor:
-        from attrs import evolve as replace
+        from dataclasses import replace
         return replace(self, tags=tags)
 
 
-@attrs.frozen(eq=False, repr=False, hash=True, cache_hash=True)
+@dataclasses.dataclass(frozen=True, eq=False, repr=False, unsafe_hash=True)
```
inducer, line 434: Where applicable, all array-like dataclasses should use the same, unparametrized decorator.

Example: https://github.com/inducer/pytools/blob/5afdb2d7ea8b40b9ea226393a0e94a61860d83e3/pytools/tag.py#L128-L133

That same decorator should also generate code for a cached hash function. (Maybe this can be factored out into a separate function?) Example: https://github.com/inducer/pymbolic/pull/125/files#diff-768f1abbde39f5f7efff7c209a95315a87e28e0ba4f42e7a468903203aba8ffcR718

pytato/array.py
```diff
@@ -873,8 +926,7 @@ class DictOfNamedArrays(AbstractResultWithNamedArrays):
 
     .. automethod:: __init__
     """
-    _data: Mapping[str, Array] = attrs.field(
-        validator=attrs.validators.instance_of(immutabledict))
+    _data: Mapping[str, Array]
```
matthiasdiener, line 929: I can't get this particular validator to work; why should `_data` be an immutable dict here?

pytato/array.py
```diff
@@ -1124,10 +1120,8 @@ class Einsum(_SuppliedAxesAndTagsMixin, Array):
     access_descriptors: Tuple[Tuple[EinsumAxisDescriptor, ...], ...]
     args: Tuple[Array, ...]
     redn_axis_to_redn_descr: Mapping[EinsumReductionAxis,
-                                     ReductionDescriptor] = \
-        attrs.field(validator=attrs.validators.instance_of(immutabledict))
-    index_to_access_descr: Mapping[str, EinsumAxisDescriptor] = \
-        attrs.field(validator=attrs.validators.instance_of(immutabledict))
```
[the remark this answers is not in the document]
matthiasdiener: Done in 2da059ea

pytato/array.py
matthiasdiener: Done in [f78cdbf](https://github.com/inducer/pytato/pull/521/commits/f78cdbffe031aeede2b250ff0171297a4fe831f4), [4590326](https://github.com/inducer/pytato/pull/521/commits/459032678e94c2f690083b76d0957029740fa30e).

matthiasdiener:
This is ready for review.

inducer:
Thanks!

pytato/loopy.py
```diff
@@ -151,13 +152,12 @@ def __iter__(self) -> Iterator[str]:
         return iter(self._result_names)
 
 
-@attrs.frozen(eq=False, hash=True, cache_hash=True)
+@array_dataclass()
 class LoopyCallResult(NamedArray):
```
matthiasdiener, line 156: Here is a workaround for the mypy error: 

```suggestion
@array_dataclass()
class LoopyCallResult(NamedArray):  # type: ignore[override] # https://github.com/python/mypy/issues/17623
```

inducer: Thanks! I was in the process of falling down the same dumb rabbithole.

PR #523, opened 2024-07-19
CI: test mirgecom example

matthiasdiener:


inducer:
#524 should take care of the mypy warnings.

PR #530, opened 2024-07-25
Add `TransformMapper`/`TransformMapperWithExtraArgs`

majosm:
Adds `TransformMapper`/`TransformMapperWithExtraArgs` which represent generic mappers that map from arrays to other arrays. Separates out the caching/cloning parts of `CopyMapper`/`CopyMapperWithExtraArgs` so that other mappers can reuse them without also having to inherit the default mapper method implementations.

(Possibly easier to read commit by commit, the diff is a little messed up.)

### DIFF
--- a/pytato/transform/__init__.py
+++ b/pytato/transform/__init__.py
@@ -89,6 +89,8 @@ ArrayOrNames = Union[Array, AbstractResultWithNamedArrays]
 MappedT = TypeVar("MappedT",
                   Array, AbstractResultWithNamedArrays, ArrayOrNames)
 CombineT = TypeVar("CombineT")  # used in CombineMapper
+TransformMapperResultT = TypeVar("TransformMapperResultT",  # used in TransformMapper
+                            Array, AbstractResultWithNamedArrays, ArrayOrNames)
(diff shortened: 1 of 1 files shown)


pytato/transform/__init__.py
```diff
@@ -256,6 +260,88 @@ def __call__(self, expr: ArrayOrNames) -> CachedMapperT:
 # }}}
 
 
+# {{{ TransformMapper
+
+class TransformMapper(CachedMapper[ArrayOrNames]):
```
inducer, line 263: What's the difference between a `TransformMapper` and a normal mapper? They look pretty similar. :thinking:

inducer: I.e. take the words from the PR description and add them here.

majosm: How's https://github.com/inducer/pytato/pull/530/commits/aa76efa6bcff11d92d08f4fb847f2508fec5a48e?

inducer: LGTM!

inducer (approved):

inducer:
Thx!

PR #531, opened 2024-07-26
Cache mapped functions

majosm:
Caches the results of mapping `FunctionDefinition`s. Caching is done globally to avoid re-traversing functions in different call stack frames.

Depends on ~~#503~~ (merged) and ~~#530~~ (merged).

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -26,7 +26,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Never
 
 from typing_extensions import Self
 
@@ -318,7 +318,7 @@ def is_einsum_similar_to_subscript(expr: Einsum, subscripts: str) -> bool:
 
 # {{{ DirectPredecessorsGetter
 
-class DirectPredecessorsGetter(Mapper[frozenset[ArrayOrNames], None, []]):
+class DirectPredecessorsGetter(Mapper[frozenset[ArrayOrNames], Never, []]):
     """
     Mapper to get the
     `direct predecessors
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -28,9 +28,10 @@ THE SOFTWARE.
 
 from typing import TYPE_CHECKING, Any
 
+from typing_extensions import Self
+
 from loopy.tools import LoopyKeyBuilder
 from pymbolic.mapper.optimize import optimize_mapper
-from pytools import memoize_method
 
 from pytato.array import (
     Array,
@@ -76,7 +77,7 @@ __doc__ = """
 
 # {{{ NUserCollector
 
-class NUserCollector(Mapper[None, []]):
+class NUserCollector(Mapper[None, None, []]):
     """
     A :class:`pytato.transform.CachedWalkMapper` that records the number of
     times an array expression is a direct dependency of other nodes.
@@ -317,7 +318,7 @@ def is_einsum_similar_to_subscript(expr: Einsum, subscripts: str) -> bool:
 
 # {{{ DirectPredecessorsGetter
 
-class DirectPredecessorsGetter(Mapper[frozenset[ArrayOrNames], []]):
+class DirectPredecessorsGetter(Mapper[frozenset[ArrayOrNames], None, []]):
     """
     Mapper to get the
     `direct predecessors
@@ -413,15 +414,30 @@ class NodeCountMapper(CachedWalkMapper[[]]):
        Dictionary mapping node types to number of nodes of that type.
     """
 
-    def __init__(self, count_duplicates: bool = False) -> None:
+    def __init__(
+            self,
+            count_duplicates: bool = False,
+            _visited_functions: set[Any] | None = None,
+            ) -> None:
+        super().__init__(_visited_functions=_visited_functions)
+
         from collections import defaultdict
-        super().__init__()
         self.expr_type_counts: dict[type[Any], int] = defaultdict(int)
         self.count_duplicates = count_duplicates
 
     def get_cache_key(self, expr: ArrayOrNames) -> int | ArrayOrNames:
         # Returns unique nodes only if count_duplicates is False
         return id(expr) if self.count_duplicates else expr
+
+    def get_function_definition_cache_key(
+            self, expr: FunctionDefinition) -> int | FunctionDefinition:
+        # Returns unique nodes only if count_duplicates is False
+        return id(expr) if self.count_duplicates else expr
+
+    def clone_for_callee(self, function: FunctionDefinition) -> Self:
+        return type(self)(
+            count_duplicates=self.count_duplicates,
+            _visited_functions=self._visited_functions)
 
     def post_visit(self, expr: Any) -> None:
         if not isinstance(expr, DictOfNamedArrays):
@@ -488,12 +504,17 @@ class NodeMultiplicityMapper(CachedWalkMapper[[]]):
 
     .. autoattribute:: expr_multiplicity_counts
     """
-    def __init__(self) -> None:
+    def __init__(self, _visited_functions: set[Any] | None = None) -> None:
+        super().__init__(_visited_functions=_visited_functions)
+
         from collections import defaultdict
-        super().__init__()
         self.expr_multiplicity_counts: dict[Array, int] = defaultdict(int)
 
     def get_cache_key(self, expr: ArrayOrNames) -> int:
+        # Returns each node, including nodes that are duplicates
+        return id(expr)
+
+    def get_function_definition_cache_key(self, expr: FunctionDefinition) -> int:
         # Returns each node, including nodes that are duplicates
         return id(expr)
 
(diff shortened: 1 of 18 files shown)


majosm:
@inducer Should be ready for a look. I think the mypy errors are unrelated, they're occurring on main too.

inducer:
Uh-oh. The recent typing apocalypse has made a bit of a mess of this. I've started resolving conflicts, but I got bogged down in `pytato.transform`. Could you take a look? (Or, if you've lost patience with me, I can also try and finish.)

majosm:
@inducer Should be ready for review again (mypy error looks unrelated?). In e57fe22309fffa47fd86c779830e39b21eaa19bf I added a few questions/comments about things I noticed while resolving conflicts. I'll remove them after you've had a chance to glance at them.

inducer:
https://github.com/inducer/pytato/pull/571 should address the mypy issue.

pytato/transform/__init__.py
```diff
@@ -218,11 +217,25 @@ def rec(self, expr: ArrayOrNames, *args: P.args, **kwargs: P.kwargs) -> ResultT:
...
+    def rec_function_definition(
+            self, expr: FunctionDefinition, *args: P.args, **kwargs: P.kwargs
+            ) -> FunctionResultT:
+        """Call the mapper method of *expr* and return the result."""
+        method: Callable[..., Any] | None
+
+        try:
+            method = self.map_function_definition  # type: ignore[attr-defined]
+        except AttributeError:
+            return cast("FunctionResultT", self.map_foreign(expr, *args, **kwargs))
```
inducer, line 234: Can this be an error instead?

pytato/transform/__init__.py
```diff
@@ -194,7 +193,7 @@ def handle_unsupported_array(self, expr: MappedT,
         raise UnsupportedArrayError(
                 f"{type(self).__name__} cannot handle expressions of type {type(expr)}")
 
-    def map_foreign(self, expr: Any, *args: P.args, **kwargs: P.kwargs) -> ResultT:
+    def map_foreign(self, expr: Any, *args: P.args, **kwargs: P.kwargs) -> Any:
```
inducer, line 196: I think we can get rid of `map_foreign`.

pytato/transform/__init__.py
```diff
@@ -233,23 +246,42 @@ def __call__(self,
...
+            ) -> None:
         super().__init__()
         self._cache: dict[Hashable, ResultT] = {}
 
+        if _function_cache is not None:
+            function_cache = _function_cache
+        else:
+            function_cache = {}
+
+        self._function_cache: dict[Hashable, FunctionResultT] = function_cache
```
inducer, line 273: Can this be an `if` expression?

pytato/transform/__init__.py
```diff
@@ -259,63 +291,68 @@ def rec(self, expr: ArrayOrNames, *args: P.args, **kwargs: P.kwargs) -> ResultT:
...
 
 
 # {{{ TransformMapperWithExtraArgs
 
 class TransformMapperWithExtraArgs(
-            CachedMapper[ArrayOrNames, P],
-            Mapper[ArrayOrNames, P]
+            # Why multiple inheriting?
+            CachedMapper[ArrayOrNames, FunctionDefinition, P],
+            Mapper[ArrayOrNames, FunctionDefinition, P]
```
inducer, line 341:
```suggestion
            CachedMapper[ArrayOrNames, FunctionDefinition, P],
```
I don't think it should be.

pytato/transform/__init__.py
```diff
@@ -259,63 +291,68 @@ def rec(self, expr: ArrayOrNames, *args: P.args, **kwargs: P.kwargs) -> ResultT:
...
     """
     Similar to :class:`TransformMapper`, but each mapper method takes extra
     ``*args``, ``**kwargs`` that are propagated along a path by default.
 
     The logic in :class:`TransformMapper` purposely does not take the extra
     arguments to keep the cost of its each call frame low.
-
-    .. automethod:: clone_for_callee
     """
+    # FIXME: This inflates recursion depth
```
inducer, line 350:
```
if TYPE_CHECKING:
    ... (same)
else:
    @property
    def rec_ary(self):
        return self.rec
```

inducer, line 350:
```
def is_array(expr: ArrayOrNames) -> Array:
    assert isinstance(expr, Array)
    return expr
```
(This is probably the saner way.)

pytato/transform/__init__.py
```diff
@@ -744,6 +780,9 @@ class CombineMapper(Mapper[ResultT, []]):
     def __init__(self) -> None:
         super().__init__()
         self.cache: dict[ArrayOrNames, ResultT] = {}
+        # Don't need to pass function cache as argument here, because unlike
+        # CachedMapper we're not creating a new mapper for each call
+        self.function_cache: dict[FunctionDefinition, FunctionResultT] = {}
```
inducer, line 785: I think the behavior should be consistent across mapper types.

pytato/transform/__init__.py
```diff
@@ -1255,24 +1300,46 @@ class CachedWalkMapper(WalkMapper[P]):
...
         super().__init__()
-        self._visited_nodes: set[Any] = set()
+        self._visited_arrays_or_names: set[Any] = set()
+
+        if _visited_functions is not None:
+            visited_functions = _visited_functions
+        else:
+            visited_functions = set()
+
+        self._visited_functions: set[Any] = visited_functions
```
inducer, line 1314: `if` expr?

pytato/target/loopy/codegen.py
```diff
@@ -384,7 +384,7 @@ def update_t_unit(self, t_unit: lp.TranslationUnit) -> None:
 
 # {{{ codegen mapper
 
-class CodeGenMapper(Mapper[ImplementedResult, [CodeGenState]]):
+class CodeGenMapper(Mapper[ImplementedResult, None, [CodeGenState]]):
```
inducer, line 387:
```suggestion
class CodeGenMapper(Mapper[ImplementedResult, Never, [CodeGenState]]):
```

pytato/transform/__init__.py
majosm: `Reprifier` appears to override `map_foreign` to handle stringifying non-array fields. Seems like it needs to stay?

pytato/transform/__init__.py
majosm: Seems to work fine without it.

pytato/transform/__init__.py
majosm: Added `_function_cache` to the `__init__` args like the others. I didn't go the whole way and add `clone_for_callee`, etc., yet because I'm worried that's going to make a mess of my downstream changes.

pytato/target/loopy/codegen.py
majosm: Added here and in a few other places.

pytato/transform/__init__.py
inducer: Let's move it to `Reprifier` then. I don't think it needs to be a feature of the generic machinery.

pytato/transform/__init__.py
majosm: Something like https://github.com/inducer/pytato/pull/531/commits/ca4f3366c669085b155f11a04f8cd30b07223b45?

(The other way appears to work too: https://github.com/majosm/pytato/commit/34af8ef15414b192c8259ea8b30e18456f4415fd)

pytato/transform/__init__.py
majosm: Done.

pytato/transform/__init__.py
majosm: Done.

majosm:
Ready for another look I think @inducer.

pytato/target/python/numpy_like.py
```diff
@@ -171,7 +172,7 @@ def _is_slice_trivial(slice_: NormalizedSlice,
 }
 
 
-class NumpyCodegenMapper(CachedMapper[str, []]):
+class NumpyCodegenMapper(CachedMapper[str, FunctionDefinition, []]):
```
inducer, line 175: Is this annotation right? Does the numpy target generate appropriate code for functions?

pytato/transform/calls.py
```diff
@@ -58,6 +58,7 @@ class PlaceholderSubstitutor(CopyMapper):
     """
 
     def __init__(self, substitutions: Mapping[str, Array]) -> None:
+        # Ignoring function cache, not needed
```
inducer, line 61: How come? Not using the function cache may lead to large amounts of redundant work?

(I'm OK with not acting on this right now, but maybe this comment should be a FIXME?)

pytato/transform/__init__.py
```diff
@@ -142,13 +139,15 @@
 
     A type variable representing the input type of a :class:`Mapper`.
 
-.. class:: CombineT
+.. class:: ResultT
 
-    A type variable representing the type of a :class:`CombineMapper`.
```
inducer, line 147: Did the `CombineT` docs get deleted on purpose?

inducer (commented):
Thanks! A few questions/comments below.

pytato/transform/__init__.py
majosm: `CombineT` doesn't appear to exist anymore (looks like it got removed in #558).

pytato/target/python/numpy_like.py
majosm: `NumpyCodegenMapper` doesn't support functions, so I guess this should be `Never`. Same for `FancyDotWriter`. Fixed these.

pytato/analysis/__init__.py
```diff
@@ -413,16 +414,31 @@ class NodeCountMapper(CachedWalkMapper[[]]):
        Dictionary mapping node types to number of nodes of that type.
     """
 
-    def __init__(self, count_duplicates: bool = False) -> None:
+    def __init__(
+            self,
+            count_duplicates: bool = False,
+            _visited_functions: set[Any] | None = None,
```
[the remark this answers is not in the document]
majosm: Looks like I did that to match `_visited_arrays_or_names` in `CachedWalkMapper`, which also uses `Any`. I guess I could change both to `Hashable`?

majosm:
I changed `CachedWalkMapper` to use an alias of `Hashable` instead of `Any`, and I removed a few other uses of `Any` that I found. Should be ready for another look.

pytato/transform/__init__.py
inducer: Oh! This is deleting stale docs. Got it, sorry about the noise.

inducer (approved):
Thanks! This looks great. In it goes.

PR #534, opened 2024-08-06
Einsum lowered to IndexLambda variable name fix

nkoskelo:
The mapper to convert an Einsum to an IndexLambda we do not use the reserved variable names for the bindings. This PR intends to fix that.


inducer:
Good catch, thanks! All those mypy failures are not your fault, fwiw. That's due to https://github.com/inducer/loopy/commit/4009eee43651ee3009e4a774600d48d06d6a01c2, which allowed pytato to actually see loopy's types. I've got changes locally that fix these, but those in turn rely on https://github.com/inducer/loopy/pull/785. :thinking:

PR #535, opened 2024-08-14
Add edge counter functionality and tests 

kajalpatelinfo:
This PR adds functionality to count the number of edges in a DAG, with or without duplicates. It also adds tests for these functionalities.


pytato/analysis/__init__.py
```diff
@@ -515,6 +524,118 @@ def get_node_multiplicities(
...
+
+    def get_dependencies(self, expr: Any) -> frozenset[Any]:
+        # Retrieve dependencies based on the type of the expression
+        if hasattr(expr, "bindings") or isinstance(expr, IndexLambda):
+            return frozenset(expr.bindings.values())
+        elif isinstance(expr, (BasicIndex, Reshape, AxisPermutation)):
+            return frozenset([expr.array])
+        elif isinstance(expr, Einsum):
+            return frozenset(expr.args)
+        return frozenset()
```
majosm, line 620: (Same deal here with `DirectPredecessorsGetter`.)

test/test_pytato.py
```diff
@@ -765,6 +765,126 @@ def test_large_dag_with_duplicates_count():
         dag, count_duplicates=False)
 
 
+def test_empty_dag_edge_count():
+    from pytato.analysis import get_edge_multiplicities, get_num_edges
+
+    empty_dag = pt.make_dict_of_named_arrays({})
+
+    # Verify that get_num_edges returns 0 for an empty DAG
+    assert get_num_edges(empty_dag, count_duplicates=False) == 0
```
majosm, line 774:
```suggestion
    assert get_num_edges(empty_dag) == 0
```
(And same for all the rest.)

majosm (commented):
Thanks! I added a few suggestions.

Also, there's an edge (ha) case that we need to think about here: what happens if the same array appears twice as a dependency? For example, `y = x + x`; should there be one edge between `x` and `y`, or two? The behavior should ideally match whatever the visualization does (I don't remember offhand if it emits a single edge or multiple). If it needs to be multiple edges, I think one way to do that would be to tweak `DirectPredecessorsGetter` to return either a `list` or a `frozenset` based on an argument to `__init__`.

PR #539, opened 2024-08-21
Add compression for IndexLambda operations in DAG

kajalpatelinfo:
This PR adds functionality that compresses IndexLambda nodes into just the symbol of the operation it represents.

image


pytato/visualization/dot.py
```diff
@@ -88,6 +86,84 @@ class _SubgraphTree:
     subgraphs: dict[str, _SubgraphTree]
 
 
+def extract_operation_symbol(expr):
+
+    operation_replacements = {
```
kaushikcfd, line 91: We could potentially use https://github.com/inducer/pytato/blob/main/pytato/raising.py#L255 for this.

kaushikcfd:
New visualization looks quite nice! 😃

PR #538, opened 2024-08-20
Revert "Implement numpy 2 type promotion"

matthiasdiener:
This reverts commit f0deaea1130fe0dd4dfa52bee540910305de9c70.

cc #510.

PR #540, opened 2024-08-23
use numpy type promotion directly

matthiasdiener:
(Note that this doesn't address the slowdown we have seen in #538)

_Please squash_

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -306,87 +306,10 @@ DtypeOrPyScalarType = Union[_dtype_any, PyScalarType]
 ArrayOrScalar = Union["Array", Scalar]
 
 
-# https://numpy.org/neps/nep-0050-scalar-promotion.html
-class DtypeKindCategory(IntEnum):
-    BOOLEAN = 0
-    INTEGRAL = 1
-    INEXACT = 2
-
-
-_dtype_kind_char_to_kind_cat = {
-    "b": DtypeKindCategory.BOOLEAN,
-    "i": DtypeKindCategory.INTEGRAL,
-    "u": DtypeKindCategory.INTEGRAL,
-    "f": DtypeKindCategory.INEXACT,
-    "c": DtypeKindCategory.INEXACT,
-}
-
-
-_py_type_to_kind_cat = {
-    bool: DtypeKindCategory.BOOLEAN,
-    int: DtypeKindCategory.INTEGRAL,
-    float: DtypeKindCategory.INEXACT,
-    complex: DtypeKindCategory.INEXACT,
-}
-
-
-_float_dtype_to_complex: dict[np.dtype[Any], np.dtype[Any]] = {
-    np.dtype(np.float32): np.dtype(np.complex64),
-    np.dtype(np.float64): np.dtype(np.complex128),
-}
-
-
-def _complexify_dtype(dtype: np.dtype[Any]) -> np.dtype[Any]:
-    if dtype.kind == "c":
-        return dtype
-    elif dtype.kind == "f":
-        return _float_dtype_to_complex[dtype]
-    else:
-        raise ValueError("can only complexify types that are already inexact")
-
-
-def _np_result_dtype(*dtypes: DtypeOrPyScalarType) -> np.dtype[Any]:
-    # For numpy 2.0, np.result_type does not implement numpy's type
-    # promotion behavior. Weird. Hence all this nonsense is needed.
-
-    py_types = [dtype for dtype in dtypes if isinstance(dtype, type)]
-
-    if not py_types:
-        return np.result_type(*dtypes)
-
-    np_dtypes = [dtype for dtype in dtypes if isinstance(dtype, np.dtype)]
-    np_kind_cats = {
-        _dtype_kind_char_to_kind_cat[dtype.kind] for dtype in np_dtypes}
-    py_kind_cats = {_py_type_to_kind_cat[tp] for tp in py_types}
-    kind_cats = np_kind_cats | py_kind_cats
-
-    res_kind_cat = max(kind_cats)
-    max_py_kind_cats = max(py_kind_cats)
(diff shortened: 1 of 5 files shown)


pytato/array.py
```diff
@@ -306,87 +306,10 @@ def normalize_shape_component(
...
-        return dtype
-    elif dtype.kind == "f":
-        return _float_dtype_to_complex[dtype]
-    else:
-        raise ValueError("can only complexify types that are already inexact")
-
-
-def _np_result_dtype(*dtypes: DtypeOrPyScalarType) -> np.dtype[Any]:
-    # For numpy 2.0, np.result_type does not implement numpy's type
-    # promotion behavior. Weird. Hence all this nonsense is needed.
```
inducer, line 350: Instead of a review, I'll point you to this bit here.

matthiasdiener: I saw the comment, but couldn't find an example where its behavior doesn't match the promotion rules. Do you happen to have an example that demonstrates this? The relevant tests seem to work (the pytest errors should be fixed by #537).

inducer: Upon further discussion today, I think the comment (in the code) is in error, as evidenced by the fact that it's still passing the tests when making this change. What the code comment is complaining about is this (perceived?) inconsistency:
```
>>> np.result_type(int, np.float32)
dtype('float64')
>>> np.result_type(7, np.float32)
dtype('float32')
```

inducer: @matthiasdiener (in chat) just mentioned that this may be deliberate (on numpy's part), since `int` may get fed through `np.dtype` before processing, where:
```
>>> np.dtype(int)
dtype('int64')
```

pytato/utils.py
```diff
@@ -221,9 +205,7 @@ def broadcast_binary_op(a1: ArrayOrScalar, a2: ArrayOrScalar,
         return evaluate(op(a1, a2))  # type: ignore
 
     result_shape = get_shape_after_broadcasting([a1, a2])
-
-    dtypes = _extract_dtypes([a1, a2])
-    result_dtype = get_result_type(*dtypes)
+    result_dtype = get_result_type(a1, a2)
```
inducer, line 208: It is a little weird that we're passing pytato arrays to numpy here. Since `np.array(pytato_array)` fails, we can be relatively sure that it's not doing anything stupid (like evaluating the array). I think I'd be OK leaving this as is as long as there's a comment flagging that that's what we're doing, and replicating some of the details from this PR discussion.

inducer:
I think we should do this.

pytato/utils.py
matthiasdiener: I looked at the relevant numpy code and did a trace through the execution when running this - as far as I can tell, numpy only accesses `.dtype` of arrays that get passed to `np.result_type`.

pytato/array.py
matthiasdiener: I agree that this looks inconsistent, but I think it matches the documentation in [migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion), [NEP50](https://numpy.org/neps/nep-0050-scalar-promotion.html):

> Python scalars will be considered “weakly” typed. When a NumPy array/scalar is combined with a Python scalar, it will be converted to the NumPy dtype, such that:
>```python
> np.array([1, 2, 3], dtype=np.uint8) + 1  # returns a uint8 array
> np.array([1, 2, 3], dtype=np.float32) + 2.  # returns a float32 array
>```

This explains (I *think*) why:
```python
>>> np.result_type(7, np.float32)
dtype('float32')
```

For the other case, I could not find a direct example/explanation in the NEP50 docs, but I think it is the conversion to `int64` from `int` that @inducer mentioned:

```python
>>> np.result_type(int, np.float32)
dtype('float64')
```

Perhaps we could suggest an addition to the numpy docs with this.

inducer: :+1: on asking numpy to document the behavior.

pytato/utils.py
matthiasdiener: Sure, I agree in adding a comment. We could also add a brief test to make sure that nothing else gets accessed, maybe something like:

```python
import numpy as np

class Foo:
    @property
    def dtype(self):
        return np.dtype("float64")

assert str(np.result_type(42, Foo())) == "float64"
```

inducer: I like that idea.

matthiasdiener: Done in [89a6a43](https://github.com/inducer/pytato/pull/540/commits/89a6a439165219362db993241fc4581cf13f1d47)

matthiasdiener:
This is ready for review. I tested this PR manually with numpy 1.26, and it seems to work fine. I don't know if there are more things that need to be reverted from #510.

test/test_codegen.py
```diff
@@ -319,18 +316,9 @@ def test_scalar_array_binary_arith(ctx_factory, which, reverse):
             out = outputs[dtype]
             out_ref = np_op(x_in, y_orig.astype(dtype))
 
-            if not is_old_numpy:
-                assert out.dtype == out_ref.dtype, (out.dtype, out_ref.dtype)
-
+            assert out.dtype == out_ref.dtype, (out.dtype, out_ref.dtype)
             # In some cases ops are done in float32 in loopy but float64 in numpy.
```
matthiasdiener, line 320: Is this comment still accurate?

inducer: The comment is not precise enough to falsify, so :shrug: . Fine to remove.

In [its original form](https://github.com/inducer/pytato/commit/870842e65af7de9ae23c870f860fa0aa0a117629), it applied to `__truediv__` only.

pytato/array.py
matthiasdiener: I started a discussion here: https://github.com/numpy/numpy/issues/27325

inducer:
Thx!

Issue #542, opened 2024-09-10
Type casting of binary op arguments causes kernel slowdowns


matthiasdiener:
Contrary to what I speculated in #538, the cause of our slowdowns in mirgecom is not the numpy type promotion itself, but the casting of operands for binary ops: https://github.com/inducer/pytato/blob/3fedafc22829365e73aa5afbdf2d0061a19a574a/pytato/utils.py#L235-L254

As an example, consider:

```python
import numpy as np
import pytato as pt
import pyopencl as cl

a = np.array([1, 2, 3], dtype=np.uint32)
b = np.int32(7)

res = pt.make_data_wrapper(a) - b

ctx = cl.create_some_context()
queue = cl.CommandQueue(ctx)

prg = pt.generate_loopy(res)

print(prg.program)

_, out = prg(queue)

print("pytato", out[0], out[0].dtype)
print("numpy", a - b, (a-b).dtype)
```

With current pytato `main`, this generates:

```
---------------------------------------------------------------------------
KERNEL: _pt_kernel
---------------------------------------------------------------------------
ARGUMENTS:
_pt_data: type: np:dtype('uint32'), shape: (3), dim_tags: (N0:stride:1), offset: <class 'loopy.typing.auto'> in aspace: global
_pt_out: type: np:dtype('int64'), shape: (3), dim_tags: (N0:stride:1) out aspace: global
---------------------------------------------------------------------------
DOMAINS:
{  :  }
{ [_pt_out_dim0] : 0 <= _pt_out_dim0 <= 2 }
---------------------------------------------------------------------------
INAME TAGS:
_pt_out_dim0: None
---------------------------------------------------------------------------
INSTRUCTIONS:
for _pt_out_dim0
  _pt_out[_pt_out_dim0] = cast(np:dtype('int64'), _pt_data[_pt_out_dim0]) + (-1)*cast(np:dtype('int64'), 7)  {id=_pt_out_store}
end _pt_out_dim0
---------------------------------------------------------------------------

pytato [-6 -5 -4] int64
numpy [-6 -5 -4] int64
```

With the argument casting disabled (line 250 above), this generates:

```
---------------------------------------------------------------------------
KERNEL: _pt_kernel
---------------------------------------------------------------------------
ARGUMENTS:
_pt_data: type: np:dtype('uint32'), shape: (3), dim_tags: (N0:stride:1), offset: <class 'loopy.typing.auto'> in aspace: global
[...]

PR #543, opened 2024-09-11
Disable type casting for pow() int arguments

matthiasdiener:
Fixes (works around?) #542.

cc #538 https://github.com/illinois-ceesd/mirgecom/pull/1055

_Please squash_

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -582,6 +582,7 @@ class Array(Taggable):
                         np.dtype[Any]] = _np_result_dtype,
                 reverse: bool = False,
                 cast_to_result_dtype: bool = True,
+                is_pow: bool = False,
             ) -> Array:
 
         # {{{ sanity checks
@@ -601,14 +602,16 @@ class Array(Taggable):
                          get_result_type,
                          tags=tags,
                          non_equality_tags=non_equality_tags,
-                         cast_to_result_dtype=cast_to_result_dtype)
+                         cast_to_result_dtype=cast_to_result_dtype,
+                         is_pow=is_pow)
         else:
             result = utils.broadcast_binary_op(
                          self, other, op,
                          get_result_type,
                          tags=tags,
                          non_equality_tags=non_equality_tags,
-                         cast_to_result_dtype=cast_to_result_dtype)
+                         cast_to_result_dtype=cast_to_result_dtype,
+                         is_pow=is_pow)
 
         assert isinstance(result, Array)
         return result
@@ -648,8 +651,8 @@ class Array(Taggable):
     __rtruediv__ = partialmethod(_binary_op, operator.truediv,
             get_result_type=_truediv_result_type, reverse=True)
 
-    __pow__ = partialmethod(_binary_op, operator.pow)
-    __rpow__ = partialmethod(_binary_op, operator.pow, reverse=True)
+    __pow__ = partialmethod(_binary_op, operator.pow, is_pow=True)
+    __rpow__ = partialmethod(_binary_op, operator.pow, reverse=True, is_pow=True)
 
     __neg__ = partialmethod(_unary_op, operator.neg)
 
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -2406,7 +2406,8 @@ def _compare(x1: ArrayOrScalar, x2: ArrayOrScalar, which: str) -> Array | bool:
                             lambda x, y: np.dtype(np.bool_),
                             tags=_get_default_tags(),
                             non_equality_tags=_get_created_at_tag(stacklevel=2),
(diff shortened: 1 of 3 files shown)


kaushikcfd:
Interesting and very strange 😅. Thanks for the diffs. Is the performance dip consistent with different opencl implementations.

matthiasdiener:
> Is the performance dip consistent with different opencl implementations.

Yes, somewhat. With pocl-cuda, there is a ~10%-20% slowdown (see e.g. https://github.com/illinois-ceesd/mirgecom/pull/1055#issuecomment-2307740309). I haven't tried other CL implementations yet.

matthiasdiener:
This is quite messy, but I think it is ready for a first look @inducer @kaushikcfd

kaushikcfd:
LGTM. I don't mind it. (The timings are something of a puzzle... but these changes don't seem to be breaking the CI, so should be fine :))

pytato/utils.py
```diff
@@ -227,7 +227,13 @@ def cast_to_result_type(
 
             expr = TypeCast(result_dtype, expr)
         elif isinstance(expr, SCALAR_CLASSES):
-            expr = result_dtype.type(expr)
+            import operator
+            # See https://github.com/inducer/pytato/issues/542
+            # on why pow() + integers is not typecast to float or complex.
+            if not ((op == prim.Power or op == operator.pow)
```
inducer, line 233: I'm not loving that we're identifying the operator by its callable; that's brittle. (Somebody could just wrap things in a `lambda` and this would forget what it was doing.)

pytato/utils.py
inducer, line 1: Could you add a test?

matthiasdiener: There is a test in `test_codegen.py`, is there another test you would like?

pytato/utils.py
matthiasdiener: I understand, but I'm not sure how to improve this. Do you have a suggestion?

test/test_codegen.py
```diff
@@ -2008,6 +2005,27 @@ def call_bar(tracer, x, y):
...
+    # Check that pow() arguments are not typecast from int
+    ctx = ctx_factory()
+    cq = cl.CommandQueue(ctx)
+
+    x_np = np.random.rand(10, 4)
+    x = pt.make_data_wrapper(x_np)
+
+    out = x ** -2
+    _, (pt_out,) = pt.generate_loopy(out)(cq)
+    assert isinstance(out.expr.exponent, int)
```
matthiasdiener, line 2018: Here it checks that the exponent is still of type `int`.

pytato/utils.py
```diff
@@ -227,7 +227,13 @@ def cast_to_result_type(
...
+                    and np.issubdtype(type(expr), np.integer)
```
kaushikcfd, line 234: Would the casts be present if the exponent was a pytato array with integer-dtyped exponent? If yes, we should probably avoid them. (Or acknowledge them as a fixme)

pytato/utils.py
inducer: I had just overlooked the test wholesale. Sorry!

pytato/utils.py
matthiasdiener: Do you mean this scenario:
```python
3** pt.make_data_wrapper(np.array([1,2,3], np.int32))
```
?

kaushikcfd: Yep!

matthiasdiener: What do you think of [4daf10](https://github.com/inducer/pytato/pull/543/commits/e4daf10e8d9f4598c2312eed496bdb81ab0793dd)?

pytato/array.py
```diff
@@ -2403,7 +2406,8 @@ def _compare(x1: ArrayOrScalar, x2: ArrayOrScalar, which: str) -> Array | bool:
                             lambda x, y: np.dtype(np.bool_),
                             tags=_get_default_tags(),
                             non_equality_tags=_get_created_at_tag(stacklevel=2),
-                            cast_to_result_dtype=False
+                            cast_to_result_dtype=False,
+                            is_pow=False,
```
inducer, line 2410: Why not rely on the default for this (and the following two)?

matthiasdiener: `broadcast_binary_op` does not have a default (I kept the structure similar to the `cast_to_result_dtype` argument), but I can add a default if you like.

inducer: Ah, nvm.

pytato/utils.py
kaushikcfd: Looks good. Thanks!

kaushikcfd (approved):
Thanks!

PR #546, opened 2024-09-19
[RFC] EqualityComparer: use data for DataWrapper comparison

matthiasdiener:
Something like this is necessary to compare pickled/unpickled DataWrappers meaningfully.

I suspect this implementation is *slightly* controversial, but perhaps we could find a better way to implement this? (E.g., maybe calculating some hash over the data on DataWrapper construction?)


inducer:
(see discussion in #547)

Issue #547, opened 2024-09-21
Persistent hashing for DAGs


inducer:
This is a follow-up to #546. I sympathize with wanting to hash DAGs persistently, I don't see a good reason why we shouldn't. At the same time, needing to hash data that physically sits on a GPU is weird and fraught with all sorts of complication, as #546 demonstrates.

For a moment, I was tempted to say that we should insist that `DataWrapper`s should only contain data on the host, but that's nonsense, because it'll destroy any type of eager-ish `thaw`/`freeze` operation by overwhelming it with data transfer cost of questionable utility. (How to cache those compiles is a separate question; let's have that discussion but maybe not now.)

The main place in which this hashing is relevant is pytato-to-loopy transforms/code generation. I previously assumed this was quick, but recent data from @matthiasdiener and @majosm perhaps suggests otherwise.

So why don't we just introduce a small mapper that transfers all the array data to the host before transforms/codegen and another that transfers them back after? Then, at least in between those two transfers, we can use all the hashing we need.

cc @matthiasdiener @kaushikcfd

matthiasdiener:
Now that #459 and https://github.com/inducer/arraycontext/pull/282 have been merged, is there anything else that needs to be done here?

inducer:
Nope, I think we're done here. Thanks all!

PR #548, opened 2024-09-24
add a CL device<->host transfer mapper

matthiasdiener:
Implements part of #547

TODOs:
- [ ] add to docs
- [ ] add tests
- [ ] check TaggableCLArray support

### DIFF
--- a/pytato/transform/__init__.py
+++ b/pytato/transform/__init__.py
@@ -1951,4 +1951,52 @@ def deduplicate_data_wrappers(array_or_names: ArrayOrNames) -> ArrayOrNames:
 
 # }}}
 
+
+# {{{ TransferMapper
+
+class TransferMapper(CopyMapper):
+    def __init__(self, to_device: bool, queue: Any, allocator: Any = None) -> None:
+        super().__init__()
+        self.to_device = to_device
+        self.queue = queue
+        self.allocator = allocator
+
+    def map_data_wrapper(self, expr: DataWrapper) -> Array:
+        import sys
+        if "pyopencl" not in sys.modules:
+            return super().map_data_wrapper(expr)
+
+        from pyopencl.array import Array as CLArray, to_device
+
+        if isinstance(expr.data, CLArray) and not self.to_device:
+            data = expr.data.get()
+            return DataWrapper(
+                data=data,
+                shape=expr.shape,
+                axes=expr.axes,
+                tags=expr.tags,
+                non_equality_tags=expr.non_equality_tags)
+        elif isinstance(expr.data, np.ndarray) and self.to_device:
+            data = to_device(self.queue, expr.data, allocator=self.allocator)
+            return DataWrapper(
+                data=data,
+                shape=expr.shape,
+                axes=expr.axes,
+                tags=expr.tags,
+                non_equality_tags=expr.non_equality_tags)
+
+        return super().map_data_wrapper(expr)
+
+
+def transfer_to_device(expr: ArrayOrNames, queue: Any,
+                       allocator: Any = None) -> ArrayOrNames:
+    return TransferMapper(True, queue, allocator)(expr)
+
+
+def transfer_to_host(expr: ArrayOrNames, queue: Any,
+                     allocator: Any = None) -> ArrayOrNames:
+    return TransferMapper(False, queue, allocator)(expr)
+
+# }}}
+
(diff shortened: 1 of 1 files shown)


matthiasdiener:
Is this going in the direction you had in mind @inducer ?

pytato/transform/__init__.py
```diff
@@ -1951,4 +1951,52 @@ def cached_data_wrapper_if_present(ary: ArrayOrNames) -> ArrayOrNames:
 
 # }}}
 
+
+# {{{ TransferMapper
+
+class TransferMapper(CopyMapper):
+    def __init__(self, to_device: bool, queue: Any, allocator: Any = None) -> None:
+        super().__init__()
+        self.to_device = to_device
```
inducer, line 1960: I think I would prefer two different mappers, one per direction.

pytato/transform/__init__.py
```diff
@@ -1951,4 +1951,52 @@ def cached_data_wrapper_if_present(ary: ArrayOrNames) -> ArrayOrNames:
 
 # }}}
 
+
+# {{{ TransferMapper
+
+class TransferMapper(CopyMapper):
```
inducer, line 1957: I think this guy/these guys would be better off in arraycontext.

pytato/transform/__init__.py
```diff
@@ -1951,4 +1951,52 @@ def cached_data_wrapper_if_present(ary: ArrayOrNames) -> ArrayOrNames:
...
+        self.allocator = allocator
+
+    def map_data_wrapper(self, expr: DataWrapper) -> Array:
+        import sys
+        if "pyopencl" not in sys.modules:
+            return super().map_data_wrapper(expr)
+
+        from pyopencl.array import Array as CLArray, to_device
+
+        if isinstance(expr.data, CLArray) and not self.to_device:
```
inducer, line 1971: I think the to-host version of this should error if a non-numpy array is encountered.

inducer (commented):
Generally looks good, two minor nits.

pytato/transform/__init__.py
majosm: Would the idea be to use `to_numpy`/`from_numpy`  instead of `.data.get` then?

pytato/transform/__init__.py
majosm: Non-numpy or non-CL?

pytato/transform/__init__.py
```diff
@@ -1951,4 +1951,52 @@ def cached_data_wrapper_if_present(ary: ArrayOrNames) -> ArrayOrNames:
...
+            data = expr.data.get()
+            return DataWrapper(
+                data=data,
+                shape=expr.shape,
+                axes=expr.axes,
+                tags=expr.tags,
+                non_equality_tags=expr.non_equality_tags)
```
majosm, line 1978: Can we make this return an instance of a new `DataWrapper` type that can `__hash__` by data instead of `id`? That would remove the need for data wrapper deduplication, if I understand the intended use of these mappers correctly.

pytato/transform/__init__.py
matthiasdiener: Wouldn't it be worthwhile to support a scenario such as `transfer_to_host(transfer_to_host(foo_dag))`?

pytato/transform/__init__.py
inducer: Yeah.

pytato/transform/__init__.py
inducer: :thinking: Not sure.

Off the bat: maybe not?

pytato/transform/__init__.py
matthiasdiener: Done in https://github.com/inducer/arraycontext/pull/282

pytato/transform/__init__.py
matthiasdiener: removed support for this scenario in https://github.com/inducer/arraycontext/pull/282

pytato/transform/__init__.py
matthiasdiener: Done in https://github.com/inducer/arraycontext/pull/282

matthiasdiener:
Closing here, will be continued in https://github.com/inducer/arraycontext/pull/282

PR #550, opened 2024-09-24
Add duplication checks

majosm:
Adds two checks to `CachedMapper`-derived mapper class traversals:

1. `err_on_collision` in `CachedMapper`, which checks for multiple distinct instances of input expressions that have the same key. These are duplicates in the incoming DAG.
2. `err_on_no_op_duplication` in `TransformMapper`/`TransformMapperWithExtraArgs`, which checks for mappings between two distinct expression instances where both the input and the output have the same key. These are duplicates that are produced by the mapper (e.g., the current `CopyMapper.map_*` implementations that create new instances even if nothing has changed).

Depends on #531 and #549.


pytato/transform/__init__.py
```diff
@@ -385,70 +415,172 @@ class CachedMapper(Mapper[ResultT, FunctionResultT, P]):
...
+
+    def add(
+            self,
+            inputs: CacheInputs[CacheExprT, P],
+            result: CacheExprT) -> CacheExprT:
+        """
+        Cache a mapping result.
+
+        Returns the cached result (which may not be identical to *result* if a
+        result was already cached with the same result key).
```
inducer, line 556: Not true yet.

pytato/transform/__init__.py
```diff
@@ -187,6 +189,14 @@ class ForeignObjectError(ValueError):
     pass
 
 
+class CacheCollisionError(ValueError):
+    pass
+
+
+class CacheNoOpDuplicationError(ValueError):
```
inducer, line 196:
```suggestion
class MapperCreatedDuplicateError(ValueError):
```

pytato/transform/__init__.py
```diff
@@ -385,70 +415,172 @@ class CachedMapper(Mapper[ResultT, FunctionResultT, P]):
...
+            from pytato.analysis import DirectPredecessorsGetter
+            pred_getter = DirectPredecessorsGetter(include_functions=True)
+            if (
+                    hash(result) == hash(inputs.expr)
+                    and result == inputs.expr
+                    and result is not inputs.expr
+                    # Need this check in order to handle input DAGs that have existing
+                    # duplicates. Deduplication will potentially replace predecessors
+                    # of `expr` with cached versions, producing a new `result` that has
+                    # the same cache key as `expr`.
```
inducer, line 570:
```suggestion
                     # Heuristic: Don't complain if predecessors are not the same objects.
                     # ???
                     # Need this check in order to handle input DAGs that have existing
                    # duplicates. Deduplication will potentially replace predecessors
                    # of `expr` with cached versions, producing a new `result` that has
                    # the same cache key as `expr`.
```
Doesn't this also disable some *true* positives?

Answer: it doesn't because each chain of duplicates would have a root, and the error would fire for that.

pytato/transform/__init__.py
majosm: Removed.

pytato/transform/__init__.py
majosm: Done.

majosm:
Ready for a look after #585 and #583.

majosm:
@inducer This one should be good to go now.

PR #555, opened 2024-10-28
Any reductions in the scalar expression?

nkoskelo:
Add a function ```contains_reduction(expr)``` which returns True if any operation within the scalar expression, ```expr```, is a reduction operation.

### DIFF
--- a/pytato/scalar_expr.py
+++ b/pytato/scalar_expr.py
@@ -349,5 +349,5 @@ def contains_reduction(expr: prim.Expression) -> bool:
     Returns true if any operation in the scalar expression, expr, is a reduction
     operation.
     """
-    return (len(get_reduction_induction_variables(expr)) > 0)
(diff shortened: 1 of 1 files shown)


pytato/scalar_expr.py
```diff
@@ -340,4 +343,11 @@ def get_reduction_induction_variables(expr: prim.Expression) -> frozenset[str]:
     """
     return InductionVariableCollector()(expr)  # type: ignore[no-any-return]
 
+
+def contains_reduction(expr: prim.Expression) -> bool:
+    """
+    Returns true if any operation in the scalar expression, expr, is a reduction
+    operation.
+    """
+    return (len(get_reduction_induction_variables(expr)) > 0)
```
inducer, line 352:
```suggestion
    return bool(get_reduction_induction_variables(expr))
```

PR #556, opened 2024-10-28
Revert "Any reductions in the scalar expression?"

inducer:
Reverts inducer/pytato#555

It turns out this is not needed, hence I'm reverting this. A simpler way to check whether a node contains a reduction is to check https://documen.tician.de/pytato/array.html#pytato.array.IndexLambda.var_to_reduction_descr for non-emptiness.

cc @nkoskelo


inducer:
Oh whoops. There was that `map_type_cast` in there that was probably worth having. :facepalm: I'll add that to my working copy, it'll come back before too long.

PR #557, opened 2024-11-04
Move the AxesEquationCollector to not use the raising.py operations.

nkoskelo:
Update metadata propagation to not use the `index_lambda_to_high_level_op` function but instead look at the object's metadata.


nkoskelo:
@inducer gentle request for review.

inducer:
Could you look over those mypy failures?

PR #563, opened 2024-11-21
do not pickle cached hash of arrays

matthiasdiener:
Prevents errors of the type:

`
pytato.distributed.verify.MissingSendError: no send for 'CommunicationOpIdentifier(src_rank=1, dest_rank=0, comm_tag=('fluid', (<class 'mirgecom.diffusion._DiffusionStateTag'>, (<class 'y3prediction.prediction._SmoothCharDiffFluidCommTag'>, Placeholder(shape=(), dtype='int64', name='_actx_in_1'))), (0,)))'
`

 (see e.g. https://github.com/illinois-ceesd/mirgecom/actions/runs/11958222838/job/33337251412?pr=926)

cc @MTCam

Followup of #521.

_Please squash_

### DIFF
--- a/pytato/array.py
+++ b/pytato/array.py
@@ -336,15 +336,34 @@ def _augment_array_dataclass(
             cls: type,
             generate_hash: bool,
         ) -> None:
-    from dataclasses import fields
-    attr_tuple = ", ".join(f"self.{fld.name}"
-                           for fld in fields(cls) if fld.name != "non_equality_tags")
-    if attr_tuple:
-        attr_tuple = f"({attr_tuple},)"
-    else:
-        attr_tuple = "()"
+
+    # {{{ hashing and hash caching
 
     if generate_hash:
+        from dataclasses import fields
+
+        attr_tuple = ", ".join(f"self.{fld.name}" for fld in fields(cls))
+        if attr_tuple:
+            attr_tuple = f"({attr_tuple},)"
+        else:
+            attr_tuple = "()"
+
+        fld_name_tuple = ", ".join(f"'{fld.name}'" for fld in fields(cls))
+        if fld_name_tuple:
+            fld_name_tuple = f"({fld_name_tuple},)"
+        else:
+            fld_name_tuple = "()"
+
+        # Non-equality tags are automatically excluded from equality in
+        # EqualityComparer, and are excluded here from hashing.
+        attr_tuple_hash = ", ".join(f"self.{fld.name}"
+                            for fld in fields(cls) if fld.name != "non_equality_tags")
+
+        if attr_tuple_hash:
+            attr_tuple_hash = f"({attr_tuple_hash},)"
+        else:
+            attr_tuple_hash = "()"
+
         from pytools.codegen import remove_common_indentation
         augment_code = remove_common_indentation(
             f"""
@@ -354,16 +373,25 @@ def _augment_array_dataclass(
                 except AttributeError:
                     pass
 
-                h = hash(frozenset({attr_tuple}))
+                h = hash(frozenset({attr_tuple_hash}))
                 object.__setattr__(self, "_hash_value", h)
                 return h
 
             cls.__hash__ = {cls.__name__}_hash
+
+            def {cls.__name__}_getstate(self):
+                # This must not return the cached hash value.
+                return {{name: val
+                for name, val in zip({fld_name_tuple}, {attr_tuple}, strict=True)}}
+
+            cls.__getstate__ = {cls.__name__}_getstate
             """)
         exec_dict = {"cls": cls, "_MODULE_SOURCE_CODE": augment_code}
         exec(compile(augment_code,
                      f"<dataclass augmentation code for {cls}>", "exec"),
              exec_dict)
+
+    # }}}
 
     # {{{ assign mapper_method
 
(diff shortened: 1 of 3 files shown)


pytato/array.py
```diff
@@ -336,15 +336,34 @@ def _augment_array_dataclass(
...
+        if attr_tuple:
+            attr_tuple = f"({attr_tuple},)"
+        else:
+            attr_tuple = "()"
+
+        fld_name_tuple = ", ".join(f"'{fld.name}'" for fld in fields(cls))
+        if fld_name_tuple:
+            fld_name_tuple = f"({fld_name_tuple},)"
+        else:
+            fld_name_tuple = "()"
```
matthiasdiener, line 355: Not sure if all of this is necessary since it is only used once below...

pytato/array.py
```diff
@@ -354,17 +373,26 @@ def {cls.__name__}_hash(self):
...
                     pass
 
-                h = hash(frozenset({attr_tuple}))
+                h = hash(frozenset({attr_tuple_hash}))
                 object.__setattr__(self, "_hash_value", h)
                 return h
 
             cls.__hash__ = {cls.__name__}_hash
+
+            def {cls.__name__}_getstate(self):
```
inducer, line 382: Is it OK to define `__getstate__` but not `__setstate__`? I looked but could not find a conclusive answer.

pytato/array.py
```diff
@@ -354,17 +373,26 @@ def {cls.__name__}_hash(self):
...
+                # This must not return the cached hash value.
+                return {{name: val
+                for name, val in zip({fld_name_tuple}, {attr_tuple}, strict=True)}}
+
+            cls.__getstate__ = {cls.__name__}_getstate
```
inducer, line 387: This is covered by `if generate_hash`. I'm guessing this is intentional? It could still be moved out, for improved clarity.

pytato/array.py
```diff
@@ -336,15 +336,34 @@ def _augment_array_dataclass(
...
+    # {{{ hashing and hash caching
 
     if generate_hash:
+        from dataclasses import fields
+
+        attr_tuple = ", ".join(f"self.{fld.name}" for fld in fields(cls))
+        if attr_tuple:
+            attr_tuple = f"({attr_tuple},)"
+        else:
+            attr_tuple = "()"
```
inducer, line 349: A small helper function to make these tuples might make this nicer to look at?

pytato/array.py
```diff
@@ -336,15 +336,34 @@ def _augment_array_dataclass(
...
+
+        fld_name_tuple = ", ".join(f"'{fld.name}'" for fld in fields(cls))
```
inducer, line 351: This and `attr_tuple` look the same?

inducer (commented):
Completely agree that this needs to be fixed. Just a few nits to pick below.

pytato/array.py
matthiasdiener: I think it is, but not 100% sure. Not necessary with [ab456e8](https://github.com/inducer/pytato/pull/563/commits/ab456e8fb96ad3564e9417d70976ca33c4d93660)

pytato/array.py
matthiasdiener: My idea was that we only need to mess with getstate/setstate when we use the hash caching. Not sure which way is better.

pytato/array.py
matthiasdiener: Not necessary with [ab456e8](https://github.com/inducer/pytato/pull/563/commits/ab456e8fb96ad3564e9417d70976ca33c4d93660)

pytato/array.py
matthiasdiener: They look similar, but aren't the same ;-)

pytato/array.py
```diff
@@ -379,12 +381,8 @@ def {cls.__name__}_hash(self):
...
             cls.__hash__ = {cls.__name__}_hash
 
-            def {cls.__name__}_getstate(self):
-                # This must not return the cached hash value.
-                return {{name: val
-                for name, val in zip({fld_name_tuple}, {attr_tuple}, strict=True)}}
-
-            cls.__getstate__ = {cls.__name__}_getstate
+            cls.__getstate__ = dataclasses._dataclass_getstate
+            cls.__setstate__ = dataclasses._dataclass_setstate
```
inducer, line 387: I'm not loving this, since it relies on undocumented functionality. Would you be up for simply reimplementing them?

matthiasdiener: Sure, what do you think of [e389e0f](https://github.com/inducer/pytato/pull/563/commits/e389e0fab42bbb4cbfa20e51ddca734d04bd1b17)?

matthiasdiener:
A somewhat unorthodox thought - should we even cache the hash of dataclasses, or could we rely on that any "large" field of the dataclass (say, an `immutabledict`) already caches their own hash?

matthiasdiener:
> A somewhat unorthodox thought - should we even cache the hash of dataclasses, or could we rely on that any "large" field of the dataclass (say, an `immutabledict`) already caches their own hash?

Nevermind, for recursive data structures like the DAG we will need cached hashes.

matthiasdiener:
A gentle ping for a review @inducer

inducer:
Thanks, LGTM!

PR #567, opened 2024-12-10
Remove `frozenset()` call from dataclass augmentation code

inducer:


pytato/array.py
```diff
@@ -371,7 +371,7 @@ def {cls.__name__}_hash(self):
                 except AttributeError:
                     pass
 
-                h = hash(frozenset({attr_tuple_hash}))
```
inducer, line 374: @matthiasdiener It looks like this was there from the [beginning](https://github.com/inducer/pytato/pull/521/files#diff-edba40fd7883253a818c8fb49fabe814d360c14b89a23c82a47874b983170c36R357), but I can't figure out why. Any thoughts?

It blew up with the release of numpy 2.2, failling like this: https://github.com/inducer/pytools/actions/runs/12245644757/job/34159956196.

matthiasdiener: Huh, this is very strange. This part of the code in #521 was inspired by https://github.com/inducer/pymbolic/pull/125, which does not have a `frozenset()` call. Not sure at the moment why I added it.

matthiasdiener: I suspect calling frozenset() here is wrong in any case, since not everything might be hashable?

inducer: I agree, it should not be here.

PR #568, opened 2024-12-10
Properly ignore axes tagged with `AxisIgnoredForPropagationTag` in tag propagation

a-alveyblanc:
Currently, tag propagation in the case of axes tagged with `AxisIgnoredForPropagationTag` is flawed in that it allows tags to be propagated through supposed ignored axes. However, the intended use is for propagation to be blocked to, from, and through ignored axes. This PR implements the intended handling by updating how reachable nodes are found in the undirected graph built from the edges computed by `AxesTagsEquationsCollector` . Effectively, nodes representing ignored axes or the `AxisIgnoredForPropagationTag` tag itself are considered partition points in the graph.

### DIFF
--- a/pytato/transform/metadata.py
+++ b/pytato/transform/metadata.py
@@ -755,9 +755,6 @@ def unify_axes_tags(
     for tag, var in equations_collector.known_tag_to_var.items():
         reachable_nodes = get_reachable_subgraph(propagation_graph, var)
         for reachable_var in (reachable_nodes - known_tag_vars):
-            ary, ax = equations_collector.axis_to_var.inverse[reachable_var]
-            if ary.axes[ax].tags_of_type(AxisIgnoredForPropagationTag):
-                continue
             axis_to_solved_tags.setdefault(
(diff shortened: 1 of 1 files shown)


pytato/transform/metadata.py
```diff
@@ -704,22 +700,57 @@ def unify_axes_tags(
...
 
     propagation_graph = undirected_graph_from_edges(
-        equations_collector.equations
-    )
+        equations_collector.equations)
 
-    for tag, var in equations_collector.known_tag_to_var.items():
-        if isinstance(tag, AxisIgnoredForPropagationTag):
-            continue
+    def get_reachable_nodes(
```
inducer, line 710: Why not add an `exclude_nodes` set argument to the generic (pytools) `get_reachable_nodes`? I think that would come out shorter and nicer-looking perhaps.

a-alveyblanc: That's a good idea. I'll close this and open a pytools PR

PR #569, opened 2024-12-12
Properly handle ignored axes during tag propagation

a-alveyblanc:
Creates a set of partition points from nodes representing `AxisIgnoredForPropagationTag` tags and axes tagged with `AxisIgnoredForPropagationTag`. 

Utilizes https://github.com/inducer/pytools/pull/278 to traverse the propagation graph while restricting traversals to partitions defined by the previously mentioned partition points.

This PR will stay as a draft until https://github.com/inducer/pytools/pull/278 gets merged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
     "bidict",
     "immutabledict",
     "loopy>=2020.2",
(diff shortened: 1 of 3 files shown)


inducer:
Thx!

PR #575, opened 2025-01-27
Reshape simple expressions in index lambda

### LINKED ISSUE #574: Lowering a Reshape to IndexLambda should produce simple expressions if possible
When we lower to an `IndexLambda` a `Reshape` expression we output a scalar expression that flattens out the indexing. This is not always necessary. Here is a small example.

```
import pytato as pt

x = pt.make_placeholder(name="x", shape=(10,4), dtype=float)
expr = pt.expand_dims(x, [2, 3])
idx_lambda = pt.transform.lower_to_index_lambda.to_index_lambda(expr)

first_index_expr = idx_lambda.expr.index_tuple[0]
from pymbolic.primitives import Variable
assert idx_lambda.expr.index_tuple[0] == Variable("_0")

assert idx_lambda.expr.index_tuple[1] == Variable("_1")
```

Currently, the variable `first_index_expr ` will have structure the `((0 + _0*1) % 10) // 1`.

nkoskelo:
Closes #574

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -420,9 +420,9 @@ class NodeCountMapper(CachedWalkMapper[[]]):
     def __init__(
             self,
             count_duplicates: bool = False,
-            _visited_functions: set[Any] | None = None,
+            visited_functions: set[Any] | None = None,
             ) -> None:
-        super().__init__(_visited_functions=_visited_functions)
+        super().__init__(visited_functions=visited_functions)
 
         from collections import defaultdict
         self.expr_type_counts: dict[type[Any], int] = defaultdict(int)
@@ -440,7 +440,7 @@ class NodeCountMapper(CachedWalkMapper[[]]):
     def clone_for_callee(self, function: FunctionDefinition) -> Self:
         return type(self)(
             count_duplicates=self.count_duplicates,
-            _visited_functions=self._visited_functions)
+            visited_functions=self._visited_functions)
 
     def post_visit(self, expr: Any) -> None:
         if not isinstance(expr, DictOfNamedArrays):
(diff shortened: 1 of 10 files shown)


nkoskelo:
@inducer this is ready for your review.

pytato/transform/lower_to_index_lambda.py
```diff
@@ -111,9 +111,11 @@ def _generate_index_expressions(
...
 
     return tuple(
-        # Mypy has a point: complex numbers don't support '//'.
-        (flattened_index_expn % old_size_till) // old_stride  # type: ignore[operator]
-        for old_size_till, old_stride in zip(old_size_tills, old_strides, strict=True))
+            # Mypy has a point: complex numbers don't support '//'.
+            (flattened_index_expn % old_size_till) // old_stride  # type: ignore[operator]
+            for old_size_till, old_stride in zip(old_size_tills,
+                                                 old_strides,
+                                                 strict=True))
```
inducer, line 118: Is there a change intended here? It seems this code is only reformatted. If so, why?

pytato/transform/lower_to_index_lambda.py
```diff
@@ -213,8 +215,33 @@ def _get_reshaped_indices(expr: Reshape) -> tuple[ScalarExpression, ...]:
...
+                                                  sub_new_shape[nind:], order,
+                                                  sub_index_vars[nind:]))
+                break
+        if len(sub_exprs) < len(sub_old_shape):
+            tmp = _generate_index_expressions(sub_old_shape[oind:],
+                                              sub_new_shape[nind-1:], order,
+                                              sub_index_vars[nind-1:])
+
+            sub_exprs = (*sub_exprs, *tmp)
+        index_expressions.append(sub_exprs)
```
inducer, line 244: This seems to go "against the grain" of the current design. So far, the idea is that first an mapping between input and output axes is established by looking at the shapes, and then fairly simple code produces the index expression *without* considering the shapes. The proposed change, IIUC, computes an inaccurate index mapping and then seems to want to paper over this fact by reexamining the shapes. IMO, an accurate index mapping should be computed from the get-go. A tool that may be helpful is allowing empty `*_ax_indices`.

nkoskelo: I moved the changes to `_get_indexed_expression` to maintain separation of concerns. I do not think we need to have empty `*_ax_indices` at this moment.

pytato/analysis/__init__.py
inducer, line 1: There are piles of seemingly unrelated changes here. What happened?

nkoskelo: Mypy was complaining about keyword arguments that begin with an underscore no longer existing.

PR #577, opened 2025-01-30
Reshape simplicity

### LINKED ISSUE #574: Lowering a Reshape to IndexLambda should produce simple expressions if possible

inducer:
@a-alveyblanc, could you review, since this touches code you wrote?

cc @nkoskelo 

Closes #575.
Closes #574.

### DIFF
--- a/pytato/transform/lower_to_index_lambda.py
+++ b/pytato/transform/lower_to_index_lambda.py
@@ -68,9 +68,9 @@ ToIndexLambdaT = TypeVar("ToIndexLambdaT", Array, AbstractResultWithNamedArrays)
 
 
 @dataclass(frozen=True)
-class _ReshapeIndexGroup:
-    old_ax_indices: tuple[ShapeComponent, ...]
-    new_ax_indices: tuple[ShapeComponent, ...]
+class _ReshapeShapeGroup:
+    old_ax_shape_group: tuple[ShapeComponent, ...]
+    new_ax_shape_group: tuple[ShapeComponent, ...]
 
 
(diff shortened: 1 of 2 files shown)


pytato/transform/lower_to_index_lambda.py
```diff
@@ -84,6 +84,15 @@ def _generate_index_expressions(
...
     new_strides = new_strides[:len(new_shape)]
 
+    if not old_shape:
+        assert new_shape == (1,)
+        return (0,)
+
+    if old_shape == new_shape:
+        # Avoid generating modulo expressions for direct pass-through
+        assert len(old_shape) == 1
+        return (index_vars[0],)
```
inducer, line 94: This may help avoid (quite) a few unnecessary modulos.

a-alveyblanc:
LGTM!

PR #578, opened 2025-02-01
Allow passing `FunctionDefinition` directly to `Mapper`

majosm:
Adds an overload to `Mapper.__call__` for passing `FunctionDefinition`, instead of just `ArrayOrNames`.

### DIFF
--- a/pytato/transform/__init__.py
+++ b/pytato/transform/__init__.py
@@ -38,6 +38,7 @@ from typing import (
     TypeAlias,
     TypeVar,
     cast,
+    overload,
 )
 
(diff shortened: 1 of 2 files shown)


majosm:
@inducer Mappers that use `@optimize_mapper` don't seem to like this change. They trigger this error:

```
NotImplementedError: You should not call an overloaded function. A series of @overload-decorated functions outside a stub module should always be followed by an implementation that is not @overload-ed.
```

Any idea why that might be?

inducer (approved):

PR #582, opened 2025-02-06
Change argument of `get_num_tags_of_type` to match `Taggable.tags_of_type`

### LINKED ISSUE #580: Suggestion: Make `get_num_tags_of_type` consistent with `Taggable.tags_of_type`
Currently `get_num_tags_of_type` accepts instances of tags, and `Taggable.tags_of_type` accepts tag types. Based on the function naming, it seems like these should accept the same thing. Probably accepting types makes more sense? Or, if instances are preferred, maybe the name should be changed? (`get_num_tagged`?)

@matthiasdiener Thoughts?

majosm:
FIxes #580.

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -31,7 +31,6 @@ from typing import TYPE_CHECKING, Any, Never
 from orderedsets import FrozenOrderedSet
 from typing_extensions import Self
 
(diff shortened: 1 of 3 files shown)


matthiasdiener (approved):

inducer:
Thx!

PR #583, opened 2025-02-17
Add `CacheInputs` class to simplify passing of `expr` + optional `args`/`kwargs`/`key`

majosm:
Attempts to simplify the handling for mappers with/without extra arguments, and also for keys being computed up front and passed in.

Note: This PR also disables the default implementations of `get_cache_key`/`get_function_definition_cache_key` for the extra args case; the previous implementations seem ambiguous, since a given argument can be passed with/without keyword and end up in different parts of the key tuple. I need to check if this works OK with mirgecom.

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -622,9 +622,9 @@ class TagCountMapper(CombineMapper[int, Never]):
         return sum(args)
 
     def rec(self, expr: ArrayOrNames) -> int:
-        key = self._cache.get_key(expr)
+        inputs = self._make_cache_inputs(expr)
         try:
-            return self._cache.retrieve(expr, key=key)
+            return self._cache.retrieve(inputs)
         except KeyError:
             s = super().rec(expr)
             if (
@@ -636,7 +636,7 @@ class TagCountMapper(CombineMapper[int, Never]):
             else:
                 result = 0 + s
 
-            self._cache.add(expr, 0, key=key)
+            self._cache.add(inputs, 0)
             return result
 
 
--- a/pytato/transform/__init__.py
+++ b/pytato/transform/__init__.py
@@ -354,7 +354,7 @@ class CachedMapperCache(Generic[CacheExprT, CacheResultT, P]):
     """
     def __init__(self) -> None:
         """Initialize the cache."""
-        self._expr_key_to_result: dict[CacheKeyT, CacheResultT] = {}
+        self._input_key_to_result: dict[CacheKeyT, CacheResultT] = {}
 
     def add(
             self,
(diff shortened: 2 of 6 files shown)


pytato/transform/__init__.py
```diff
@@ -304,74 +306,73 @@ def __call__(
 CacheKeyT: TypeAlias = Hashable
 
 
-class CachedMapperCache(Generic[CacheExprT, CacheResultT]):
+class CacheInputs(Generic[CacheExprT, P]):
```
inducer, line 309:
```suggestion
class CacheInputWithKey(Generic[CacheExprT, P]):
```
```suggestion
class CacheKeyable(Generic[CacheExprT, P]):
```
?

inducer (commented):
Forgot to submit these yesterday. This should be good to go once these are addressed.

pytato/transform/__init__.py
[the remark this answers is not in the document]
majosm: The previous version had caused a bit of a slowdown, so I refactored it a bit. Now it performs roughly the same as the current main.

majosm:
Ready for a look after #585.

pytato/analysis/__init__.py
```diff
@@ -622,11 +622,11 @@ def combine(self, *args: int) -> int:
...
 
     def rec(self, expr: ArrayOrNames) -> int:
-        key = self._cache.get_key(expr)
+        inputs = self._make_cache_inputs(expr)
         try:
-            return self._cache.retrieve(expr, key=key)
+            return self._cache.retrieve(inputs)
         except KeyError:
-            s = super().rec(expr)
+            s = Mapper.rec(self, expr)
```
inducer, line 629: Add comment as in #585.

pytato/transform/__init__.py
```diff
@@ -386,52 +390,67 @@ class CachedMapper(Mapper[ResultT, FunctionResultT, P]):
...
+                "Derived classes must override get_cache_key if using extra inputs.")
+        return expr
 
     def get_function_definition_cache_key(
                 self, expr: FunctionDefinition, *args: P.args, **kwargs: P.kwargs
-            ) -> Hashable:
-        return (expr, *args, tuple(sorted(kwargs.items())))
+            ) -> CacheKeyT:
+        if args or kwargs:
+            raise NotImplementedError(
```
inducer, line 420: Explain that cache could be different based on by-kw or by-position passing.

pytato/transform/__init__.py
```diff
@@ -386,52 +390,67 @@ class CachedMapper(Mapper[ResultT, FunctionResultT, P]):
...
-            else CachedMapperCache(self.get_function_definition_cache_key))
+            else CachedMapperCache())
 
     def get_cache_key(
                 self, expr: ArrayOrNames, *args: P.args, **kwargs: P.kwargs
-            ) -> Hashable:
-        return (expr, *args, tuple(sorted(kwargs.items())))
+            ) -> CacheKeyT:
+        if args or kwargs:
+            raise NotImplementedError(
```
inducer, line 412: Explain that cache could be different based on by-kw or by-position passing.

pytato/transform/__init__.py
majosm: Added.

inducer:
Thx!

PR #587, opened 2025-03-04
Placeholder: allow representation as lp.ValueArg

matthiasdiener:
Towards https://github.com/inducer/arraycontext/issues/303

### DIFF
--- a/pytato/tags.py
+++ b/pytato/tags.py
@@ -17,6 +17,7 @@ Pre-Defined Tags
 .. autoclass:: FunctionIdentifier
 .. autoclass:: CallImplementationTag
 .. autoclass:: InlineCallTag
+.. autoclass:: ForceValueArgTag
 """
 
 from dataclasses import dataclass
@@ -232,4 +233,11 @@ class InlineCallTag(CallImplementationTag):
     r"""
     A :class:`CallImplementationTag` that directs the
     :class:`pytato.target.Target` to inline the call site.
+    """
+
+
+@tag_dataclass
+class ForceValueArgTag(UniqueTag):
+    r"""
+    A :class:`Unique` that causes a Placeholder to become a ValueArg.
     """
--- a/pytato/tags.py
+++ b/pytato/tags.py
@@ -239,7 +239,8 @@ class InlineCallTag(CallImplementationTag):
 @tag_dataclass
 class ForceValueArgTag(UniqueTag):
     r"""
-    A :class:`UniqueTag` that causes a :class:`Placeholder` to become a :class:`loopy.ValueArg`. This is specific to code generation via
+    A tag that causes a :class:`Placeholder` to become a :class:`loopy.ValueArg`.
+    This is specific to code generation via
     :class:`pytato.target.loopy`. It is an error to apply this to a
-    :class:`Placeholder` that does not have empty shape.
+    :class:`Placeholder` that does not have an empty shape.
     """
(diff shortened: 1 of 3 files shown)


pytato/tags.py
```diff
@@ -233,3 +234,10 @@ class InlineCallTag(CallImplementationTag):
     A :class:`CallImplementationTag` that directs the
     :class:`pytato.target.Target` to inline the call site.
     """
+
+
+@tag_dataclass
+class ForceValueArgTag(UniqueTag):
+    r"""
+    A :class:`UniqueTag` that causes a Placeholder to become a ValueArg.
```
inducer, line 242:
```suggestion
    A :class:`UniqueTag` that causes a :class:`Placeholder` to become a :class:`loopy.ValueArg`. This is specific to code generation via
    :class:`pytato.target.loopy`. It is an error to apply this to a
    :class:`Placeholder` that does not have empty shape.
```

pytato/target/loopy/codegen.py
```diff
@@ -424,7 +425,15 @@ def map_placeholder(self, expr: Placeholder,
 
         shape = shape_to_scalar_expression(expr.shape, self, state)
 
-        arg = lp.GlobalArg(expr.name,
+        if expr.tags_of_type(ForceValueArgTag):
+            arg: lp.ArrayArg | lp.ValueArg = lp.ValueArg(expr.name,
```
inducer, line 429: Check that the shape is `()`, error if not.

pytato/tags.py
```diff
@@ -233,3 +234,10 @@ class InlineCallTag(CallImplementationTag):
     A :class:`CallImplementationTag` that directs the
     :class:`pytato.target.Target` to inline the call site.
     """
+
+
+@tag_dataclass
+class ForceValueArgTag(UniqueTag):
```
inducer, line 240: I wonder if we should contort ourselves here to portray this as target-independent.

inducer, line 240: Needs tests.

inducer (commented):
Thanks! In general, this is exactly what I had in mind. Just a few nits/thoughts below.

pytato/tags.py
matthiasdiener: Maybe rename to `ForceLoopyValueArgTag` or so?

pytato/target/loopy/codegen.py
matthiasdiener: Added in [f6a676a](https://github.com/inducer/pytato/pull/587/commits/f6a676a5d3105581175c492abeb69a515afd71c1)

kaushikcfd: Why not make scalar shaped placeholders ValueArgs all the time? Given that placeholders in potato are read-only entities, it should be sound.

inducer: Disagree here. On-device "array" scalars are a thing (think: result of a dot product). We should leave the door open for having them. Only user code can know what type of scalar we have.

pytato/tags.py
inducer: Upon further thought, I think this is good enough as it is.

pytato/target/loopy/codegen.py
```diff
@@ -424,7 +425,16 @@ def map_placeholder(self, expr: Placeholder,
 
         shape = shape_to_scalar_expression(expr.shape, self, state)
 
-        arg = lp.GlobalArg(expr.name,
+        if expr.tags_of_type(ForceValueArgTag):
+            assert expr.shape == ()
```
inducer, line 429:
```suggestion
            if expr.shape != ():
                raise ValueError("ForceValueArgTag applied to non-scalar")
```

PR #589, opened 2025-03-24
Don't use deduplicated predecessors in mapper-created duplication check

majosm:
`DirectPredecessorsGetter` returns a deduplicated set of predecessors, which causes problems with the "mapper-created duplication" check in `TransformMapper`. Example: Suppose an `IndexLambda` has two bindings that are duplicates of each other. When applying a mapper that deduplicates the DAG, such as `CopyMapper`, the recursion will deduplicate these two arrays, which means that a new `IndexLambda` must be created. This should not trigger the mapper-created duplicate error, because it should see that one of the predecessors has changed. However because `DirectPredecessorsGetter` deduplicates internally (by virtue of returning a `FrozenOrderedSet`), it will report that the `IndexLambda` has only one predecessor. Subsequently, the duplication check won't see that a predecessor has changed, and it will trigger the error.

This change adds a `ListOfDirectPredecessorsGetter` that returns a list instead of a set and uses that to implement the duplication check.

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -26,7 +26,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import TYPE_CHECKING, Any, Never
+from typing import TYPE_CHECKING, Any, Never, overload
 
 from orderedsets import FrozenOrderedSet
 from typing_extensions import Self
@@ -75,6 +75,7 @@ __doc__ = """
 .. autofunction:: get_num_call_sites
 
 .. autoclass:: DirectPredecessorsGetter
+.. autoclass:: ListOfDirectPredecessorsGetter
 
 .. autoclass:: TagCountMapper
 .. autofunction:: get_num_tags_of_type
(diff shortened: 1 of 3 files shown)


pytato/analysis/__init__.py
inducer, line 1: Add regression test.

pytato/transform/__init__.py
```diff
@@ -576,8 +576,8 @@ def add(
             f"Cache entry is already present for key '{key}'."
 
         if self.err_on_created_duplicate:
-            from pytato.analysis import DirectPredecessorsGetter
-            pred_getter = DirectPredecessorsGetter(include_functions=True)
+            from pytato.analysis import NonUniqueDirectPredecessorsGetter
+            pred_getter = NonUniqueDirectPredecessorsGetter(include_functions=True)
```
inducer, line 580: Add a comment on the rationale for the `NonUniqueDPG` is preserved

pytato/analysis/__init__.py
majosm: Added. Without this fix it fails with the following error:

```
        # Should not raise
>       NonDuplicatingMapper()(dag)

test_pytato.py:1139: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../pytato/transform/__init__.py:298: in __call__
../pytato/transform/__init__.py:510: in rec
    return self._cache_add(inputs, Mapper.rec(self, expr, *args, **kwargs))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test_pytato.NonDuplicatingMapper object at 0x11fbb6630>, inputs = <pytato.transform.CacheInputsWithKey object at 0x12ad2ea80>
result = IndexLambda(shape=(10, 10), dtype='float64', expr=Sum((Subscript(Variable('_in0'), (Variable('_0'), Variable('_1'))), ...laceholder(shape=(10, 10), dtype='float64', name='x'), '_in1': Placeholder(shape=(10, 10), dtype='float64', name='x')})

    def _cache_add(
            self,
            inputs: CacheInputsWithKey[ArrayOrNames, []],
            result: ArrayOrNames) -> ArrayOrNames:
        try:
            return self._cache.add(inputs, result)
        except MapperCreatedDuplicateError as e:
>           raise ValueError(
                f"mapper-created duplicate detected on {type(inputs.expr)} in "
                f"{type(self)}.") from e
E           ValueError: mapper-created duplicate detected on <class 'pytato.array.IndexLambda'> in <class 'test_pytato.NonDuplicatingMapper'>.

../pytato/transform/__init__.py:654: ValueError
```

pytato/transform/__init__.py
majosm: Added.

pytato/analysis/__init__.py
```diff
@@ -321,12 +322,12 @@ def is_einsum_similar_to_subscript(expr: Einsum, subscripts: str) -> bool:
 # }}}
 
 
-# {{{ DirectPredecessorsGetter
+# {{{ NonUniqueDirectPredecessorsGetter
 
-class DirectPredecessorsGetter(
+class NonUniqueDirectPredecessorsGetter(
```
inducer, line 327:
```suggestion
class ListOfDirectPredecessorsGetter(
```

pytato/analysis/__init__.py
```diff
@@ -342,90 +343,121 @@ def __init__(self, *, include_functions: bool = False) -> None:
...
+            []]):
+    """
+    Mapper to get the
+    `direct predecessors
+    <https://en.wikipedia.org/wiki/Glossary_of_graph_theory#direct_predecessor>`__
+    of a node.
+
+    .. note::
+
+        We only consider the predecessors of a nodes in a data-flow sense.
```
inducer, line 445:
```suggestion
        We only consider the predecessors of a node in a data-flow sense.
```
(Also above.)

test/test_pytato.py
```diff
@@ -1049,6 +1054,97 @@ def test_created_at():
     # }}}
 
 
+class NonDuplicatingMapper(pt.transform.TransformMapper):
```
inducer, line 1057:
```suggestion
# {{{ test_mapper_duplication_check

class NonDuplicatingMapper(pt.transform.TransformMapper):
```

test/test_pytato.py
```diff
@@ -1049,6 +1054,97 @@ def test_created_at():
...
+    NonDuplicatingMapper()(deduped_dag)
+
+    # Should not raise
+    NonDuplicatingMapper()(dag)
+
+    with pytest.raises(ValueError):
+        DuplicatingMapper()(deduped_dag)
+
+    with pytest.raises(ValueError):
+        DuplicatingMapper()(dag)
```
inducer, line 1145:
```suggestion
        DuplicatingMapper()(dag)

# }}}
```

pytato/analysis/__init__.py
```diff
@@ -342,90 +343,121 @@ def __init__(self, *, include_functions: bool = False) -> None:
...
+# }}}
 
 
+# {{{ DirectPredecessorsGetter
+
+class DirectPredecessorsGetter(
+        Mapper[
+            FrozenOrderedSet[ArrayOrNames | FunctionDefinition],
+            FrozenOrderedSet[ArrayOrNames],
+            []]):
```
inducer, line 436: Can this avoid inheriting from `Mapper`, given that, in normal operation, it should use none of its functionality? (`__call__` notwithstanding, which could be reimplemented.)

inducer (approved):
A few simple things, then this is ready to go.

pytato/analysis/__init__.py
majosm: Done.

PR #597, opened 2025-05-01
[RFC] execute_distribute_partition: allow recording profiling events

matthiasdiener:


inducer:
This is not the way to go, IMO. Pytato knows nothing about array contexts, and I'd like to keep it that way. Instead, we could replace the loopy executor upon code generation with a profiling one that's connected to the array context.

PR #605, opened 2025-06-02
Loosen type constraint on `data` in `make_dict_of_named_arrays`

majosm:
`DictOfNamedArrays.__init__` takes `Mapping[str, Array]`, so `make_dict_of_named_arrays` might as well use that too.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -27072,14 +27072,6 @@
             {
                 "code": "reportUnknownParameterType",
                 "range": {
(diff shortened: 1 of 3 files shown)


inducer:
> Side question: Why does `DictOfNamedArrays` use `Mapping` internally, instead of something more specific like `immutabledict`?

No particular reason. A switch to `constantdict` could be done, but it might entail a bunch of type conversion that's not necessary for the generic `Mapping`.

PR #606, opened 2025-06-04
Tweak `Array` towards compatibility with arraycontext's `Array` protocol

majosm:
pyright doesn't like `partialmethod`, nor `Array.__getitem__` having an argument name that doesn't match arraycontext's `Array` protocol.

### DIFF
--- a/pytato/cmath.py
+++ b/pytato/cmath.py
@@ -145,6 +145,8 @@ def _get_dtype(x: ArrayOrScalar) -> _dtype_any:
         return np.dtype(type(x))
 
 
+# FIXME: Overload these instead of returning union type?
+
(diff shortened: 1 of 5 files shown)


pytato/cmath.py
```diff
@@ -145,6 +145,8 @@ def _get_dtype(x: ArrayOrScalar) -> _dtype_any:
         return np.dtype(type(x))
 
 
+# FIXME: Overload these instead of returning union type?
```
inducer, line 148: Yes, I think that'd be an improvement.

inducer (approved):
Thanks for puzzling this out! This looks great. I'll merge this now, the element-wise overloads can be done in a separate PR.

PR #614, opened 2025-07-01
Enable duplication checks by default

majosm:
Duplication checks have been off by default due to some subpackages still having code that duplicates. Now that inducer/arraycontext#221 has been merged, it may be possible to enable the checks by default. Likely there are still a few spots in arraycontext/meshmode/grudge (and the production versions of those) that need fixing, but hopefully they should be minor.

Draft because:

- [x] Requires inducer/arraycontext#324 and inducer/arraycontext#325
- [x] Requires inducer/meshmode#467
- [x] Requires inducer/grudge#384
- [x] Merge meshmode production changes

### DIFF
--- a/pytato/transform/__init__.py
+++ b/pytato/transform/__init__.py
@@ -435,7 +435,7 @@ class CachedMapper(Mapper[ResultT, FunctionResultT, P]):
     """
     def __init__(
             self,
-            err_on_collision: bool = False,
+            err_on_collision: bool = __debug__,
             _cache:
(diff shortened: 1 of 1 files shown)


inducer:
#616 should clean up some of the bpr nonsense.

majosm:
Note to self: These should probably be `err_on_<whatever>=__debug__` rather than `err_on_<whatever>=True`.

inducer (approved):
LGTM subject to:

- Let's defer `sorted` removal until we're on `constantdict`

PR #623, opened 2025-10-20
Flop counting

inducer:
Very WIP right now.

cc @majosm 

TODO:
- [x]  ~~Depends on #657~~
- [x]  ~~Depends on #660~~

### DIFF
--- a/pytato/analysis/__init__.py
+++ b/pytato/analysis/__init__.py
@@ -976,15 +976,10 @@ class MaterializedNodeFlopCounter(CachedWalkMapper[[]]):
     def __init__(
             self,
             op_name_to_num_flops: Mapping[str, int],
-            _visited_functions: set[VisitKeyT] | None = None,
-            _function_to_nflops: dict[FunctionDefinition, ArrayOrScalar] | None = None
             ) -> None:
-        super().__init__(_visited_functions=_visited_functions)
+        super().__init__()
         self.op_name_to_num_flops: Mapping[str, int] = op_name_to_num_flops
         self.materialized_node_to_nflops: dict[Array, ArrayOrScalar] = {}
-        self.call_to_nflops: dict[Call, ArrayOrScalar] = {}
-        self._function_to_nflops: dict[FunctionDefinition, ArrayOrScalar] = \
-            _function_to_nflops if _function_to_nflops is not None else {}
         self._per_entry_flop_counter: _PerEntryFlopCounter = _PerEntryFlopCounter(
             self.op_name_to_num_flops)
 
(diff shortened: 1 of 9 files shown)


pytato/analysis/__init__.py
```diff
@@ -500,7 +525,7 @@ def clone_for_callee(self, function: FunctionDefinition) -> Self:
             count_duplicates=self.count_duplicates,
             _visited_functions=self._visited_functions)
 
-    def post_visit(self, expr: Any) -> None:
+    def post_visit(self, expr: ArrayOrNames | FunctionDefinition) -> None:
```
inducer: Would it make sense to define `ArrayOrNamesOrFunction`?

pytato/scalar_expr.py
```diff
@@ -241,6 +243,140 @@ def map_type_cast(self,
...
+    def map_variable(self, expr: prim.Variable) -> set[str]:
+        return {expr.name}
+
+
+class FlopCounter(FlopCounterBase):
+    def __init__(
+            self,
+            op_name_to_num_flops: Mapping[str, ArithmeticExpression] | None = None):
+        super().__init__()
+        self.op_name_to_num_flops: dict[str, ArithmeticExpression]
```
inducer: Move this to the class body for better visibility?

pytato/transform/calls.py
```diff
@@ -56,6 +63,89 @@
...
+    def combine(self, *args: frozenset[Array]) -> frozenset[Array]:
+        from functools import reduce
+        return reduce(lambda a, b: a | b, args, cast("frozenset[Array]", frozenset()))
+
+    @override
+    def map_call(self, expr: Call) -> frozenset[Array]:
+        return frozenset(expr.bindings.values())
+
+
+class _CallMaterializer(CopyMapper):
```
inducer: I'm not sure I understand this. This appears to enforce a specific view of whether function arguments and/or return values are materialized, when I would prefer that pytato have no opinion in the matter. They could/could not be materialized; they're just expressions, just like everyone else.

Could you explain the intent behind this?

pytato/utils.py
```diff
@@ -734,4 +737,20 @@ def get_einsum_specification(expr: Einsum) -> str:
...
+def is_materializable(expr: ArrayOrNames | FunctionDefinition) -> bool:
+    """
+    Returns *True* if *expr* is an instance of an array type that can be materialized.
+    """
+    from pytato.array import InputArgumentBase, NamedArray
+    from pytato.distributed.nodes import DistributedRecv, DistributedSendRefHolder
+    return (
+        isinstance(expr, Array)
+        and not isinstance(expr, (
+            # FIXME: Is there a nice way to generalize this?
```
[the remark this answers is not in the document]
majosm: I'm using this in the flop counting code to tell me a couple of things:

1. Am I allowed to materialize/unmaterialize this expression? (Can't for things like `NamedArray` and `DistributedSendRefHolder`.)
2. Can this expression be lowered to an index lambda? (Can't for things like `InputArgumentBase` and `DistributedRecv`.)

pytato/analysis/__init__.py
```diff
@@ -728,4 +755,363 @@ def update_for_Array(self, key_hash: Any, key: Any) -> None:
...
+        "min": 1,
+        "max": 1}
+
+
+def get_num_flops(
+        expr: ArrayOrNames,
+        op_name_to_num_flops: Mapping[str, int] | None = None,
+    ) -> ArrayOrScalar:
+    """
+    Count the total number of floating point operations in the DAG *expr*.
```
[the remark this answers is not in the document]
majosm: Something like https://github.com/inducer/pytato/pull/623/commits/40dbe202e8420cde2f393313cd58406ecf2e4d5c?

pytato/transform/calls.py
majosm: I removed the call materialization and made the flop counting require function calls to be inlined beforehand for now.

pytato/scalar_expr.py
```diff
@@ -241,6 +243,141 @@ def map_type_cast(self,
...
         return f"cast({expr.dtype}, {inner_str})"
 
+
+class InputGatherer(Collector[str, []]):
+    @override
+    def map_variable(self, expr: prim.Variable) -> set[str]:
+        return {expr.name}
+
+
+class FlopCounter(FlopCounterBase):
```
inducer: Let's do a separate node type for `nflops`.

inducer (commented):
Just the enum thing and then this is good to go from my perspective. 🎉

kaushikcfd (approved):
LGTM! Thanks, I will be a user of this FLOP-counting routine soon! :)

PR #632, opened 2026-02-06
Fix default colorize_code in pt.generate_jax.

kaushikcfd:
With `None` as the default, pt.generate_jax will respected the NO_COLOR (<https://no-color.org/>) environment variable.


inducer:
Thx!

PR #661, opened 2026-04-24
Support axis tagging in `CSRMatmul`

majosm:
Adds axis tagging for the reduction in `CSRMatmul`, analogous to those in `IndexLambda` and `Einsum`.


inducer:
Thanks!

